PR #3, opened 2012-05-03
Added timeout and charset parameters to mysql connection string

lgsonic:
Extended Mysql backend connection string by supporting timeout and charset parameters. In addition, added Mysql reconnect option.

### DIFF
--- a/src/backends/mysql/session.cpp
+++ b/src/backends/mysql/session.cpp
@@ -153,7 +153,9 @@ void parse_connect_string(const string & connectString,
     string *password, bool *password_p,
     string *db, bool *db_p,
     string *unix_socket, bool *unix_socket_p,
-    int *port, bool *port_p)
+    int *port, bool *port_p,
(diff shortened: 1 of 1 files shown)


vadz:
FWIW this is still relevant, but I don't think it's critical for 4.0.0, so I'm removing the milestone.

mloskot:
This should be either closed as not merged or reworked using the new `#include "soci/connection-parameters.h"` feature and, I think, there should be four parameters, one for each of the options:

```
MYSQL_OPT_RECONNECT
MYSQL_OPT_CONNECT_TIMEOUT
MYSQL_OPT_READ_TIMEOUT
MYSQL_OPT_WRITE_TIMEOUT
```

So users can control the timeouts separately as they demand. Default behaviour should stay unchanged.

Thoughts?

vadz:
FWIW I agree that this should be implemented in terms of `connection_parameters`, please reopen/recreate if you can do it like this, TIA!

Issue #19, opened 2012-11-19
Server-side memory leak in PostgreSQL


b-s-a:
I try use SOCI in my project. But after some time I found, that SQL server (postgresql-8.4, 9.1, and 9.2) has eat all memory and killed by oom-killer. After some investigations I update library to recent master (today). But problem not fixed yet.

PR #24, opened 2012-11-30
Fix for issue #19

b-s-a:
Statement with repeatable query was never deallocated on server side
see https://github.com/SOCI/soci/issues/19

### DIFF
--- a/src/backends/postgresql/soci-postgresql.h
+++ b/src/backends/postgresql/soci-postgresql.h
@@ -144,6 +144,7 @@ struct postgresql_session_backend;
 struct postgresql_statement_backend : details::statement_backend
 {
     postgresql_statement_backend(postgresql_session_backend & session);
+    ~postgresql_statement_backend();
 
     virtual void alloc();
     virtual void clean_up();
--- a/src/backends/postgresql/statement.cpp
+++ b/src/backends/postgresql/statement.cpp
@@ -37,6 +37,12 @@ postgresql_statement_backend::postgresql_statement_backend(
 {
 }
 
+postgresql_statement_backend::~postgresql_statement_backend()
+{
+    if (statementName_.empty() == false)
+        session_.deallocate_prepared_statement(statementName_);
+}
+
 void postgresql_statement_backend::alloc()
 {
     // nothing to do here


michaelainsworth:
I'm currently experiencing this issue with soci 3.2.2 (installed via homebrew on Mac OS X) and PostgreSQL 9.4.5 (also installed via homebrew).

Can I give you more information in trying to diagnose why this issue has reappeared?

snikulov:
@michaelainsworth I'm not sure that it's same issue. AFAIS changes is present in v3.2.2.
BTW, can you change update homebrew for latest version v3.2.3?

michaelainsworth:
@snikulov, I had a quick attempt at updating the homebrew formula earlier this evening and will have another attempt when I get the chance. Will let you know. It's definitely the same error message re unnamed queries and there is no obvious pattern to its occurrence.

michaelainsworth:
@snikulov, I've just tested this with soci 3.2.3 (https://github.com/SOCI/soci/archive/3.2.3.zip, sha256 "10d4a66218780f2ea85e5b9454e478e2c45370f7264b5b03957ba1abaf9cd451") with PostgreSQL 9.5.0 and it's present.

michaelainsworth:
I put some debug messages in the destructor. Looks like either the named statement is duplicated, or the string containing the name is not set to an empty string when the deallocate statement is issued (I'm guessing - haven't had much of a chance to look properly).

[soci-issue.txt](https://github.com/SOCI/soci/files/102224/soci-issue.txt)

michaelainsworth:
Might not have been right about the destructor. I turned PostgreSQL logging to DEBUG5 and found the following:

```
DEBUG:  bind <unnamed> to <unnamed>
ERROR:  unnamed prepared statement does not exist
```

For correct prepared statements, this line should look like:

```
DEBUG:  bind <unnamed> to st_8
```

michaelainsworth:
I don't know if this helps, but on line 382 of statement.cpp, the call to `result_.check_for_data()` throws. Immediately prior to this, if you output the contents of `statementName_` and `query_` to `std::cout`, you'll see that they're both empty strings.

michaelainsworth:
Sorry for so many posts, but hopefully I'm helping solve the problem here. The `statementName_` and `query_` variables are definitely being changed to an empty string somewhere. In the output below, the memory address 140493244604192 (as an `unsigned long long`) points to a `postgresql_statement_backend`, and on the third execution, you can see the two variables mentioned becoming empty strings.

```
RE-EXECUTING: 140493244604192: st_3
    select event_uuid::text, event_type_name, event_timestamp::text, event_data::text, aggregate_uuid::text, aggregate_type_name, aggregate_version from event_log where event_type_name = $1 order by event_timestamp
RE-EXECUTING: 140493244604192: st_3
    select event_uuid::text, event_type_name, event_timestamp::text, event_data::text, aggregate_uuid::text, aggregate_type_name, aggregate_version from event_log where event_type_name = $1 order by event_timestamp
RE-EXECUTING: 140493244604192:

ABOUT TO THROW: 7 PGRES_FATAL_ERROR
ERROR IN CHECKING FOR DATA: 140493244604192:

unknown location:0: fatal error in "DuplicateUuids": std::runtime_error: Cannot execute query 2 ERROR:  unnamed prepared statement does not exist
```

vadz:
I'm very confused by the recent comments here, it looks like there is a problem, but it's not clear at all what the problem actually _is_. Do you have a [SSCCE](http://sscce.org/) for it or at least an explanation of what exactly goes wrong?

Also, it might be better to open a new ticket for it to keep things more clear.

PR #30, opened 2013-01-23
Set INSTALL_NAME_DIR as used on OS X

arsenm:


vadz:
@arsenm Apparently this wasn't the right thing to do and needs to be reverted, please see the discussions in #929. Do you have any objections to/thoughts about it by chance?

PR #115, opened 2013-03-31
Auto-reconnect if the connection goes bad during statement execution

ricardofandrade:
It also re-prepares previously-prepared statements in this case
Internal commands such as BEGIN, ROLLBACK, COMMIT, DEALLOCATE are also able to reconnect
Interactive test included
Enjoy :)

### DIFF
--- a/src/backends/postgresql/session.cpp
+++ b/src/backends/postgresql/session.cpp
@@ -24,13 +24,21 @@
 #pragma warning(disable:4355 4996)
 #endif
 
+#ifndef WIN32
+#include <unistd.h> // for sleep()
+#else
+namespace {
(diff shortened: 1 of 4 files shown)


mloskot:
@ricardofandrade As you may know, we have changed the branching model and renamed master to develop.deleted the develop branch, that's closed this PR.

Because this PR is based on old source structure, I think it's OK to keep it closed.
You mentioned, it's more a sample and new implementation should target 4.0.0.
So, if you are still going to work on proposal of the new implementation, please do it in new PR.
Sorry for trouble.

ricardofandrade:
It may take a while, but I'll try it again for 4.0.0

Issue #136, opened 2013-04-15
MySQL test8 assumes allocation will happen at same address.


ricardofandrade:
If you allocate any amount of memory inside `session::reconnect()` just before the line:
`backEnd_ = lastFactory->make_session(lastConnectParameters_);` MySQL test8 crashes.
MySQL statement backend implementation is assuming that the pointer returned by this line will be the same of the one deleted by a previous call to `session::close()`.

PR #156, opened 2013-06-02
Fix the issue #136 across all backends.

ricardofandrade:
See issue #136

The issue is cause because the session backend gets deleted during reconnect.
This solution simply don't do that.
With this fix session backends gained support for open connections and close them (via clean_up).
Also, it's possible to check if session backend is opened though in very simplistic way that just works. That can be improved in later versions.

No public API was changed and I tried to be conservative as possible with the backend objects.

This is a reasonable complex issue, so reviews are welcome.

I'm targeting soci 3.2.2 with fix pull request.

Please note DB2 backend depend (lightly) on #146 to work properly.
Other small DB2 fix is included in this request.

### DIFF
--- a/src/backends/db2/factory.cpp
+++ b/src/backends/db2/factory.cpp
@@ -15,10 +15,9 @@ using namespace soci::details;
 
 
 // concrete factory for ODBC concrete strategies
-db2_session_backend * db2_backend_factory::make_session(
-     connection_parameters const & parameters) const
+db2_session_backend * db2_backend_factory::make_session() const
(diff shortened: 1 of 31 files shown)


mloskot:
This has been dangling for long time w/o any move forward. It also seems to target SOCI 3.x.

@vadz If you don't mind, I'm going to close this PR as not merged or partially merged.

vadz:
Closing as suggested 3.5 years ago as this is not applicable any longer.

PR #190, opened 2013-08-27
Time stamps from Postgres are UTC

jsonn:


vadz:
Just for the record (and to clear the confusion due to closing/opening it), there is definitely a bug here which needs to be addressed. What I don't understand is how do our existing tests, which check for round tripping timestamps, pass currently...

vadz:
> What I don't understand is how do our existing tests, which check for round tripping timestamps, pass currently...

OK, I understand it now: our tests only check `struct tm` fields which are, of course, exactly the same whether we use `mktime()` or `timegm()`. The difference between these two functions is their return value, but we throw it away anyhow, and, for the platforms which have `tm_gmtoff` and/or `tm_zone` fields in `struct tm` (i.e. not Windows with MSVC), the values of these fields.

So if I add comparison of `tm_gmtoff` in the original struct and the one obtained from SOCI, I can indeed make the test fail for PostgreSQL. The trouble is that it *also* fails for SQLite3, MySQL, ODBC and, although I didn't test those, I think DB2 and Oracle as well. Basically all the uses of `mktime()` in the current code seem to be incorrect.

So should we replace all of them with `timegm()` wrapper?

mloskot:
Good catch! I support your idea of the wrapper.

I need to set  up Oracle environment soon.

vadz:
Damn, time zones are confusing. I did implement the proposed wrapper, but now I realize that it doesn't really fix the problem, it just replaces one problem with another one: before the following check
```cpp
        std::time_t const t = std::time(NULL);
        std::tm tm_in = *mktime(&t);
        sql << "insert into soci_test(tm) values(:tm)", use(tm_in);
        std::tm tm_out;
        sql << "select tm from soci_test", into(tm_out);
        CHECK(tm_out.tm_gmtoff == tm_in.tm_gmtoff);
```
passed. With my changes it passes _only_ if `mktime()` is replaced with `gmtime()`, which is hardly better.

Unfortunately, thinking more about this (perhaps I should have started by doing this...), I just don't see how could it work in both cases: the time zone information is lost when writing to the database because we don't use `TIMESTAMP WITH TIME ZONE` type (and it's not clear if we should).

So the only reasonable thing to do is to tell people that they must ignore the time zone stored in `struct tm` they get from SOCI on the platforms where it has one inside it. And in this case it doesn't matter whether we use `mktime()` or `gmtime()` anyhow, so there is nothing more to do here...

PR #203, opened 2013-09-18
Add WITH_CXX11 configuration flag (currently for gcc and clang)

vinesworth:
Hello!

Here's a patch for SociConfig.cmake. Instead of forcing the C++ standard to be 98, it adds a configuration switch. Unless the latter is explicitly turned on, the default mode is 98. Tested with Clang 3.3 and GCC 4.8.1 on Linux manually, since I'm not sure how to unit-test it (and not familiar with Travis, sorry).

(As of MSVC, its current level of C++11 support leaves me doubtful if it would compile at all...)

### DIFF
--- a/cmake/SociConfig.cmake
+++ b/cmake/SociConfig.cmake
@@ -12,6 +12,15 @@
 # Force compilation flags and set desired warnings level
 #
 
+if(WITH_CXX11)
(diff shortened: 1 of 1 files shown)


mloskot:
Since #336, we've got `SOCI_CXX_C11` (default is OFF).

PR #213, opened 2013-10-22
Fix in soci::oracle to allow spaces in the params

ayllon:
chop_connect_string didn't accept parameters containing spaces inside.
Now, spaces can be used encapsulating the value between quotes.

user=oracle-user password=oracle-password service="(DESCRIPTION = ( ADDRESS= .... ) )"


vadz:
I have no way to test this, but the code looks fine, so I'll go ahead and merge it -- and hope that someone who does use Oracle backend can test it.

Thanks!

PR #215, opened 2013-10-22
minor bug fixes and improvments

ArnaudD-FR:
Same as #184 but without shared_ptr. (and based on soci:develop branch)


mloskot:
Most of the changes here have been already applied to the current master, one way or another.

I just cherry-picked remaining two. Even if they are not accepted, they can be reviewed easier and archived.

PR #245, opened 2014-04-01
Add logger class to allow customizing SOCI logging operations.

vadz:
No big changes yet, but it is now possible to use a custom logger object which
may do something else than just write the query string passed to it to a
stream.

The old behaviour is preserved by using the standard SOCI logger by default
which does write the query to the stream.

---

This is the first step in addressing #138. This approach should be less disruptive (full backwards compatibility is preserved) than the patch currently there and, eventually, more powerful.

### DIFF
--- a/src/core/session.cpp
+++ b/src/core/session.cpp
@@ -26,17 +26,63 @@ void ensureConnected(session_backend * backEnd)
     }
 }
 
+// Standard logger class used by default.
+class standard_logger_impl : public logger_impl
+{
+public:
+    standard_logger_impl()
+    {
+        logStream_ = NULL;
+    }
+
+    virtual void start_query(std::string const & query)
+    {
+        if (logStream_ != NULL)
+        {
+            *logStream_ << query << '\n';
+        }
+
+        lastQuery_ = query;
+    }
+
+    virtual void set_stream(std::ostream * s)
+    {
+        logStream_ = s;
+    }
+
+    virtual std::ostream * get_stream() const
+    {
+        return logStream_;
+    }
+
+    virtual std::string get_last_query() const
+    {
+        return lastQuery_;
+    }
+
+private:
+    virtual logger_impl* do_clone() const
+    {
+        return new standard_logger_impl;
+    }
+
+    std::ostream * logStream_;
+    std::string lastQuery_;
+};
+
 } // namespace anonymous
 
 session::session()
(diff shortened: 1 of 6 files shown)


mloskot:
@vads I see your branch `better-query-log` is still present in your fork. I'm going to pull it and merge, unless you'd prefer to do it.

I'm sorry for the inconvenience caused by premature closing of this PR.

vadz:
@mloskot I've completely forgot everything about it by now, to be honest, but I'll try to have a look at it soon to refresh my memory.

mloskot:
@vadz If you want and have time and still care about this, otherwise we can just close and forget.

vadz:
I think it can be merged now, please review the changes if possible.

Note that I had to restart a few of Travis CI builds and even had to do it multiple times with the DB2 one as they were failing because they couldn't connect to ppa.launchpad.net and so failed to install the required CMake version. I have no idea why was this happening, let's hope it's just some Travis network issue that will go away on its own.

mloskot:
@vadz I can confirm the Travis CI issue. It seems to be related to `apt` cache or something, and it is not uncommon, not specific to SOCI, I've encountered it with at least 2-3 other projects. Annoying.

src/core/session.cpp
```diff
@@ -26,17 +26,63 @@ void ensureConnected(session_backend * backEnd)
...
+        return logStream_;
+    }
+
+    virtual std::string get_last_query() const
+    {
+        return lastQuery_;
+    }
+
+private:
+    virtual logger_impl* do_clone() const
```
mloskot: Wouldn't it be clearer to name it `make_clone()`?

At first, I wondered if there could be just one `clone()` method though, but I guess there is need to separate cloning action from a clone construction as two distinct operations.

vadz: It's just the convention of using a private `do_xxx()` to "really do" whatever the public `xxx()` does, after performing some extra checks on top of it. Note that this convention is already used in SOCI (e.g. a quick search finds `do_unload()` in src/core/backend-loader.cpp). In this case, `do_clone()` makes the the real clone and the public `clone()` verifies that it has actually succeeded. It might be an overkill, but OTOH it doesn't really cost much and allows to detect problems in the custom `do_xxx()` implementation sooner and signals them via an exception rather than crashes, and it's not like logger cloning is a performance-sensitive operation, so why not?

mloskot: Your explanation makes sense. Thx

PR #265, opened 2014-07-07
Multiple changes in SOCI

ravselj:
- MSVC warnings fixes and pragmas
- CMAKE changes(added MSVC /MP multiproces compiling support, CMAKE_DEBUG_POSTFIX variable can be use to tell which suffix soci wil use for debug modules)
- common test28, test29 produce memory leaks so solution is made in boost-fusion.h and values.h
- soci_backend added new virtual function create_database(const std::string&)
- soci_firebird change in common.cpp( see SQL_TEXT )
- some additional firebird and sqlite specific tests
- sqlite blob reimlemented using std::vector<char>


vadz:
Sorry, I don't think this can be realistically applied in the current state. The commits are too confusing: they do a lot of things (instead of each commit doing one thing and one thing only) and there are no explanations about why do they do them. E.g. the commit related `sqlite3_close()`:

 1. Does a lot of things not related to this function.
 2. Doesn't say *why* does it replace it with the old version.

There are undoubtedly valuable parts here, but this needs to be reworked/cherry-picked and can't be just merged as is.

ravselj:
Sounds reasonable. I will close this PR and divide changes  into multiple PRs with better explanation why changes are needed.

PR #267, opened 2014-07-19
support blob on native level

revitalyr:


mloskot:
If there are no objections, I'd like to split this PR into two and first merge only the blob support.
It needs to be updated with test cases as well.

I have no opinion about the C++11 tuple support, motivation is unclear to me.

BTW, there has been [similar request on soci-devel](http://sourceforge.net/p/soci/mailman/message/34276729/).

OniDaito:
Blob for sqlite3 is a pretty cool and handy thing (UTF-8 and the like). Yeah, not sure about the tuple stuff. Might look into that

vadz:
I think this was implemented a long time ago, at least there is blob (`dt_blob`) support in SQLite3 backend and the unit test "SQLite blob" passes.

So I guess this PR is obsolete, but please correct me if I'm wrong.

PR #277, opened 2014-10-31
function describe_column changed to get all column properties via column_properties in the row class

cthun:
1. fix odbc_statement_backend::describe_column
fix handling for numeric columns -> return dt_integer, dt_long_long or
dt_double (merged from oracle backend)

2.  get all column properties during describe call
class column_properties enhanced to store informations about colSize,
decDigits and isNullable flag of a column
use a pointer of class describe_column for the describe_column function
call (be able to use the informations outside of SOCI)

Because i had to change all db-backends there is still an open task to add the informations from the backend to the column_properties class. Currently only ODBC and Oracle are adding all informations to the pointer. The other backends only add name and type.


vadz:
Unfortunately I don't think this PR can be realistically expected to be merged, there are too many (apparently completely unrelated) changes here at least some of which are seriously questionable and it's not clear at all what is the actual benefit, to the library users, of the others.

I'd like to encourage you to submit several PRs with individual changes and ideally slightly more explanations about why are they done.

TIA!

mloskot:
**[The smaller the PR, the better!](https://github.com/SOCI/soci/wiki/Contributing)**

PR #281, opened 2015-01-04
Update version.h

denisarnaud:
SOCI_VERSION and SOCI_LIB_VERSION are expected to reflect the same SOCI version, more specifically 3.2.2 here. So, I guess that it was a typo.


vadz:
Let's just change the version to 3.2.3 instead (done in 6e9312c4bb3748907bd28d62c40feca42878cfef), like this the current master sources will have a different version from the 3.2.2 release which is useful IMHO as it allows the code to distinguish between them.

PR #283, opened 2015-01-29
nested transaction support in sqlite

MonsieurNicolas:
sqlite doesn't follow the standard when it comes
to nested transactions and therefore needs special handling


MonsieurNicolas:
replaced by https://github.com/SOCI/soci/pull/288

PR #286, opened 2015-03-07
Adding cmake SOCI_SHARED option and lowercasing windows.h

bingmann:
This fixes one basic issue: sometimes one does not want to build the shared libraries.
And an issue when cross-compiling with MinGW on Linux: apparently all headers are lower-cased.


vadz:
Thanks, merged into master but not to the develop branch as the cmake commit can't be trivially cherry picked into it (the `windows.h` issue had been already fixed there).

bingmann:
Thanks. I'll put merges onto develop in future.

PR #288, opened 2015-03-10
Nested transaction support

MonsieurNicolas:
Added support for nesting transactions.
I only included configuration I could test:
* constructed SQL using the most common syntax for nesting transactions (using savepoints)
* sqlite3, postgresql and mysql seems fine
* some backends seem to not support it at all (ODBC layer for example)
* Oracle seems to support this syntax, but I don't have a way to test this

replaces the previous PR I had https://github.com/SOCI/soci/pull/283

### DIFF
--- a/src/backends/db2/session.cpp
+++ b/src/backends/db2/session.cpp
@@ -127,7 +127,7 @@ db2_session_backend::~db2_session_backend()
     clean_up();
 }
 
-void db2_session_backend::begin()
+void db2_session_backend::begin(const char* /*beginTx*/)
 {
     // In DB2, transations begin implicitly; however, autocommit must be disabled for the duration of the transaction
     if(autocommit)
@@ -145,7 +145,7 @@ void db2_session_backend::begin()
     in_transaction = true;
 }
 
-void db2_session_backend::commit()
+void db2_session_backend::commit(const char* /*commitTx*/)
 {
     if (!autocommit || in_transaction) {
         in_transaction = false;
@@ -168,7 +168,7 @@ void db2_session_backend::commit()
     }
 }
 
-void db2_session_backend::rollback()
+void db2_session_backend::rollback(const char* /*rollbackTx*/)
 {
     if (!autocommit || in_transaction) {
         in_transaction = false;
--- a/src/backends/db2/soci-db2.h
+++ b/src/backends/db2/soci-db2.h
@@ -234,9 +234,9 @@ struct db2_session_backend : details::session_backend
 
     ~db2_session_backend();
 
-    void begin();
-    void commit();
-    void rollback();
+    virtual void begin(const char* beginTx);
+    virtual void commit(const char* commitTx);
+    virtual void rollback(const char* rollbackTx);
 
     std::string get_backend_name() const { return "DB2"; }
 
--- a/src/backends/db2/test/test-db2.cpp
+++ b/src/backends/db2/test/test-db2.cpp
@@ -96,6 +96,17 @@ public:
     {
         return "to_date('" + pi_datdt_string + "', 'YYYY-MM-DD HH24:MI:SS')";
     }
+
+    bool supportsTransactions() const
+    {
+        return true;
+    }
+
+    bool supportsNestedTransactions() const
+    {
+        return false;
+    }
+
 };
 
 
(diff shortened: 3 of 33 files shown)


vadz:
Thanks for the update, but apparently DB2 doesn't like this -- while Firebird simply doesn't compile...

MonsieurNicolas:
updated,

* the db2 tests were not respecting the fact that a "commit" must have a corresponding "begin" statement (at the SOCI layer). I am not sure why this was allowed before as it clearly makes the code non portable.

* in Firebird - similar issue with the addition of the compile issue that I missed.

MonsieurNicolas:
No idea why Firebird is failing here:
exception would be caused by a "commit" or "rollback" called at the wrong time but I don't see either test1 doesn't even use transactions or explicit begin/commit, so there seems to be some implicit commit done in the session destructor (but I don't see the session's commit being invoked from the backend's destructor).
I can't setup Firebird on my system to debug this so I am going to abandon this PR for now.

vadz:
I've tested it here and the problem is due to `commit()` being called before `begin()` in `TableCreator1` (and also `2` and `3`), see [the code](https://github.com/SOCI/soci/blob/develop/tests/assert/firebird/test-firebird.cpp#L1242). It is there since almost the very beginning (see 58c602e8), but still seems wrong to me.

After fixing this, the tests pass. However normally Firebird does [support nested transactions](http://www.firebirdsql.org/refdocs/langrefupd15-savepoint.html), so I tried enabling the corresponding test but this gave an error:
```
soci_firebird_test: .../soci/tests/assert/common-tests.h:1911: void soci::tests::common_tests::test10nested(): Assertion `count == 4' failed.
```
so something still seems to be wrong.

I also am not really sure if we want to break the existing code by mandating calling `begin()` before `commit()` anyhow. It looks like Firebird starts an implicit transaction on connection, so calling `begin()` could remain optional. What do you think?

MonsieurNicolas:
Cool, thanks.

For the failing test it looks like while Firebird supports nested transactions, the implementation of the backend session.begin/commit doesn't. See https://github.com/MonsieurNicolas/soci/blob/master/src/backends/firebird/session.cpp#L300

As far as trying to keep the implicit begin:
in order to communicate the fact that the backend starts a transaction by itself, it would need a way to tell that to the core's session object. Right now, there is no such back pointer. From a layering point of view it seems cleaner to not try to add this back dependency.

Maybe a different way would be to create a firebird transaction that lasts for the duration of a statement if there is no current transaction? The performance would be terrible for people not following proper begin/commit semantics, so failing like it does now might be still be better.

vadz:
> For the failing test it looks like while Firebird supports nested transactions, the implementation of the backend session.begin/commit doesn't

Yes, I didn't realize that the patch doesn't actually implement this for Firebird (I saw that it touched its files and assumed it did, but I was just wrong). FWIW it looks like there are API function which allow us to do everything we need, see [their documentation](http://docwiki.embarcadero.com/InterBase/XE3/en/Working_with_Savepoints).

In fact, after reading the above and other databases documentation, I think the save points are actually the right mechanism to use for all of them. Save points seem to be supported by all the RDBMS (see https://en.wikibooks.org/wiki/SQL_Dialects_Reference/Transactions), unlike the "real" nested transactions which don't work at all as expected in MS SQL Server, for example (which is not directly supported by SOCI but we use it extensively via the ODBC backend), where rolling back any nested transaction rolls back the outermost one.

Also, concerning the implicit `begin()` in Firebird backend, I've just done something that should make it work more consistently with the other backends, see 53cd24ecc80605ebe80e00dd6ac5381aa3cb6e77. I'm not sure if this really changes anything for this discussion, but wanted to mention it in case it does.

MonsieurNicolas:
sorry for not pushing this PR forward - we're quite a bit behind master at this point.
I can try to rebase this if you're interested?
Problem on my end is that I only have access to sqlite and postgresql, so I can only try things out via the CI pipeline.

Would you be OK with me just revving this to work for sqlite, postgresql and mysql? I can make sure that other backends preserve the existing behavior.
SOCI already has a few features that are not supported on all backends already. I am worried that other backends like firebird have weird quirks that will break existing code (but pass all the tests we can think of).

As far as code that mixes direct SQL statements  "BEGIN"/"COMMIT"/"ROLLBACK" and session.begin/commit/rollback I think it's acceptable to break it.

vadz:
It's not a problem to do something that only works with some specific backends, but IMHO it would be nice to have some way of testing whether such features do or do not work.

I also still think that using savepoints might be more fruitful, as mentioned in my previous comment.

PR #296, opened 2015-03-19
Use a base-class member instead of shadowing it.

nbougalis:
The class `ref_counted_prepare_info` shadows a member variable of its parent to the same thing: it keeps a redundant reference to a `session`.

Remove the redundant reference and use the one from the parent instead.


vadz:
This looks correct to me, but presumably should be merged into develop branch and not master.

nbougalis:
@vadz I can reopen it against develop if you'd prefer.

vadz:
@nbougalis  Yes, please. TIA!

PR #298, opened 2015-03-26
Convert the tests to use CATCH with minimal changes.

vadz:
This is not supposed to be merged yet as the changes are not finished.

Comments about them are already welcome though.


mloskot:
Responded with my comments in the thread [RFD: Using CATCH for tests](http://sourceforge.net/p/soci/mailman/message/33640616/)

vadz:
I think this can be merged now. A lot of things could be improved and I'll try to do at least some of them in the future commits, but for now I'd like to add more tests and this means committing this first.

mloskot:
:+1:

vadz:
Merged now.

PR #302, opened 2015-04-01
Provide error context in the exceptions.

vadz:
Include information about the failed query and its parameters, if applicable, in the exception object itself.


soyersoyer:
The soci_error's copy constructor and copy assigment operator shouldn't throw.
https://www.securecoding.cert.org/confluence/display/cplusplus/ERR13-CPP.+Do+not+allow+an+exception+class%27s+copy+constructor+to+throw+exceptions

vadz:
@soyersoyer good point, thanks for noticing.

Issue #303, opened 2015-04-09
Merge master 3.2.3 into develop


mloskot:
We have released 3.2.3.
The current `master` branch now hosts the snapshot of that release.
The current `develop` branch is being behind that release, because the recent changes in `master` have not been merged into `develop`.
The task is to merge the `master` so we don't loose fixes and improvements released in 3.2.3

I'm currently working on the merge. The merge is not an easy one due to structure changes and numerous conflicts.

It would be helpful if we could limit commits into `develop` for a couple of days, thanks.

vadz:
Sorry to hear about the merge conflicts, I think they may be also due to the fact that some commits were applied to both branches already.

Have you tried using [git-imerge](https://github.com/mhagger/git-imerge)? IME it's very helpful with non trivial merges.

mloskot:
@vadz No, I haven't tried git-merge, but looks interesting. Meanwhile, I've resolved all the conflicts and now it's matter of making the Travis CI green, see #304

mloskot:
Done.

PR #304, opened 2015-04-09
Merging master (SOCI 3.2.3) into develop

mloskot:
This is a 'walking' PR following #303 as attempt merge of the master with current SOCI 3.2.3 changes into develop.

All the conflicts have been resolved.
Further changes, if any, should be only related to fixing broken builds.


vadz:
Would be nice to avoid reintroducing the trailing whitespace that I removed in develop.

Otherwise I looked over the changes and noticed just a raw `&` in an HTML file which should be `&lt;`, all the rest looks good, thanks!

mloskot:
@vadz Thank you for the review. I've restored the trailing whitespace removal and corrected the `&`.

Assuming we're happy with this PR, I'll do `git co develop && git merge develop-3.2.3`. OK?

mloskot:
FYI, I used the old fashion command line way.

PR #243, opened 2014-03-31
Add a Travis build using PostgreSQL server at Postgression.

vadz:
Use the test server from http://www.postgression.com/ to test with another
(latest, it seems), PostgreSQL version in addition to the one available on
Travis.

Closes #167.

Issue #167, opened 2013-06-26
Set up postgression service for Travis CI


mloskot:
This task is about enabling PostgreSQL test on http://www.postgression.com, where [apparently](https://groups.google.com/d/msg/travis-ci/v36cHwvMJnw/3ta3h1uxdBYJ) we could test against different versions of PostgreSQL.

vadz:
Access to Postgression servers either doesn't to work at all from Travis or is too slow, so apparently this just can't be done.

PR #306, opened 2015-04-10
Add postgression service to Travis CI setup

mloskot:
This is my take on Vadim's #243 to add address #167 and add Travis build using PostgreSQL server at Postgression.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -14,29 +14,18 @@ services:
   - postgresql
 
 env:
-  - SOCI_TRAVIS_BACKEND=db2
-  - SOCI_TRAVIS_BACKEND=empty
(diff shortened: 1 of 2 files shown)


mloskot:
@vadz I've tweaked your original PR to make it work (see comments in `bin/ci/script_postgression.sh`). It took me a few iterations, then I squashed all the *work-in-progress* commits and updated this PR.

Thanks for working on that.

vadz:
I wonder if it's time to revert this because the service seems to never be available any longer (victim of their own success, I guess, but it doesn't change the fact that it's unusable and just wastes Travis CI resources).

mloskot:
Yes, good idea.
Let me know if you want me to do it

PR #314, opened 2015-04-13
Updated FindMySQL.cmake to work under 64-bit windows.

rpiotaix:
I don't know why, but it seems that $ENV{ProgramFiles} was referring to "Program Files (x86)" on 64-bit windows.

In addition, I refactored the two 'IF(WIN32)' into one by appending all the suggested paths to find libmysql (or mysqllclient) into one list.


mloskot:
Thanks!

mloskot:
Merged into the master too as db59a48ce45240cf9d67aa53835428fa2a825910

PR #315, opened 2015-04-14
Reorganize "Use and into" unit test to use CATCH sections.

vadz:
This makes the test slightly shorter (the line creating the table doesn't need
to be repeated many times) and, more importantly, allows us to specify the
descriptions which can be useful in case of test failures.


mloskot:
:+1: for mastering CATCH, great!

PR #316, opened 2015-04-15
Small cleanup remaining from the attempt to fix timezone handling

vadz:
As I wrote in #190, I finally don't think it's right to call `timegm()`, but I could be wrong and if we do decide to do it, it would be much simpler to change it if it can be done in one place only, so I think it would make sense to commit this refactoring which I had done in order to call `timegm()` -- even if we don't do anything else, it still makes the code slightly tidier IMHO.


mloskot:
I like it. Thanks  :+1:

PR #317, opened 2015-04-15
Make SOCI_NORETURN public and use it in declaration.

vadz:
SOCI_NORETURN had to be moved to a public header to avoid MSVC error C2381
("'function' : redefinition; __declspec(noreturn) differs") which was given if
SOCI_NORETURN was specified on the function definition only and not its
declaration.

Use SOCI_NORETURN when both declaring and defining the function now (notice
that C++11 [[noreturn]] attribute also must be used on declarations).

### DIFF
--- a/include/private/soci-compiler.h
+++ b/include/private/soci-compiler.h
@@ -36,12 +36,4 @@
 #   define GCC_WARNING_RESTORE(x)
 #endif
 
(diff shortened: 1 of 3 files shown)


mloskot:
:+1:

vadz:
Merged, thanks for the review.

PR #318, opened 2015-04-16
Use parameter names from query in the error messages.

vadz:
The important commit is the last one, the other two just prepare for it.


mloskot:
Nice feature.

(note to self: think of testing such features too)

vadz:
I didn't add a test for this because it would fail for Oracle... Of course, I could just test for the backend being used and skip the test then, but if we could implement `get_parameter_name()` for Oracle too, it would be even better. I don't have enough forces to tackle OCI right now though.

mloskot:
Right, another reason to get the Oracle setup. I'll work on it next week.

PR #319, opened 2015-04-16
Avoid warnings about int-to-pointer conversions in DB2 backend.

vadz:
These warnings are harmless as the cast to pointer needs to be used just
because SQLSetStmtAttr() is "polymorphic" and takes both pointers and
integers, but annoying, so avoid them by using a helper cast-like function.

### DIFF
--- a/include/soci/db2/soci-db2.h
+++ b/include/soci/db2/soci-db2.h
@@ -46,6 +46,17 @@ namespace soci
             BOUND_BY_NAME,
             BOUND_BY_POSITION
         };
+
+        inline SQLPOINTER int_as_ptr(int n)
+        {
+            union
+            {
+                SQLPOINTER p;
+                int n;
+            } u;
+            u.n = n;
+            return u.p;
+        }
     }}
 
     static const std::size_t maxBuffer =  1024 * 1024 * 1024; //CLI limit is about 3 GB, but 1GB should be enough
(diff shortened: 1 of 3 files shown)


mloskot:
Looks OK.

AFAIR from most of ODBC drivers I've seen, in SQL_ATTR_ROW_ARRAY_SIZE case, SQLSetStmtAttr casts it back to `SQLINTEGER` or `SQLBIGINT`.

vadz:
I actually don't understand why I don't get the same warnings for ODBC... Maybe the ODBC headers define `SQLPOINTER` differently?

mloskot:
AFAIU, `SQLSetStmtAttr` signature should be exactly the same in both. Perhaps ODBC headers disable that warning.

vadz:
OK, I solved the mystery: the warning is given in DB2 code when casting from (32 bit) `int` to (64 bit) `SQLPOINTER`, but not in ODBC code because it casts from `SQLULEN` which is also 64 bit.

Not sure if it's worth changing the code, I might do it if we run into this warning under some other platform in the future, but for now let's leave it as is.

mloskot:
Right. Indeed, cast to 64-bit integer is what, for instance, SQLite ODBC driver does too:

```
static SQLRETURN
drvsetstmtattr(SQLHSTMT stmt, SQLINTEGER attr, SQLPOINTER val,
	       SQLINTEGER buflen)
{
    STMT *s = (STMT *) stmt;
#if defined(SQL_BIGINT) && defined(__WORDSIZE) && (__WORDSIZE == 64)
    SQLBIGINT uval;

    uval = (SQLBIGINT) val;
#else
    SQLULEN uval;

    uval = (SQLULEN) val;
#endif
...
```

PR #320, opened 2015-04-20
Use SQLT_BDOUBLE instead of SQLT_FLT in Oracle backend.

vadz:
This is recommended by Oracle documentation for performance reasons as it
avoids conversion from the native double to Oracle format and back but, more
importantly, avoids rounding problems and allows the "Repeated and bulk fetch"
unit test to pass for doubles whereas it failed previously because the double
value of 0.6*3 was written into the database as exactly 1.8, which is
different from IEEE 754 representation.

Notice that this commit assumes that Oracle 10.1+ is used during run-time if
10.1+ headers are used for compiling SOCI. This is not true in general, but is
hopefully good enough in practice -- and if it isn't, run-time Oracle version
checking will need to be added later.

### DIFF
--- a/CHANGES
+++ b/CHANGES
@@ -10,6 +10,9 @@ Version 4.0.0 differs from 3.2.x in the following ways:
 - Firebird
 -- Add SOCI_FIREBIRD_EMBEDDED option to allow building with embedded library.
 
+- Oracle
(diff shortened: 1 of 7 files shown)


vadz:
NB: I've tested that the unit tests pass with Oracle (at least 11.2 I tested with) now.

mloskot:
:+1:

Issue #300, opened 2015-03-30
Get rid of asserts in the code


vadz:
Either _compile_ time assertions or exceptions should be used to report errors and not the `assert()` macro which is probably more harmful than useful (application shouldn't necessarily crash just because some database operation failed) and can easily result in ODR violations when it is used in headers, as is the case for a few of them in SOCI.

PR #321, opened 2015-04-20
Remove all asserts from SOCI.

vadz:


include/soci/rowset.h
```diff
@@ -41,10 +41,6 @@ class rowset_iterator
     rowset_iterator(statement & st, T & define)
         : st_(&st), define_(&define)
     {
-        assert(0 != st_);
```
mloskot: While I'm convinced by the reasoning in #300:
> application shouldn't necessarily crash just because some database operation failed
I'm not sure removal of assertions checking the class invariants like this and the following one fall into this category. The `rowset_iterator` must not contain nullptr reference to statement or define and that assertion documents this fact.

vadz: I actually spent some time reading the code before making this change as it was indeed not completely obvious that it wasn't needed, but I did convince myself about it finally:

 1. `rowset_iterator` is only constructed by `rowset::begin()` and `end()`, with the latter using the default ctor (we could enforce this by making the non default ctor private probably).
 1. `rowset::begin()` just forwards to `rowset_impl::begin()`.
 1. `rowset_impl::begin()` uses ``rowset_impl::st_` which is known to never be `NULL`.

So IMHO there is no point in keeping this assert as it's tautologically true.

But, in any case, my reasoning is always the same: if we think we must guard against the pointer being null/empty here, let's do it -- but just not using `assert()` but by throwing an exception. Do you think it's worth adding the checks here and throwing if they fail?

mloskot: > rowset_impl::begin() uses `rowset_impl::st_ which is known to never be NULL.
--------------------------------------------------------------------^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^

That's exactly why the assertion is there.

> IMHO there is no point in keeping this assert as it's tautologically true.

That exactly my point. An assertion in code documents that fact explicitly, so reader does not have to go through the code to come to those 1-2-3 conclusions.

Those particular assertions I'm discussing tell those pointers mustn't be nullptr. It could be a comment, comments come and go, whereas assertion has higher severity.

>  if we think we must guard against the pointer being null/empty here, let's do it -- 
> but just not using assert() but by throwing an exception

I completely agree.
It's just that those assertions don't do that here. They don't guard against nullptr. They just explicitly document those pointers mustn' be null there.

IOW, if `assert(0 != st_);` is not fulfilled and causes a program to abort, then even if there is no assertion here, the program should abort or crash anyway. Throwing exception there would cause invalid questions: so, there can be a `rowset` object without statement attached, can't there?

I'm not arguing to keep those assertions, they can go.
I'm just surprised those assertions are considered harmful to end users (if they can, they will be harmed without those assertions anyway).

vadz: I have 2 problems with using asserts: first one is the clear potential for the ODR violations when they're used in headers. Granted, usually this will be harmless, but sometimes it can result in really weird and hard to debug problems, so, after having encountered such problems one time too many, UI now try to avoid any possibility of ODR violations in the first place.

The second objection is more fundamental and is that different people use asserts for different things and you never know what they're for. Your use of them for documentation purposes only is one common case but another common case is for enforcing the preconditions and this is very different. Personally I think they are inappropriate for both of these purposes: when you see a pointer being dereferenced in the code, it's a clear documentation of the fact that it's not supposed to be used already and, anyhow, documentation shouldn't be executable code. As for preconditions, they should be handled more explicitly and, notably, always and not just when `NDEBUG` happens to be undefined.

mloskot:
@vadz I understand your points. No need for 
As I mentioned, I'm not objecting the proposed changes, just asking about them. 

From my point, I vote +0 :), so feel free merge it.

PR #325, opened 2015-04-21
C++11 Standard

OniDaito:
Moving from C++98 to C++11. 

General changes include:
- removing auto_ptr, swapping with unique_ptr.
- Changing areas of the CMakeLists
- Warnings as Errors (because it's neater and good style)
- Unfixed warnings documented where necessary (in relation to the above)
- Currently, boost::optional seems to be causing problems with catch.hpp testing so I've overridden it for now
- Missing library dependencies for the static tests

### DIFF
--- a/cmake/SociBackend.cmake
+++ b/cmake/SociBackend.cmake
@@ -189,6 +189,11 @@ macro(soci_backend NAME)
           ${THIS_BACKEND_SOURCES}
           ${THIS_BACKEND_HEADERS})
 
+        target_link_libraries(${THIS_BACKEND_TARGET_STATIC}
+          ${SOCI_CORE_TARGET}
+          ${THIS_BACKEND_DEPENDS_LIBRARIES})
+
+
         set_target_properties(${THIS_BACKEND_TARGET_STATIC}
           PROPERTIES
           OUTPUT_NAME ${THIS_BACKEND_OUTPUT_NAME}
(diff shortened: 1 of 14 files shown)


vadz:
Personally I need to continue using SOCI with C++98 so for me this change is not acceptable, C++11 should be an option, not hard coded as it is now.

I also don't believe it's a good idea to add `-Werror` only to also add pragmas suppressing the warnings in the entire file(s), I'd rather try to understand where do the individual warnings come from and leave them as a reminder until this can be done.

OniDaito:
Can always manuover the pragmas closer to where they need to be. That would be best. It makes things explicit - can be fixed later but at least recognised, which they werent previously. This is not my suggestion but one in C++ Coding Standards (Sutter, Alexandrescu), and one I agree with.

C++11 could be an option, though I suppose that would involve pragmas here and there. Can do that instead.

vadz:
Yes, using pragmas locally would be much better. BTW notice that these pragmas are gcc 4.6+ only, see `include/private/soci-compiler.h` for my workaround allowing to use them without breaking compilation for earlier gcc version -- although it won't really help if we use `-Werror`.

As for C++11 it simply must be an option, as much as I'd like to leave C++98 behind, it's just not always possible.

OniDaito:
Well, I can't comment on the C++98 except to say I sympathise ;) 
I shall make these changes and add it as an option, though it would be nice to have that as the default.

mloskot:
@OniDaito FYI, I agree with @vadz - we can't drop C++98 yet.

If we could receive C++11 support as an option, that would be a very helpful first step. We could also add C++11 mode builds to our Travis CI matrix.

OniDaito:
That all sounds very positive. We ourselves (at Queen Mary University) have a strong need for SOCI (packages as well as C++11 and such) and we are keen to give back the bits we've done (such as they are). I'll redo and re-commit later today.

mloskot:
@OniDaito I'm happy to hear. Thanks!

Issue #224, opened 2014-01-07
Install Oracle on Travis CI server


mloskot:
A per [Recipes for connecting to external database servers ( Oracle, SQL Server)](https://groups.google.com/d/msg/travis-ci/D_0oKVF4eaI/a7DVfMCvSDcJ) thread on Travis CI list:

> however no external Oracle installation is required, see: https://github.com/p6spy/p6spy/tree/master/script/travis

P6Spy folks adopted original SOCI configuration extending it, so it should be trivial to port their improvements into SOCI.
Kudos to @p6spy!

PR #326, opened 2015-04-22
Restore Oracle setup on Travis CI server

mloskot:
Solution for task #224 will be either based on p6spy/p6spy#116 or https://github.com/cbandy/travis-oracle/


vadz:
It looks like `ORACLE_HOME` is set incorrectly (for 11 and not 10)?

mloskot:
Vadim, thank you for the extra pair of eyes.

I hope it's been corrected now.

mloskot:
There is one more issue to solve: SOCI build vs OCI binaries architecture.
Summary:
* Travis is hosted on Ubuntu 64-bit
* Oracle 10g is installed from 32-bit binaries
* Travis builds SOCI 64-bit binaries.
* Problem: SOCI 64-bit attempt to link against OCI 32-bit binaries

Possible solutions:
1. Switch `SOCI_TRAVIS_BACKEND=oracle` to 32-bit build. It seems **easier**.
2. Install InstantClient 64-bit and use it to build SOCI, but still connect to the installed Oracle 10g 32-bit. It seems **tricky** - can we switch ORACLE_HOME to different locations, first for installation, second for build, will it work?

mloskot:
The option 1. discussed above seems to work, partially.
The problem is CMake finds 64-bit boost, even if 32-bit build is requested.

mloskot:
Holy cow, it's green now!

Summary: **64-bit Travis CI builds 32-bit SOCI without Boost but with Oracle.**

mloskot:
It was green for a moment, then somewhat mysteriously it stopped working due to ORA-12520 errors.

Apparently, SOCI tests require higher value of Oracle XE's PROCESSES parameter. After bumping it from default 40 to value from range of 100-200 it's green again.

PR #327, opened 2015-04-23
C++11 Standard Support and improvement of some pragmas

OniDaito:
As discussed previously, here is another version that has C++11 as an option rather than hard-coded. It basically sets a preprocessor macro.

In addition, testing with the latest branch has removed many of the compiler warnings on Linux. Unsure about OSX Clang at present. Will need to test that too in the near future. I've removed the pragmas for that but kept warnings as errors. Under GCC 4.8.2 it compiles cleanly. 

Not sure which versions of GCC cause the errors with the tests. I've gone for anything greater than 4.6. I'll need to investigate further.

### DIFF
--- a/tests/common-tests.h
+++ b/tests/common-tests.h
@@ -3659,18 +3659,8 @@ void run_query_transformation_test(test_context_base const& tc, session& sql)
         CHECK(count == 'z' - 's');
     }
 
-// Bug in Visual Studio __cplusplus still means C++03
-// https://connect.microsoft.com/VisualStudio/feedback/details/763051/
-#if defined _MSC_VER && _MSC_VER>=1600
-#define SOCI_HAVE_CPP11 1
-#elif __cplusplus >= 201103L
-#define SOCI_HAVE_CPP11 1
-#else
-#undef SOCI_HAVE_CPP11
-#endif
-
-#ifdef SOCI_HAVE_CPP11
-    // lambda
+#if 0
+    // lambda is just presented as an example to curious users
     {
         sql.set_query_transformation(
             [](std::string const& query) {
@@ -3682,7 +3672,6 @@ void run_query_transformation_test(test_context_base const& tc, session& sql)
         CHECK(count == 'j' - 'h');
     }
 #endif
-#undef SOCI_HAVE_CPP11
 
     // prepared statements
 


mloskot:
[Build 608](https://travis-ci.org/SOCI/soci/builds/59719738) indicates we may need to install later GCC on Travis CI:
```
cc1plus: error: unrecognized command line option ‘-std=c++11'
```

Travis CI is based on Ubuntu 12.04 with GCC 4.6.3 as default version, whereas option `-std=c++11` [requires 4.7 or later](https://gcc.gnu.org/projects/cxx0x.html).

Unless we can't or don't want to downgrade the option to `-std=c++0x`, I will need to look for the compiler update.

OniDaito:
Ah yes, I noticed that the first time around. I did wonder where they c++0x came from. Happy to help in any way. I've not worked too much with Travis, as our internal build system here is Jenkins.

jsonn:
Many projects like LLVM consider 4.7 to be the bare minimum for usable C++11. Both due to compiler bugs and incompletely runtime.

OniDaito:
In that case we should get another travis build matrix up and running with 4.7 or above methinx. Might as well do it properly and go the whole way instead of middling if we do it at all.

mloskot:
@OniDaito in your [OniDaito/soci/blob/master/.travis.yml](https://github.com/OniDaito/soci/blob/master/.travis.yml#L9), please change 
```
compiler:
  - g++
```
to
```
compiler:
  - clang
```
and update the PR with this change.

Travis CI also provides clang 3.1, so hopefully it's enough to test your PR.
Once it's tested, I'll merge it and think how to add the C++11 build based on GCC 4.8 to the matrix on Travis CI.

OniDaito:
Ok hopefully that last commit should fix that I think. I've had to add a few things here and there. Im testing with clang version 3.4-1ubuntu3 on Ubuntu Trusty

mloskot:
It's almost green :)

Could you try to add this line in [bin/ci/before_install.sh](https://github.com/OniDaito/soci/blob/master/bin/ci/before_install.sh#L11), after the line 10:
```
apt-get install libstdc++5 libstdc++5:i386
```
AFAIS, Ubuntu 12.04 does have those packages, http://packages.ubuntu.com/precise/libstdc++5

OniDaito:
Hopefully thats got it?

mloskot:
Sorry for delays.
The mixture of Oracle 32bit vs 64bit host vs target architectures is still a troublemaker.
I'm hoping to clear soon.

cmake/SociConfig.cmake
```diff
@@ -9,6 +9,13 @@
 ################################################################################
 
 #
+# C++11 Flag
+#
+
+set(CXX_C11 ON CACHE BOOL "Build to the C++11 Standard as oppose to GNUC++98.")
```
mloskot: Two things:
* CXX_C11 mode needs to be off by default
* CXX_C11 is better renamed to SOCI_CXX_C11, a purely SOCI's option

cmake/SociConfig.cmake
```diff
@@ -28,18 +35,22 @@ if (MSVC)
 
 else()
 
-  set(SOCI_GCC_CLANG_COMMON_FLAGS
-	"-pedantic -ansi -Wall -Wpointer-arith -Wcast-align -Wcast-qual -Wfloat-equal -Wredundant-decls -Wno-long-long")
+  set(SOCI_GCC_CLANG_COMMON_FLAGS "-pedantic -Wall -Werror -Wpointer-arith -Wcast-align -Wcast-qual -Wfloat-equal -Wredundant-decls -Wno-long-long")
+
+  if (CXX_C11)
+    set(CMAKE_CXX_FLAGS "${CMAKE_CXX_FLAGS} -std=c++11")
+    add_definitions(-DCXX_C11)
```
mloskot: CXX_C11 is better renamed to SOCI_CXX_C11 a purely SOCI's def

OniDaito:
Sounds good to me :)

mloskot:
@OniDaito I'm trying to manually merge this pull and I noticed you have submitted it from the master, instead of a topic branch. The problem is, the master has been updated with the "First stab at markdown documentation" and related commits. So, the pull request has been 'polluted' with unrelated commits :-(  The merge is tricky, so , I'll have o cherry pick.
It would be helpful if you create topic branches next time.

OniDaito:
Ah sorry about that. I was under the false impression it would merge at the point of submission. How about you hold fire for now and what I'll do is go back to the last commit before I carried on, branch out and pull the latest changes you guys have made? Would that make life easier for you? :)

mloskot:
@OniDaito that would be helpful indeed. I'll wait then. Once I merge this C++11 changes, I'm going to work on the new docs.

mloskot:
Re-worked and merged as part of #336

Issue #287, opened 2015-03-10
Big std::strings in nvarchar(max)s via ODBC fails with HY104


Superlokkus:
Hi, soci throws odbc_soci_error when the std::string I use is very long (4096 worked but failed at 29906).
The ODBC error code is HY104, so I guess soci does not handle big strings right (to be fair due typical ODBC driver implementation limitations), or I overlooked the way to handle them with soci. The according column is defined as nvarchar(max), the server is an MS SQLServer 2008RC2 and the ODBC-Driver uses the SQL Server Native Client 11.0 on Windows 7 Prof.

vadz:
If the driver doesn't handle strings of this size, I don't think SOCI should do anything about it. Try using BLOBs instead of the strings instead perhaps?

Superlokkus:
@vadz AFAIK is supposed to work see https://msdn.microsoft.com/de-de/library/ms130896.aspx

mloskot:
The sole reference of [SQLBindParameter parameter ColumnSize](https://msdn.microsoft.com/en-us/library/ms710963.aspx) does not say it all.

AFAIU, in case a large object exceeds certain range (see below), then the `ColumnSize` is not set with actual size of the object, but with special constant. See [Binding Parameters for SQL Character Types](https://msdn.microsoft.com/en-us/library/ms131462.aspx):
>  If the SQL data type passed in is a character type, ColumnSize is the size in characters (not bytes). If the length of the data string in bytes is greater than 8000, ColumnSize should be set to SQL_SS_LENGTH_UNLIMITED, indicating that there is no limit to the size of the SQL type.
>
> For instance, if the SQL data type is SQL_WVARCHAR, ColumnSize should not be greater than 4000. If the actual data length is greater than 4000, then ColumnSize should be set to SQL_SS_LENGTH_UNLIMITED so that nvarchar(max) will be used by driver.

I may be misunderstanding, so it would be good if someone could confirm that.

Currently, `SQL_SS_LENGTH_UNLIMITED` case is not handled.

vadz:
I could try testing this, but what exactly is the test case here? I.e. does it happen when writing a long value to the database or when reading it from it?

Superlokkus:
Writing

mloskot:
Yeah, it's always helpful if a bug report is accompanied with (pseudo)code sample or, preferably, [CATCH](https://github.com/philsquared/Catch/blob/master/docs/tutorial.md)-based test case.

[...]

PR #328, opened 2015-04-24
Fix inserting strings longer than 8000 bytes with ODBC/MS SQL.

vadz:
Use SQL_LONGVARCHAR and SQL_SS_LENGTH_UNLIMITED for such strings, otherwise
they were silently truncated when inserting them into an nvarchar(max) column.

See #287.


mloskot:
:+1:  @vadz Awesome, thanks for light-speed fix!

vadz:
Committed as c025cc86968914bd4bd0523dee11cdd9db991aba

PR #333, opened 2015-05-07
Fix compilation and test errors (Win+VS2015)

tvb377:
I found these changes were necessary to successfully compile and test (empty, sqlite3) soci with VS2015 Preview and boost 1.58.


tvb377:
I don't know your policy for supporting older compilers I tried to be conservative here:
I replaced SOCI_HAVE_CPP11 by SOCI_HAVE_LAMBDA to be consistent with the particular SOCI_HAVE_NOEXCEPT definition and to be open for future conditional usage of certain C++11 features.
At least for supporting older MSVC versions while introducing C++11 it is necessary to split up the enabling / disabling of C++11 features. I think only VS2015 is (or will be) very close to completely supporting C++11.
So when soci is supporting e.g. the most recent three MSVC versions there is no such thing as "SOCI_HAVE_CPP11" ;-)

Since I just started using soci, I don't know yet if more than these two SOCI_HAVE_<C++11Features> would be sensible to introduce.

mloskot:
@tvb377 No real policy exists, apart from the fact we are not switching over to C++11 (or later) compilers only (as explained in comments to #327).
Your explanation is sufficient to me.

I give :1 and if no objections arrive any time soon, I'll merge.

Thank you for the contribution!

include/soci/once-temp-type.h
```diff
@@ -10,8 +10,9 @@
 
 #include "soci/ref-counted-statement.h"
 #include "soci/prepare-temp-type.h"
+#include "soci/soci-platform.h"
 
-#if __cplusplus >= 201103L
+#if SOCI_HAVE_NOEXCEPT
```
vadz: For consistency we should use `#ifdef` for testing `SOCI_HAVE_XXX` (see the existing `HAVE_BOOST` and `SOCI_HAVE_CPP11`.

include/soci/soci-platform.h
```diff
@@ -8,6 +8,26 @@
 #ifndef SOCI_PLATFORM_H_INCLUDED
 #define SOCI_PLATFORM_H_INCLUDED
 
+// Check some C++11 Features
+#if     defined(_MSC_VER)
+// Bug in Visual Studio: __cplusplus still means C++03
+// https://connect.microsoft.com/VisualStudio/feedback/details/763051/
+#define SOCI_HAVE_NOEXCEPT (_MSC_VER >= 1900)
+#define SOCI_HAVE_LAMBDA   (_MSC_VER >= 1600)
+#elif   defined(__clang__) || defined(__GNUG__)
```
vadz: I think this could be omitted, we should just test for `__cplusplus` directly, there are other C++11 compilers than gcc/clang (e.g. Intel, Oracle).

vadz:
Thanks for the changes, they look good to me but ideally I'd just get rid of `SOCI_HAVE_LAMBDA` completely: it's used in a single place in a test and it would be simpler to just use a helper function/class there than bother with defining this symbol and testing for it.

`SOCI_HAVE_NOEXCEPT` for MSVS is definitely a good addition though.

mloskot:
On 8 May 2015 at 03:19, VZ <notifications@github.com> wrote:

> I'd just get rid of SOCI_HAVE_LAMBDA completely: it's used in a single
> place in a test and it would be simpler to just use a helper function/class
> there than bother with defining this symbol and testing for it.
>

Do you mean to replace the lambda passed into
`sql.set_query_transformation` with function/functor?
That would be a different check.

include/soci/soci-platform.h
mloskot: On 8 May 2015 at 03:15, VZ <notifications@github.com> wrote:

> In include/soci/soci-platform.h
> <https://github.com/SOCI/soci/pull/333#discussion_r29911015>:
>
> > @@ -8,6 +8,26 @@
> >  #ifndef SOCI_PLATFORM_H_INCLUDED
> >  #define SOCI_PLATFORM_H_INCLUDED
> >
> > +// Check some C++11 Features
> > +#if     defined(_MSC_VER)
> > +// Bug in Visual Studio: __cplusplus still means C++03
> > +// https://connect.microsoft.com/VisualStudio/feedback/details/763051/
> > +#define SOCI_HAVE_NOEXCEPT (_MSC_VER >= 1900)
> > +#define SOCI_HAVE_LAMBDA   (_MSC_VER >= 1600)
> > +#elif   defined(__clang__) || defined(__GNUG__)
>
> I think this could be omitted, we should just test for __cplusplus
> directly, there are other C++11 compilers than gcc/clang (e.g. Intel,
> Oracle).
>

Valid point, indeed, unless the author has got a good reason to avoid
__cplusplus and list compilers explicitly.

tvb377: Line 65 in SociConfig.cmake throws a FATAL_ERROR when using a compiler other than clang/gcc/vc. That is reflected by this change.
And I didn't even know that Oracle has a C++ compiler. Neither do I know if __cplusplus is defined correctly for compilers other than MSVC,gcc and clang nor do I know if CMake reports this compiler as being something else nor if all other compilers have implemented noexcept.
That's why I said I was trying to be conservative.
Unless you know these compilers well and you are about to remove this line 65 in SociConfig.cmake I'd keep the change like I proposed and leave the necessary changes to a user of other compilers.
When you simply refer to __cplusplus when not using MSVC you might force users of untested other compilers exactly into the work I had to exercise: Why is the simple test program abort()ing right away. There is no message like "you shouldn't throw in a destructor". The message reads "abort() called".
When you error out for unknown compilers then you're pointing their users directly to the important stuff to consider.
I'd favor the latter annoyance over than tracing through the test code.

mloskot: On 8 May 2015 at 11:16, tvb377 <notifications@github.com> wrote:

> Line 65 in SociConfig.cmake throws a FATAL_ERROR when using a compiler
> other than clang/gcc/vc.
>

This is SociConfig.cmake caveat which should be fixed at some point.
(Slowly, I try to remove as much as possible of our *.cmake customisations
and rely on CMake built-ins).


> Neither do I know if __cplusplus is defined correctly for compilers other
> than MSVC,gcc and clang nor do I know if CMake reports this compiler as
> being something else nor if all other compilers have implemented noexcept.
>

We don't support compilers that don't define or define incorrectly the
standard predefined macro __cplusplus. So, we can safely rely on it.


> Unless you know these compilers well and you are about to remove this line
> 65 in SociConfig.cmake I'd keep the change like I proposed and leave the
> necessary changes to a user of other compilers.
>

I've removed the FATAL_ERROR restriction, see
1d26033e060c2108d4ffbe06bbe4a6895c24bdc7.
It was too restrictive indeed. Actually, it was more like a left-over from
early days of CMake configuration development.
Thanks for pointing that.

vadz:
Sorry, I didn't realize that we already had a test for using a function and a function object. In this case I'd just remove the test with lambda, what's the point of it? It is testing compiler lambda support rather than SOCI because for a conforming compiler it's exactly the same as the tests above.

mloskot:
@vadz I just remembered, the whole code in test was copied from example sent to a user showing all possibilities. I've removed/disabled it 2bf3aa7ed33c88945265ba7be4933c5e9720ad56

snikulov:
@mloskot @vadz @tvb377 using link_directories according to documentation is not necessary.
Link issues with msvc related to fact that auto-link conflicts with library, which is provided with full path to it. And it solves just with disabling Boost auto-link feature (ref #389 ).
Also it not required to introduce DEPS_LIBRARY_DIRS and make things more complicated.

mloskot:
@snikulov Do you suggest this whole PR can be closed as no longer relevant, or just the CMake parts? Thanks!

snikulov:
@mloskot let's investigate this after integrate #391 :)
At present time AppVeyor-CI does no complain. So it can be Boost issues involved.

tvb377:
I'm not familiar with the appveyor build system, but when I look at https://ci.appveyor.com/project/snikulov/soci it seems to me that soci doesn't build.
Looking at the build messages from some minutes ago, you might want to enable C++11 (relevant at least for g++), and use a proper boost installation.

snikulov:
@tvb377 https://ci.appveyor.com/project/snikulov/soci/build/4.0.0.38 here the branch with appveyor.yml 
MinGW is broken, but I working on it with #390 
HTH.

snikulov:
@tvb377 Could you please try latest master and check if it works for you?

@mloskot I suggest to close this issue, and if something will be wrong for @tvb377 create another issue.

tvb377:
Head revision of soci works for me.
Thanks!

Issue #334, opened 2015-05-17
Mac OS: master branch fails to build on Yosemite


dunn:
As of 2bf3aa7ed33c88945265ba7be4933c5e9720ad56

Build logs for: 
- [GCC](https://gist.github.com/dunn/414baf26442d9fa9c699)
- [Clang](https://gist.github.com/dunn/60fecfb4ded71e1aee54)

Stable release 3.2.3 from https://downloads.sourceforge.net/project/soci/soci/soci-3.2.3/soci-3.2.3.tar.gz works fine.

May be related to #258

vadz:
Yes, this is exactly the same as #258 and I don't think it's useful to keep both of them open. FWIW you could try manually adding `--undefined-suppress` to `LDFLAGS` (sorry, I don't know how to do it with Cmake but I dare hope it should be possible) and check if it fixes the problem.

mloskot:
@dunn Please, check fix proposed to #258

Issue #258, opened 2014-06-11
Compilation failure on branch develop


pfedor:
Clean client. On master compiles without problems.

```
pfedor@Pawels-iMac:~/soci$ rm -rf build
pfedor@Pawels-iMac:~/soci$ mkdir build
pfedor@Pawels-iMac:~/soci$ cd build
pfedor@Pawels-iMac:~/soci/build$ cmake -G "Unix Makefiles" -DWITH_BOOST=OFF -DMYSQL_INCLUDE_DIR=/opt/local/include/mysql55/mysql/ -DMYSQL_LIBRARIES=/opt/local/lib/mysql55/mysql/libmysqlclient.dylib ../soci/
-- The C compiler identification is Clang 5.1.0
-- The CXX compiler identification is Clang 5.1.0
-- Check for working C compiler: /usr/bin/cc
-- Check for working C compiler: /usr/bin/cc -- works
-- Detecting C compiler ABI info
-- Detecting C compiler ABI info - done
-- Check for working CXX compiler: /usr/bin/c++
-- Check for working CXX compiler: /usr/bin/c++ -- works
-- Detecting CXX compiler ABI info
-- Detecting CXX compiler ABI info - done
-- Configuring SOCI: 
-- SOCI_VERSION                             = 3.2.2 
-- SOCI_ABI_VERSION                         = 3.2 
-- SOCI_PLATFORM_NAME                       = x86 
-- SOCI_COMPILER_NAME                       = gcc-4.2.1 
-- SOCI_STATIC                              = ON 
-- SOCI_TESTS                               = ON 
-- Looking for SOCI dependencies: 
-- Threads: 
-- Looking for include file pthread.h
-- Looking for include file pthread.h - found
-- Looking for pthread_create
-- Looking for pthread_create - found
-- CMAKE_THREAD_LIBS_INIT                   =  
-- Boost: disabled, since WITH_BOOST=OFF 
-- MySQL: 
-- Performing Test HAVE_MYSQL_OPT_EMBEDDED_CONNECTION
-- Performing Test HAVE_MYSQL_OPT_EMBEDDED_CONNECTION - Success
-- Found MySQL: /opt/local/include/mysql55/mysql, /opt/local/lib/mysql55/mysql/libmysqlclient.dylib
-- MySQL Embedded not found.
-- MYSQL_INCLUDE_DIR                        = /opt/local/include/mysql55/mysql 
[...]

PR #335, opened 2015-05-18
Add -Wl,-flat_namespace -Wl,-undefined -Wl,suppress to LINK_FLAGS on Apple/OSX

mloskot:
Addresses compilation issues on OSX, see #258 and dup #334

### DIFF
--- a/cmake/SociBackend.cmake
+++ b/cmake/SociBackend.cmake
@@ -172,6 +172,12 @@ macro(soci_backend NAME)
             PROPERTIES
             SOVERSION ${${PROJECT_NAME}_SOVERSION}
             INSTALL_NAME_DIR ${CMAKE_INSTALL_PREFIX}/lib)
+
+          if(APPLE)
+            set_target_properties(${THIS_BACKEND_TARGET}
+              PROPERTIES
+              LINK_FLAGS "-Wl,-flat_namespace -Wl,-undefined -Wl,suppress")
+          endif()
         endif()
 
         set_target_properties(${THIS_BACKEND_TARGET}


OniDaito:
Happy to take a look. I've been working a little with the OSX version.

dunn:
Still failing to build; can't locate `error.h`:
- [GCC](https://gist.github.com/dunn/33401f01716951288e84)
- [Clang](https://gist.github.com/dunn/b7cddcecc45f54f3c97c)

mloskot:
Can you erase line `#include "error.h"` in `src/backends/sqlite3/error.cpp` and check if it builds?

It seems like a bug indeed, becasue there is no `error.h` header specific to the SQLite3 backend.

It is unclear, why this issue didn't leak earlier, on Linux for instance.
*  There is only public header `include/soci/error.h` which, if used, should be included usnig qualified path `#include <soci/error.h>`, it this header is not required in SQLite3.
* Another possibility is that `include/soci/error.h` is unexpectadely included  due to misconfigured `-I` paths, but as I see [in this log](https://gist.githubusercontent.com/dunn/33401f01716951288e84/raw/89fa77d7bf1e90b258c0fb610be2f8474d429754/02.make) for example, this is not the case:
```
-I/tmp/soci20150518-44411-1tywz1v
-I/tmp/soci20150518-44411-1tywz1v/include
-I/tmp/soci20150518-44411-1tywz1v/include/private
```
there is no `-I/tmp/soci20150518-44411-1tywz1v/include/soci`

Regardless, AFAIS, `#include "error.h"` is not necessary in `src/backends/sqlite3/error.cpp`.

mloskot:
It sesems the mystery behind successful builds on Linux has been solved: `#include "error.h"` directive in `src/backends/sqlite3/error.cpp` was including the GNU C Library `/usr/include/error.h`.

dunn:
New errors!
- [Clang](https://gist.github.com/dunn/662fc0a002fa59d84cd5#file-02-make-L372)
- [GCC](https://gist.github.com/dunn/aebed168a9c14ed3ac44#file-02-make-L374)

mloskot:
Apparently, there is names conflict between 
```
using namespace soci;
session sql;
```
and `struct session` defined in `/usr/include/sys/proc.h` on OSX.

@vadz You don't happen to have OSX to confirm that? Shall we `soci::` qualify names in tests?

vadz:
Qualifying names with `soci::` would definitely work but would be rather ugly, it would be nice if things could be made to work without it. I wonder where is `sys/proc.h` get included from and if there is any way to avoid it (`gcc -H` could be used to discover it).

mloskot:
I think [the logs](https://gist.github.com/dunn/aebed168a9c14ed3ac44#file-02-make-L374) we've already got pin-point source:
```
In file included from /usr/include/sys/sysctl.h:84:0,
                 from /tmp/soci20150518-55914-sm7vru/tests/catch.hpp:6752,
                 from /tmp/soci20150518-55914-sm7vru/tests/empty/test-empty.cpp:14:
/usr/include/sys/proc.h:82:8: note: candidates are: struct session
 struct session;
        ^
```

vadz:
Ah, thanks, I didn't notice this.

Unfortunately now that we know where it is included from, I still have no idea how to fix it, the obverse face of having a header-only library is that anything used in its implementation also pollutes the global namespace and this is exactly what happens here.

The only horrible hack I see is to

    // Avoid problems with "session" being defined in sys/proc.h (indirectly) included by CATCH under OS X
    #define session session_system
    #include "catch.hpp"
    #undef session

but I'm not really sure if it's less ugly than using `soci::` everywhere...

mloskot:
@vadz I'd simply replace `session` with `soci::session`. I can do it tonight.

mloskot:
@dunn Could you give #341 patch a try and see if it solves the problems for you?

PR #341, opened 2015-05-19
Fully qualify uses of session class with namespace soci::session

### LINKED ISSUE #340: OSX: struct session conflicting with unqualified soci::session
As revealed on the course of solving build issues in #335, on Apple/OSX, there is names conflict between
```
using namespace soci;
session sql;
```
and struct session defined in /usr/include/sys/proc.h:
```
In file included from /usr/include/sys/sysctl.h:84:0,
                 from /tmp/soci20150518-55914-sm7vru/tests/catch.hpp:6752,
                 from /tmp/soci20150518-55914-sm7vru/tests/empty/test-empty.cpp:14:
/usr/include/sys/proc.h:82:8: note: candidates are: struct session
 struct session;
        ^
```

The uses of `session` type need to be qualified `soci::session`.

mloskot:
This should solve compilation errors on OSX where `struct session` defined in global namespace in `/usr/include/sys/proc.h` conflicts with unqualified uses of `soci::session`.

Fixes #340

### DIFF
--- a/tests/common-tests.h
+++ b/tests/common-tests.h
@@ -450,7 +450,7 @@ TEST_CASE_METHOD(common_tests, "Exception on not connected", "[core][exception]"
 
 TEST_CASE_METHOD(common_tests, "Basic functionality")
 {
-    session sql(backEndFactory_, connectString_);
+    soci::session sql(backEndFactory_, connectString_);
 
     auto_table_creator tableCreator(tc_.table_creator_1(sql));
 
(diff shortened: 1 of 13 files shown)


dunn:
Looks like a few more need to be disambiguated: https://gist.github.com/dunn/12291204d3790fd00822#file-02-make-L422

mloskot:
@dunn Thanks for checking. I've just updated the PR with additional commit.

dunn:
Yes, it builds now!

mloskot:
Awesome, thanks!

PR #343, opened 2015-05-22
Minor C++11 Pragma bugfix and Inlining C++11 test pragmas

OniDaito:
There was one #pragma I forgot about, like an idiot! :S  And, in response to your advice, I've inlined the pragmas from the common-tests.h file.


mloskot:
Nice, thanks!

PR #344, opened 2015-05-22
Markdown Documentation

OniDaito:
Here is the first stab at the markdown documents. They are likely not complete or perfect yet, but here they are for comment at least.


mloskot:
Awesome effort. I'll work on it next week.

I've just been learning GitHub Pages what made me considering it for SOCI website. The only missing part would be the API Reference - we could generate Markdown files with API reference using [cldoc](http://jessevdk.github.io/cldoc/). Then all the final building of website would be delegated to gh-pages. Unless, I'm missing something else.

OniDaito:
Sounds about right though I'm not really familiar with gh-pages or cldoc. I'll need to look into it a bit more

mloskot:
Let's welcome the new docs in.

The next stage will be of setting GitHub pages for the SOCI organization and deploying the docs there as well.

ravselj:
FYI. Today I noticed(while cloning soci) that this commit caused drastical increase in repository size.

OniDaito:
I'm guessing thats because the original docs are still there for now? At some point they will be deleted I imagine? In addition, we need to go over the markdown docs and give them some copy-editing love.

mloskot:
@ravselj thanks for pointing this out.

AFAIS, `.git/objects/pack/pack-0ec603f6f99a311385bf63353bb68140934b68c4.pack` has grown a lot.

@OniDaito IMO, we can delete the old docs now and even do `git rename doc docs`. We can do editing/corrections afterwards.

mloskot:
Anyone brave enough to clean up the reflog :)
I guess this is what we need to run: http://stackoverflow.com/a/24693485/151641

vadz:
Not sure what exactly do you want to do, normally just re-cloning the repo should be enough, AFAIK reflog doesn't exist in the bare repository on the server.

mloskot:
@vadz I should have added "in their local clonse". Sorry for confusion. 

Re-clonning should fix it indeed.

PR #346, opened 2015-05-28
Updated classic Makefiles for Oracle Express 11.2.

msobczak:
This pull request brings the classic Makefils up-to-date with the Oracle Express 11.2 (in fact, it should be version-agnostic) edition and fixes some missing bits that accumulated during the recent evolution of SOCI and that were not taken into account in the Makefiles.


vadz:
Do we still use these makefiles at all? I had no idea they had to be maintained to be honest...

mloskot:
On 28 May 2015 23:52, "VZ" <notifications@github.com> wrote:
>
> Do we still use these makefiles at all? I had no idea they had to be
maintained to be honest...

Maciej created those makefiles as he prefers to use them. So, he was taking
care of the maintenance.

mloskot:
+1 feel free to merge

msobczak:
Please note that this patch contains the include/private/soci_backends_config.h file - this file is #included from sources, but does not exist in the repository. My understanding is that it is generated by CMake, but it should be confirmed - and that this generation is not obscured by the presence of this file.

As for the need for these Makefiles - please note that they provide the shortest and at the same time the most exact (and verifiable by execution) documentation of (external) dependencies. CMake is great in that it automates some things, but at the same time it hides them, making it more difficult for others to integrate the whole source tree with other build systems. You might not need them, but I prefer to keep them for reference.

mloskot:
[https://github.com/SOCI/soci/blob/48a8bfab0e4a735bcd623a4d03071bd7536aa822/src/core/CMakeLists.txt#L19-L21 src/core/CMakeLists.txt generates] uses [https://github.com/SOCI/soci/blob/cf76df6c404314df2986130ac1de28bf97090fcb/src/core/soci_backends_config.h.in src/core/soci_backends_config.h.in] as input to generate this header. 

CMake will still generate the header, but in build directory. So, there will be two versions of the headers and order of includes might matter. I can't check it myself, but I'd worry about it after merge..

msobczak:
The patch was corrected - the file in question is now regenerated by src/core/Makefile.basic, which means that it is not needed in the patch. Now the only files that are modified by the patch are 3 Makefiles, which should be safe with regard to the remaining part of the repository.

vadz:
I'll take the liberty of merging it then, it doesn't seem useful to just keep it sitting here.

Issue #349, opened 2015-06-18
Mysql backend memory leaks due to mysql_library_init()


thibault-martinez:
Hi,

It seems that mysql_init() implicitly calls mysql_library_init().
Unfortunately mysql_close() doesn't calls mysql_library_end(), resulting in 2 memory leaks.

You should explicitly call mysql_library_end() in your mysql_session_backend::clean_up() function.
https://github.com/SOCI/soci/blob/master/src/backends/mysql/session.cpp#L405

I'll do a pull request to add it.

Regards,

mloskot:
Regarding "it seems", you likely refer to the remarks from the MySQL docs:

> In a nonmulti-threaded environment, mysql_init() invokes mysql_library_init() automatically as necessary. However, mysql_library_init() is not thread-safe in a multi-threaded environment, and thus neither is mysql_init(). Before calling mysql_init(), either call mysql_library_init() prior to spawning any threads, or use a mutex to protect the mysql_library_init() call. This should be done prior to any other client library call.

PR is welcome!

thibault-martinez:
Absolutely !

The need to call mysql_library_end() is well explained in the [Mysql C API Function Overview](https://dev.mysql.com/doc/refman/5.0/en/c-api-function-overview.html)
> The purpose of calling mysql_library_init() and mysql_library_end() is to provide proper initialization and finalization of the MySQL library. For applications that are linked with the client library, they provide improved memory management. If you do not call mysql_library_end(), a block of memory remains allocated. (This does not increase the amount of memory used by the application, but some memory leak detectors will complain about it.) For applications that are linked with the embedded server, these calls start and stop the server.

avidboarder:
It looks like a call to `mysql_thread_end()` is missing from `mysql_session_backend::clean_up()`
See https://dev.mysql.com/doc/refman/8.0/en/mysql-thread-end.html

PR #350, opened 2015-06-18
Fix issue #349 Mysql backend memory leaks due to mysql_library_init()

thibault-martinez:
See issue #349

### DIFF
--- a/src/backends/mysql/session.cpp
+++ b/src/backends/mysql/session.cpp
@@ -403,6 +403,7 @@ void mysql_session_backend::clean_up()
     if (conn_ != NULL)
     {
         mysql_close(conn_);
+        mysql_library_end();
         conn_ = NULL;
     }
 }


vadz:
I don't know much about MySQL, but is this really correct if there is more than one connection alive? I would naively expect that `mysql_library_end()` should only be called when there are no more of them left...

thibault-martinez:
I think you're right, only the last remaining mysql backend should call `mysql_library_end()`...
Is there any way to know if other backends are still working ?

mloskot:
@thibault-martinez
> Do you have any internal way to know if other backends are still alive ?

There is no way to know it.
We can't use `std::atomic` we still support pre-C++11 compilers.
A counter internally  managed by `mysql_session_backend` sounds like a good choice.
However, it needs to take the [connection_pool](http://soci.sourceforge.net/doc/3.2/multithreading.html) use cases into account.

b1c9g76:
I saw that on high volume use this log message was appearing with this patch:

Error in my_thread_global_end(): 2 threads didn't exit

vadz:
I'm finally going to merge this, better late than never and all this. Thanks!

vadz:
Merged as 78c5a313 (not sure why GitHub didn't notice this automatically), thanks again!

PR #351, opened 2015-06-24
CMAKE changes

ravselj:
- MSVC linker failed when building with SHARED option( fix by setting parent scope )
- added additional options for MSVC( MP flag for faster compilation, some MSVC warnings suppressed)


vadz:
As usual, I'm not qualified to judge CMake-specific changes (i.e. the "SCOPE" thing), but if others don't see anything wrong with it, I could merge this. I do know that using `/MP` and avoiding the spurious deprecation warnings is definitely a good idea (I usually also define `_CRT_NONSTDC_NO_WARNINGS` to allow the use of POSIX functions).

The only suggestion I have could be to split this commit in two parts as they seem quite independent.

mloskot:
Thanks for the patch.
+1 for splitting

ravselj:
Ok. I'll split this into two PRs. Makes sense.

vadz:
One PR is fine, I think, I was just suggesting to split it into 2 commits, e.g. to allow reverting or backporting or whatever each of them separately later if needed.

TIA!

ravselj:
I haven't used commit splitting before so I had to take some time to understand it. Anyway, I have finally managed to split into two commits. Hopefully it will be ok.

CMakeLists.txt
```diff
@@ -86,6 +86,15 @@ if(NOT MSVC)
     set_directory_properties(PROPERTIES INCLUDE_DIRECTORIES ${DL_INCLUDE_DIR})
     add_definitions(-DHAVE_DL=1)
   endif()
+else()
+  option(MSVC_USE_MP "Set to ON to build with the /MP option (Visual Studio 2005 and above)." ON)
```
mloskot: Do we really need a configurable option?
Why not just always add /MP for VS2005 or later?

src/core/CMakeLists.txt
```diff
@@ -31,6 +31,7 @@ source_group("CMake Files" FILES CMakeLists.txt)
 
 # Core targets configuration
 string(TOLOWER "${PROJECT_NAME}" PROJECTNAMEL)
+set(SOCI_CORE_TARGET ${PROJECTNAMEL}_core PARENT_SCOPE) #set in parent scope so backends will also see this variable
```
mloskot: It is OK

CMakeLists.txt
```diff
@@ -86,6 +86,15 @@ if(NOT MSVC)
...
+  mark_as_advanced(MSVC_USE_MP)
+  if(MSVC_USE_MP)
+    set(CMAKE_CXX_FLAGS "${CMAKE_CXX_FLAGS} /MP")
+  endif(MSVC_USE_MP)
+  # MSVC compilation flags suppressing compilation warnings
+  add_definitions(-D_SCL_SECURE_NO_WARNINGS)
```
mloskot: It is OK, but IMO, it's better if warnings are disabled in source code using `#pragma warning`.
So, the source code is independent and portable from a build system. 
Build systems can be changed and the same problem is back, unless it's solved at the source code level.

mloskot:
Regardless my comments, I'm OK to merge.
@ravselj unless, you want to rework the patch according to suggestions, then feel free to kill this PR and open new one.
Thanks!

ravselj:
@mloskot you have a point. It's better not to mix apples and oranges so I am closing this PR. I will open new one shortly.

PR #352, opened 2015-06-25
Set CMake variable SOCI_CORE_TARGET at PARENT_SCOPE

ravselj:
@mloskot as agreed in #351 I'm opening another PR.

CMake bug fix when building with MSVC with SHARED option enabled. 

SOCI_CORE_TARGET was not visible to backend so VS projects were not generated correctly producing linker errors.


mloskot:
Thanks!

PR #353, opened 2015-06-25
MSVC warnings removal and fixes

ravselj:
I have suppressed all warnings reported during Visual Studio build process using pragmas(as suggested by @mloskot in #351 ). CMake remains unmodified.

### DIFF
--- a/include/private/firebird/common.h
+++ b/include/private/firebird/common.h
@@ -19,6 +19,11 @@
 #include <vector>
 #include <algorithm>
 
+#if defined(_MSC_VER)
+#pragma warning(disable:4127)
+#endif
+
+
 namespace soci
 {
 
@@ -65,7 +70,7 @@ const char *str2dec(const char * s, IntType &out, int &scale)
         int d = *s - '0';
         if (d < 0 || d > 9)
             return s;
-        res = res * 10 + d * sign;
+        res = res * 10 + (IntType)(d * sign);
         if (1 == sign)
         {
             if (res < out)
@@ -101,7 +106,7 @@ template<typename T1>
 void to_isc(void * val, XSQLVAR * var, int x_scale = 0)
 {
     T1 value = *reinterpret_cast<T1*>(val);
-    short scale = var->sqlscale + x_scale;
+    short scale = var->sqlscale + (short)x_scale;
     short type = var->sqltype & ~1;
     long long divisor = 1, multiplier = 1;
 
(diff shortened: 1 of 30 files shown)


src/backends/firebird/row-id.cpp
```diff
@@ -10,6 +10,10 @@
 
 using namespace soci;
 
+#if defined(_MSC_VER)
+#pragma warning(disable:4702)
```
vadz: This ("unreachable code") can be a pretty useful warning, it would be nice to avoid disabling it if possible but rather work around it in the place where it occurs.

include/private/firebird/common.h
```diff
@@ -19,6 +19,11 @@
 #include <vector>
 #include <algorithm>
 
+#if defined(_MSC_VER)
+#pragma warning(disable:4127)
```
vadz: This warning ("conditional expression is constant") is pretty useful in general, it would be better to avoid disabling it.

vadz:
Sorry, I don't think I agree with this approach. In addition to the comments above, I'd also prefer to continue defining `_CRT_SECURE_NO_DEPRECATE` and friends in the code instead of disabling C4996 completely: it could still be useful if, e.g. we deprecate something in SOCI itself one day, as we would want to be warned if it were still used inside SOCI itself.

vadz:
P.S. And I'd also prefer to use `static_cast<>` rather than C style casts for the casts where they are needed.

mloskot:
On 26 June 2015 at 00:21, VZ <notifications@github.com> wrote:

> P.S. And I'd also prefer to use static_cast<> rather than C style casts
> for the casts where they are needed.
>
Yes, me too.

mloskot:
On 26 June 2015 at 00:20, VZ <notifications@github.com> wrote:

> Sorry, I don't think I agree with this approach. In addition to the
> comments above, I'd also prefer to continue defining
> _CRT_SECURE_NO_DEPRECATE and friends in the code instead of disabling
> C4996 completely: it could still be useful if, e.g. we deprecate something
> in SOCI itself one day, as we would want to be warned if it were still used
> inside SOCI itself.
>

Disabling in SOCI itself was not my concern at all, so I suggested in-code
solution rather than via build configuration.
Ideally, if all warnings could be addressed by replacing deprecated
functions with recommended counterparts, wherever possible.

ravselj:
The comments above make sense so I have try different approach here. 
I will close this PR and reopen another one shortly.

PR #355, opened 2015-06-30
Msvc warnings

ravselj:
I created a new branch that tries to achieve same goal - removing MSVC warning while compiling(based on comments in #353 )

Some notes are also in commit messages but I will only describe the basic approach. All warnings are resolved either by:
1. changing the code directly resolving the cause( static_cast, remove unused variable )
2. pragma push, disable, pop ( this is used only once )
3. Platform specific changes in soci-platform.h( without changing any cpp code) either by:
  - pragma warning(disable: *) 
  - using defines (sprintf -> sprintf_s )
  - using defines and platform specific inline wrapper functions(  strcpy -> strcpy_secure )
4. Implemented global wrapper functions with different signature in soci namespace( currently soci::getenv, soci::localtime ) and replaced old calls with new ones. ( i.e. where getenv or std::getenv is called the call is replaced with soci::getenv ). This is necessary since function signature is not the same.

I have removed all pragma warning(disable) from all cpp files. Suppressed warnings are reduced to only two warnings which should be ignored (4251 4275). All other warnings were removed as described above. 

I have also deleted soci-config.h and moved its content into soci-platform.h. There is no need for both headers since they both contain platform dependent code so I have merged them into one header.


include/soci/soci-platform.h
```diff
@@ -18,13 +25,63 @@
...
+    {
+        errno_t err = strncpy_s(buf,count,src,count);
+        if( err != 0 )
+            return 0;
+        return buf;
+    }
+
+    inline char* strcpy_secure(char* dest, const char* src)
+    {
+        errno_t err = strcpy_s(dest,strlen(src)+1,src);
```
vadz: Sorry but this really doesn't make sense. If you unconditionally assume that the destination buffer is large enough to contain the source one, using `strcpy_s()` is just snake oil and is actually much worse than just using `strcpy()` in the first place.

vadz:
I'm not entirely convinced about the wisdom of using MSVC wrappers, as the example of `strcpy()` shows this is not completely trivial and I'd prefer to have one version of all these functions to deal with instead of two.

I'm also not sure about merging `soci-platform.h` and `soci-config.h`, while I'm not sure about the difference between them neither, there might be existing code including the latter and I don't think it's worth breaking it by removing this header.

Also, while adding casts to silence the warnings is nice, they can sometimes be fixed in better ways as I noticed in the very last change of this PR. Unfortunately the sheer number of changes makes it difficult to review all of them...

include/soci/soci-platform.h
ravselj: The goal was to remove msvc warnings without suppression. Of course it is insecure the same way as standard strcpy is.

Looks like this can only be resolved correctly by defining _CRT_SECURE_NO_WARNINGS. Would that be better?

mloskot: On 8 July 2015 at 10:57, ravselj <notifications@github.com> wrote:

> Looks like this can only be resolved correctly by defining
> _CRT_SECURE_NO_WARNINGS. Would that be better?
>

If you decide to use the preprocessor define(s), please add them to .h/cpp
files wherever necessary,
before any headers importing any 'insecure' function(s).
But, not to CMake or any build configuration.

ravselj:
>I'm not entirely convinced about the wisdom of using MSVC wrappers, as the example of strcpy() shows this is not completely trivial and I'd prefer to have one version of all these functions to deal with instead of two.

OK. I get that. Then the only solution to deprecated would be to define 

#define _CRT_SECURE_NO_WARNINGS in soci-platform.h and then we must make sure that soci-platform is the first header included. This way all MSVC wrappers will be removed.

>I'm also not sure about merging soci-platform.h and soci-config.h, while I'm not sure about the difference between them neither, there might be existing code including the latter and I don't think it's worth breaking it by removing this header.

If I add back soci-config.h with only one line including soci-platform.h with comment //this header is obsolete - use soci-platform instead. This way we preserve all client code and all specific platform code is in one header. Would this be ok?

>Also, while adding casts to silence the warnings is nice, they can sometimes be fixed in better ways and I'd like to review the changes to make sure this is not the case here but unfortunately the sheer number of changes makes it difficult to review all of them...

If you look at the first commit you'll see only static_cast changes. There aren't that many actually but I agree some of them can be resolved also without using static_cast. This can probably best be solved in separate PR since it is not directly related to this PR. Would you agree?

mloskot:
@vadz 
> I'm also not sure about merging soci-platform.h and soci-config.h, while I'm not sure about the difference between them neither, there might be existing code including the latter and I don't think it's worth breaking it by removing this header.

It is a good idea to keep both, but perhaps we should officially deprecate one or another with relevant `pragma` message.

Regarding the rest of the comments, I agree with @vadz.

include/soci/soci-platform.h
ravselj: @mloskot, @vadz 
I have two choices here:
1. use defines in every cpp and h where  warning is emitted.
2. define _CRT_SECURE_NO_WARNINGS globally in soci-platform and then include soci-platform.h first

I think second one is better since it results in cleaner code. What do you think?

mloskot: On 8 July 2015 at 12:06, ravselj <notifications@github.com> wrote:oskot
<https://github.com/mloskot>, @vadz <https://github.com/vadz>
>
> 2. define _CRT_SECURE_NO_WARNINGS globally in soci-platform and then
> include soci-platform.h first
>
> I think second one is better since it results in cleaner code. What do you
> think?
>

IMO, yes, the 2. is a good idea as well as convention to include
soci-platform.h first.

Please, wait for @vadz final opinion as he may see issues I can't - I'm not
as concerned about
breaking users code as Vadim, but I respect that, so Vadim's got final word
here.

vadz: FWIW I am perfectly fine with defining `_CRT_SECURE_NO_WARNINGS` in `soci-platform.h`.

vadz:
@ravselj I agree with your proposals, maybe we could reuse `soci-config.h` later for non-platform-specific but still variable stuff (like whether to use C++11 or C++14 or C++17?), but for now keeping it as just a wrapper for `soci-platform.h` would be perfectly fine. And yes, we probably should commit the casts for now and maybe/hopefully whittle them down later as time allows.

TIA!

ravselj:
I have commited some changes related to our conversation:
- msvc wrappers removed
- soci-config.h restored ( wrapper for soci-platform.h )
- soci-platform.h
- I have also removed some static_cast by replacing a few code( changing to proper type instead of explicit casting ).

include/private/firebird/common.h
```diff
@@ -98,14 +98,15 @@ double round_for_isc(double value)
...
-void to_isc(void * val, XSQLVAR * var, int x_scale = 0)
+void to_isc(void * val, XSQLVAR * var, short x_scale = 0)
 {
     T1 value = *reinterpret_cast<T1*>(val);
     short scale = var->sqlscale + x_scale;
     short type = var->sqltype & ~1;
     long long divisor = 1, multiplier = 1;
 
-    if ((std::numeric_limits<T1>::is_integer == false) && scale >= 0 &&
+    static bool is_integer = std::numeric_limits<T1>::is_integer;
```
vadz: This should be `const` in addition to/rather than `static`.

vadz: P.S. Although to be honest, I don't think it makes sense to really factor it out, I'm almost certain that the compiler inlines this in the optimized build anyhow.

include/private/soci-cstrtod.h
```diff
@@ -9,6 +9,7 @@
 #define SOCI_PRIVATE_SOCI_CSTRTOD_H_INCLUDED
 
 #include "soci/error.h"
+#include "soci/soci-platform.h"
```
vadz: Why is this needed here now?

include/private/soci-dtocstr.h
```diff
@@ -8,6 +8,7 @@
 #ifndef SOCI_PRIVATE_SOCI_DTOCSTR_H_INCLUDED
 #define SOCI_PRIVATE_SOCI_DTOCSTR_H_INCLUDED
 
+#include "soci/soci-platform.h"
```
vadz: And here?

include/soci/bind-values.h
```diff
@@ -58,6 +58,8 @@ class use_type_vector: public std::vector<use_type_base *>
 
         use_type_vector &p;
         Indicator &ind;
+    private:
+        use_sequence& operator=(const use_sequence&);
```
vadz: I think it might be preferable to use a `SOCI_NOT_COPYABLE(classname)` macro because relatively soon we will want to have a deleted (in C++11 sense) assignment operator instead of undefined private one (which is just a poor workaround for the really deleted one) and a macro would make it easier to use different expansions for C++98 and C++11.

include/soci/db2/soci-db2.h
```diff
@@ -201,7 +201,7 @@ struct SOCI_DB2_DECL db2_statement_backend : details::statement_backend
 
     int prepare_for_describe();
     void describe_column(int colNum, data_type& dtype, std::string& columnName);
-    std::size_t column_size(int col);
+    SQLUINTEGER column_size(int col);
```
vadz: This makes it inconsistent with the ODBC and Oracle versions, should the other 2 be changed as well? If not, I'd rather leave this one alone as well and keep the casts, in SQL/ODBC code they are admittedly unavoidable sometimes...

include/soci/soci-platform.h
```diff
@@ -8,6 +8,18 @@
 #ifndef SOCI_PLATFORM_H_INCLUDED
 #define SOCI_PLATFORM_H_INCLUDED
 
+//disable MSVC deprecated warnings
+#ifdef _MSC_VER
+#define _CRT_SECURE_NO_WARNINGS
```
vadz: Could you please add `#ifndef _CRT_SECURE_NO_WARNINGS` here to avoid redefining it if it's already predefined in the project file (we do define it already on ours)?

src/backends/firebird/row-id.cpp
```diff
@@ -10,12 +10,18 @@
 
 using namespace soci;
 
+#ifdef _MSC_VER
+#pragma warning(push)
+#pragma warning(disable:4702) //unreachable code
```
vadz: FWIW it looks like this warning could be avoided by not having `firebird_rowid_backend` class at all and just throwing directly from `firebird_session_backend::make_rowid_backend()`.

vadz:
I don't have any other comments, thanks! If @mloskot doesn't have any other ones neither, I think the "casts" part is good to go. I'd prefer to use a macro instead of manually defining private assignment operators though to make transition to C++11 simpler in the future.

include/private/soci-cstrtod.h
ravselj: It's not. This remained from previous commit. I'll remove it.

include/private/soci-dtocstr.h
ravselj: Same as above.

include/soci/bind-values.h
ravselj: OK.

include/soci/db2/soci-db2.h
ravselj: There is a diference in underlying API call. I probably better to revert this and use static_cast.

src/backends/firebird/row-id.cpp
ravselj: I agree.

include/private/firebird/common.h
ravselj: This is workaround for MSVC C4127 ( constant value in conditional expression ). If if use const (static or non static ) the warning remains an then it is no difference.

Instead of using temporary static variable we could resolve this by using templates at compile time( this way also runtime if is not necessary):

```c++
template<bool cond> struct cond_to_isc {};
template<> struct cond_to_isc<false> 
{
    static void check(short scale, short type)
    {
        if( scale >= 0 && (type == SQL_SHORT || type == SQL_LONG || type == SQL_INT64) )
            throw soci_error("Can't convert non-integral value to integral column type");
    }
};
template<> struct cond_to_isc<true> 
{ 
    static void check(short scale,short type) { SOCI_UNUSED(scale) SOCI_UNUSED(type) } 
};
// and then call like this - This way compiler will call correct check at compile time without runtime check.
cond_to_isc<std::numeric_limits<T1>::is_integer>::check(scale,type);
```

ravselj:
I have pushed modifications. You can review them in the last commit. If there are any issues left with this PR let me know.

vadz:
Thanks, I've finally committed this as a single squashed commit because the changes were pretty much intermixed between the commits anyhow and, also, I wanted to have a proper commit message and not a bunch of messages starting with dashes, this really doesn't look well in the history.

I also undid the changes to `catch.hpp` because they would very likely be lost anyhow by the next update (which will happen soon BTW as the new version has some nice changes). Finally, I stripped some trailing spaces as my pre commit hook complains about them, but I didn't change anything important, so hopefully I didn't break anything neither.

Thanks again!

PR #356, opened 2015-07-02
Common test - Blank padding

ravselj:
I have found problem in FIREBIRD back-end. The problem is with char type and trailing spaces.

Current padding tests were inappropriate because actual database column size is determined also by column encoding type.

Example:
CHAR(10) CHARACTER SET NONE - (10 bytes) (Hello is represented as 'Hello.....' (dot as space)
CHAR(10) CHARACTER SET UTF8 - (40 bytes) (Hello is represented as 'Hello...................................'

By design there is no difference if we store 'Hello' or 'Hello.....' in column of type CHAR(10) CHARACTER SET NONE. Trailing spaces will be trimmed and information is lost by design. User must use VARCHAR in this case.

There is not need for SOCI to return string with trailing spaces so this PR trims them before they are returned to user.

### DIFF
--- a/tests/common-tests.h
+++ b/tests/common-tests.h
@@ -55,7 +55,31 @@ inline namespace boost {
 #endif
 #endif // HAVE_BOOST
 
+// Firebird returns string that represents entire byte-size(not character size) of the column( i.e. CHAR(10) NONE - 10bytes, CHAR(10) UTF8 - 40bytes )
+// MySql returns right trimmed value by default or space padded up to column character size when PAD_CHAR_TO_FULL_LENGTH
+// PostgreSQL returns space padded version up to actual column character size (not byte size )
+// SQLite has TEXT affinity for all char types so it is not an issue here - macro bellow will match correctly
+// TODO: comments for other backends(DB, ODBC)
 
+// This method will check result string from column defined as fixed char
+// It will check only bytes up to the original string size.
+// If padded string is bigger than expected string then all remaining chars must be spaces so if any non-space character is found it will fail.
+inline void checkEqualPadded(const std::string& padded_str, const std::string& expected_str)
+{
+    size_t const len = expected_str.length();
+    std::string const start_str(padded_str, 0, len);
+
+    if( start_str != expected_str )
+        throw soci::soci_error("Expected string \"" + expected_str +"\" does not match with padded string \"" + padded_str + "\"");
+    if ( padded_str.length() > len )
+    {
+        std::string const end_str(padded_str, len);
+        if( end_str != std::string(padded_str.length() - len, ' ') )
+            throw soci::soci_error("\"" + padded_str + "\" starts with \"" + padded_str + "\" but non-space characater(s) are found aftewards!");
+    }
+}
+#define CHECK_EQUAL_PADDED(padded_str, expected_str) \
+    CHECK_NOTHROW(checkEqualPadded(padded_str,expected_str));
 
 // Objects used later in tests 14,15
 struct PhonebookEntry
@@ -316,6 +340,9 @@ public:
     // Override this to call commit() if it's necessary for the DDL statements
     // to be taken into account (currently this is only the case for Firebird).
     virtual void on_after_ddl(session&) const { }
+
+    // Override this if the backend does not support trailing space character type
+    virtual bool enable_std_char_padding(session&) const { return true; }
 
     virtual ~test_context_base()
     {
@@ -1977,12 +2004,13 @@ TEST_CASE_METHOD(common_tests, "Dynamic row binding", "[core][dynamic]")
         CHECK(t.tm_year == 105);
 
         // again, type char is visible as string
-        CHECK(r.get<std::string>(4) == "a");
+        // Firebird with default encoding other than NONE can return multiple bytes ('a   'for UTF-8 default encoding).
+        CHECK_EQUAL_PADDED(r.get<std::string>(4),"a");
 
         ASSERT_EQUAL_APPROX(r.get<double>("NUM_FLOAT"), 3.14);
         CHECK(r.get<int>("NUM_INT") == 123);
         CHECK(r.get<std::string>("NAME") == "Johny");
-        CHECK(r.get<std::string>("CHR") == "a");
+        CHECK_EQUAL_PADDED(r.get<std::string>("CHR"),"a");
 
         CHECK(r.get_indicator(0) == i_ok);
 
@@ -2017,7 +2045,7 @@ TEST_CASE_METHOD(common_tests, "Dynamic row binding", "[core][dynamic]")
             CHECK(t.tm_hour == 22);
             CHECK(t.tm_min == 14);
             CHECK(t.tm_sec == 17);
-            CHECK(c == "a");
+            CHECK_EQUAL_PADDED(c, "a");
         }
     }
 
@@ -2548,11 +2576,11 @@ TEST_CASE_METHOD(common_tests, "Reading rows from rowset", "[core][row][rowset]"
                 std::tm t1 = { 0 };
                 t1 = r1.get<std::tm>(3);
                 CHECK(t1.tm_year == 105);
-                CHECK(r1.get<std::string>(4) == "a");
+                CHECK_EQUAL_PADDED(r1.get<std::string>(4),"a");
                 ASSERT_EQUAL_APPROX(r1.get<double>("NUM_FLOAT"), 3.14);
                 CHECK(r1.get<int>("NUM_INT") == 123);
                 CHECK(r1.get<std::string>("NAME") == "Johny");
-                CHECK(r1.get<std::string>("CHR") == "a");
+                CHECK_EQUAL_PADDED(r1.get<std::string>("CHR"),"a");
             }
             else if (name == "Robert")
             {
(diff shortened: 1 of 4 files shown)


vadz:
Sorry, I'm confused about what exactly is the net effect of these changes. AFAICS there are 2 things here:

 1. Test changes which I don't quite understand: do you mean that the tests failed for you? They pass for me (and for Travis), so what purpose exactly do the test changes serve?
 1. Trim trailing spaces when retrieving the values from the database in SOCI. This seems to make sense to me, but I wonder if this is consistent with the behaviour of the other backends, i.e. could we have a common test verifying that all of them behave like this?

TIA for any additional explanations!

ravselj:
First I have found the issue while testing firebird so I naturally assumed this was Firebird issue. 

After your comment I've looked into MySql and PostgreSql and found out that the all share same architecture. (Sqlite is an exception with TEXT affinity - all char types are mapped to TEXT).

There are some differences though:
- MySql returns right-trimmed version by default.( i.e. CHAR(10) with value 'abc.......' is returned as abc. If PAD_CHAR_TO_FULL_LENGTH is set then also spaces are return up to full character size.
- PostreSql returns full value with padded spaces up to full character size.
- Firebird returns full value with padded spaces up to full column BYTE size.
- SQlite maps CHAR to TEXT affinity and does not have space padding.

DB2,ODBC - Not tested.

In first two cases(MySql,PostgreSql) we have no idea how much memory backends actually use to represent string. What matters is the max character size in column. 

What I have changed from previous commit is to revert back soci-firebird changes as I and added new test to common-tests.h to check for padding and made some minor modifications to firebird test where needed.

I also needed to change the way fixed CHAR types are validated in common-test. I have added new macro CHECK_CHAR which validates fixed char column data for all scenarios.

What I suggest as an upgrade to SOCI is to define some global session option in the future i.e. SOCI_CHAR_PADDING which should probably be disabled by default so trimmed data is always returned in every backend unless users enables padding. However, this is a matter for another PR.

tests/mysql/test-mysql.h
```diff
@@ -112,6 +112,13 @@ class test_context : public test_context_base
         // SQL mode to avoid silent truncation of too long values.
         return sql_mode.find("STRICT_") == std::string::npos;
     }
+
+    virtual bool has_trailing_spaces_in_char_type(session& sql) const 
+    {
+        //turn on padding on mysql - trimming by default
+        sql << "SET @@session.sql_mode = 'PAD_CHAR_TO_FULL_LENGTH'";
```
vadz: According to the docs, this mode is new in 5.1.20, I don't know if it there is still any real possibility of encountering an older MySQL version out there, but if we can check for the error here, it would be worth doing it.

tests/common-tests.h
```diff
@@ -317,6 +330,9 @@ class test_context_base
     // to be taken into account (currently this is only the case for Firebird).
     virtual void on_after_ddl(session&) const { }
 
+    // Override this if the backend does not support trailing space character type
+    virtual bool has_trailing_spaces_in_char_type(session&) const { return true; }
```
vadz: I think the function name is confusing, it doesn't check if the backend uses trailing spaces, as `has_` prefix implies, but needs to be called in order to enable, or at least try to enable, standard `CHAR` handling. So I'd call it something like `enable_std_char_padding()`, with the understanding that it returns false if this can't be done (SQLite).

tests/common-tests.h
```diff
@@ -55,7 +55,20 @@ inline namespace boost {
...
-
+// Firebird returns string that represents entire byte-size(not character size) of the column( i.e. CHAR(10) NONE - 10bytes, CHAR(10) UTF8 - 40bytes )
+// MySql returns right trimmed value by default or space padded up to column character size when PAD_CHAR_TO_FULL_LENGTH
+// PostgreSQL returns space padded version up to actual column character size (not byte size )
+// SQLite has TEXT affinity for all char types so it is not an issue here - macro bellow will match correctly
+// TODO: comments for other backends(DB, ODBC)
+//This macro will check result string from column defined as fixed char
+//It will check only bytes up to original string size.
+//If result string is bigger than original string then all remaining chars must be spaces so if any non-space char is found it will fail .
+#define CHECK_CHAR( STR, ORIG_STR ) \
```
vadz: The macro name is too generic, I'd call it `CHECK_EQUAL_PADDED()` or something like this.

Also, from stylistic point of view, it's weird to use ALL CAPS for the macro arguments, they usually use the same convention as normal variables (i.e. I'd call them `padded_str` and `expected_str` perhaps).

tests/common-tests.h
```diff
@@ -55,7 +55,20 @@ inline namespace boost {
...
+    CHECK(strncmp(std::string(STR).c_str(),std::string(ORIG_STR).c_str(),std::string(ORIG_STR).size()) == 0); \
```
vadz: This is problematic because the error message if the strings don't match won't show their values, which is very useful for debugging the problem. So I'd rather do something like this:
```
size_t const len = strlen(expected_str);
std::string const start_str(padded_str, 0, len);
CHECK(start_str == expected_str);
if ( padded_str.length() > len )
{
    std::string const end_str(padded_str, len);
    CHECK(end_str == std::string(padded_str.length() - len, ' '));
}
```

tests/common-tests.h
```diff
@@ -4041,6 +4057,52 @@ TEST_CASE_METHOD(common_tests, "Truncation error", "[core][insert][truncate][exc
...
+TEST_CASE_METHOD(common_tests, "Blank padding", "[core][insert][exception]")
+{
+    soci::session sql(backEndFactory_, connectString_);
+    if( !tc_.has_trailing_spaces_in_char_type(sql) )
+    {
+        WARN("No need to check for trailing space functionallity on backend that does not support it.");
+        return;
+    }
+
+    SECTION("Check blank padding in character types")
```
vadz: Why have a single section? This doesn't seem useful...

vadz:
I'm still not sure why do Firebird tests pass for me (under both Windows and Linux) right now but otherwise these changes make sense, thanks! If you could please address the comments above, I'd be glad to merge this.

In addition to those comments, one minor comment about style: SOCI code generally uses more spaces than you do, e.g. `// comment` instead of `//comment` and `CHECK(a, b)` instead of `CHECK(a,b)`. It is very minor, of course, but it would be nice if we could keep things consistent. TIA!

tests/mysql/test-mysql.h
ravselj: Ok. I'll catch for exception here with message indicating support as of MySQL 5.1.20.

tests/common-tests.h
ravselj: Ok.

tests/common-tests.h
ravselj: If so then maybe it's better if we make inline function rather than macro. 

inline checkEqualPadded(const std::string& padded_str, const std::string& expected_string)
{
 ... your code above
}
and then define macro just for code convenience
#define CHECK_EQUAL_PADDED(padded_string, expected_string) checkEqualPadded(padded_string, expected_string);

tests/common-tests.h
ravselj: This is copy-paste issue. Of course section here is unnecessary and will be removed.

ravselj:
I have no problem with following the code style. I'll also look all the changes again and fix style inconsistencies.

Firebird test:
I am using Windows build for testing with and connection to remote Firebird DB on Ubuntu Linux server. (Firebird TCP/IP server version LI-V2.5.2.26540 Firebird 2.5) Firebird database is created with the following command:

create database '/home/firebird/test.fdb' default character set 'UTF8' user 'SYSDBA' password '******';

this way any char or varchar is implicitelly set to UTF8.

tests/common-tests.h
vadz: The trouble is that if we use a function, the location of (failing) `CHECK`s inside it won't be correct as it would use the line number corresponding to the code of the function itself instead of the code calling it, this is the only reason to use macros for checks like this.

Unfortunately I don't think Catch provides macro like `CHECK_WITH_LOCATION(..., file, line)` that we'd need to implement such a function properly.

vadz:
I guess the default charset is the difference, thanks for the explanation.

tests/common-tests.h
ravselj: The problems with just macro are variable declaration( it can be solved by wrapping macro code in {}) and debugging so I have extended Catch with CHECK_WITH_LOCATION as you suggested above.

vadz: FWIW there is also `INTERNAL_CATCH_UNIQUE_NAME()` macro which could be used to create a unique name for a variable (but I originally indeed though about just using `{...}` around this code).

ravselj:
@vadz,

I have just pushed changes so you can review them. I have used CHECK_WITH_LOCATION in inline function but this functionally can easily be changed to just using wrapped macro if you have any objections.

ravselj:
Hello @vadz,

I have found another way in existing Catch to achieve the same goal using CHECK_NOTHROW instead of our macro. If you find this solution inappropriate there is also fallback version using macro but in my opinion I always rather use inline function over macro.

I have also renamed this PR's title and squashed all commits into single commit for clarity.

vadz:
I'd basically just like something that provides optimal results if the test fails. If the `CHECK_NOTHTOW()`-based solution shows the values of the strings that differ, I'm fine with it, I'm just not sure if CATCH shows the message specified during the exception construction?

ravselj:
This is example of exception message where check fails.
D:\dev\Projects\soci-ravselj\tests\common-tests.h(4115): FAILED
  CHECK_NOTHROW( checkEqualPadded(sc,"b") )
  Expected string "b" does not match with padded string "a   "

where message is constructed in checkEqualPadded. I think that should be sufficient.

vadz:
CATCH built-in equality failure message is a bit nicer, but I agree that this is quite good enough, thanks! Could you please just remove the commented out "fallback" before merging? TIA!

ravselj:
Done. Travis already doing it's job.

vadz:
I've split the commits and did some other cosmetic changes, please let me know if I did anything wrong.

Thanks!

PR #357, opened 2015-07-07
Improve ODBC error messages.

vadz:
As mentioned several times in the past, ODBC backend error messages are inconsistent and this PR is an attempt to improve things.

I decided to standardize on "Error doing something" instead of "Failed to do something", for example, as this seems to be the most widely used form but things are so inconsistent right now (not only in the ODBC backend but also between backends) that it's hard to be sure if this is the best choice.

E.g. we have the following error messages for the same error:

 1. `"Error connecting to database"` (DB2)
 1. Nothing, just using the driver message which doesn't necessarily make it clear that the error happened while opening the database connection at all (Firebird, MySQL, Oracle, SQLite3)
 1. `"Error Connecting to database"` (ODBC)
 1. `"Cannot establish connection to the database."` (PostgreSQL)

Clearly more could be done to make things more homogeneous but hopefully this is a step in the right direction...

### DIFF
--- a/include/soci/odbc/soci-odbc.h
+++ b/include/soci/odbc/soci-odbc.h
@@ -354,8 +354,6 @@ public:
 private:
     std::string interpret_odbc_error(SQLSMALLINT htype, SQLHANDLE hndl, std::string const& msg)
     {
-        std::ostringstream ss(msg, std::ostringstream::app);
-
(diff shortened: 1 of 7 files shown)


mloskot:
By all means, :+1:  and thanks!

PR #358, opened 2015-07-10
Split Statement::clean_up into bind_clean_up and clean_up

ArnaudD-FR:
Allow to reuse statement with new references by only cleaning bound values and not the backend statement.

### DIFF
--- a/docs/reference.md
+++ b/docs/reference.md
@@ -274,6 +274,7 @@ The `statement` class encapsulates the prepared statement.
         void exchange(*IT* const & i);
         void exchange(*IT* const & u);
         void clean_up();
+        void bind_clean_up();
 
         void prepare(std::string const & query);
         void define_and_bind();
@@ -303,6 +304,7 @@ This class contains the following members:
 * `bind` function, which is used to bind the `values` object - this is used in the object-relational mapping and normally called automatically.
 * exchange functions for registering the binding of local data - they expect the result of calling the `into` or `use` functions and are normally invoked automatically.
 * `clean_up` function for cleaning up resources, normally called automatically.
+* `bind_clean_up` function for cleaning up any bound references. It allows to keep statement in cache and reuse it later with new references by calling `exchange` for each new bind variable.
 * `prepare` function for preparing the statement for repeated execution.
 * `define_and_bind` function for actually executing the registered bindings, normally called automatically.
 * `execute` function for executing the statement. If its parameter is `false` then there is no data exchange with locally bound variables (this form should be used if later `fetch` of multiple rows is foreseen). Returns `true` if there was at least one row of data returned.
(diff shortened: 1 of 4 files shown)


vadz:
It would be nice if `docs/{backends,reference}.md` could be updated, perhaps with an example of how a statement can be reused thanks to the new function introduction?

mloskot:
We also need a test case.

It's awesome SOCI receives new contributions, but even smallest features need a test case and a tiny bit of docs. Otherwise, those who will follow with maintenance will need to spend more time than necessary to figure details out. I'm sure, I'm annoying with the obvious :)

vadz:
FWIW I totally agree with the need for the test case, I was going to mention it myself but forgot, thanks @mloskot for doing it.

ArnaudD-FR:
Of course, I can add documentation and test case, there is no problem

This is an example:
```c++
soci::session db("sqlite3", ":memory:");
db << "CREATE TABLE test(a INTEGER)";

{
    // prepare statement
    soci::statement stmt = (db.prepare << "INSERT INTO test(a) VALUES(?)");

    {
        // first insert
        int a0 = 0;

        // update reference
        stmt.exchange(soci::use(a0));

        stmt.define_and_bind();
        stmt.execute(true);
        stmt.bind_clean_up();
    }

    {
        // come later, second insert
        int a1 = 1;

        // update reference
        stmt.exchange(soci::use(a1));

        stmt.define_and_bind();
        stmt.execute(true);
        stmt.bind_clean_up();
    }
}

{
    std::vector<int> v(10);
    db << "SELECT a FROM test", soci::into(v);
    for (int i = 0; i < v.size(); ++i)
        std::cout << i << ": " << v[i] << std::endl;
}

db << "DROP TABLE test";

```

mloskot:
I guess, such [statement preparation](http://soci.sourceforge.net/doc/3.2/statements.html) with delayed re-binding is useful. 

:+1: thanks!

ArnaudD-FR:
I was wondering why `exchange` do not bind directly and I have to call `define_and_bind`. Is there any reason?

vadz:
I agree that `exchange()` could be smarter, see also [my old mail](https://www.mail-archive.com/soci-users@lists.sourceforge.net/msg01883.html) about this.

ArnaudD-FR:
Indeed, I didn't checked backends other than sqlite3.

But this method and the one from your old email didn't generate errors with valgrind, I mean running sqlite3 backend.

My point, when reviewing sqlite3 backend (see PR #360), was that when running `execute` all references and so their contents should not change. So any pointer to referenced values used by `pre_use` should be valid until `post_use` is called, both call are done in `execute`.

mloskot:
Regarding `exchange` and `define_and_bind` sequence, AFAIU, it is related to Oracle OCI legacy. Originally, SOCI was createda as "Simple OCI". Later, adoption to other DBMS followed.

Does this PR need more work or is it ready/accepted to merge?

ArnaudD-FR:
This PR is ready if nobody has comment on it.

About `exchange` I made some tests to bind directly exchanged values but it is no so easy, there is some corner cases because of prepare* and once* classes.

mloskot:
@ArnaudD-FR Thanks for your help!

PR #359, opened 2015-07-10
Add base_type for each exchange_traits overload

ArnaudD-FR:
base_type is now used in row::get to avoid convert issues

### DIFF
--- a/include/soci/exchange-traits.h
+++ b/include/soci/exchange-traits.h
@@ -30,14 +30,11 @@ struct exchange_traits
     // this is used for tag-dispatch between implementations for basic types
     // and user-defined types
     typedef user_type_tag type_family;
+    typedef typename type_conversion<T>::base_type base_type;
 
     enum // anonymous
     {
-        x_type =
-            exchange_traits
-            <
-                typename type_conversion<T>::base_type
-            >::x_type
+        x_type = exchange_traits<base_type>::x_type
     };
 };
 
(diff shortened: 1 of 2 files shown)


vadz:
Not sure which problem exactly does this solve, but Travis seems to have detected a real new one...

ArnaudD-FR:
I saw, I just installed a mysql db to investigate

mloskot:
Indeed, it would be good to elaborate what kind of convert issues are addressed here.
Yet, ideally, it would be awesome if corresponding test case(s) is added.

ArnaudD-FR:
It's an old commit I had on my branch, I will try to find my use case. I made this patch during the first weeks I discovered soci years agi, so it's maybe wrong now.

ArnaudD-FR:
I found my use case.

I want to do `get<unsigned int>` from a row and I got a bad cast exception.

I just found in unsigned-types.h the source error
```c++
template <>
struct type_conversion<unsigned int>
{
    typedef long long base_type;
....
```
So I want 32 bits values, but as it is stored in row in 64 bits I got a bad cast error. I believe this is most a conceptual error.

My patch was in fact a hack to retrieve the stored value with the correct type.

ArnaudD-FR:
And of course this why the mysql test case is failing

```c++
CHECK(r.get<long long>("val") == 0xffffff00);
CHECK(r.get<unsigned>("val") == 0xffffff00);
```

`long long` is 64bits and with my patch `unsigned` is 32bits

ArnaudD-FR:
I close this PR for the moment

PR #361, opened 2015-07-10
Add dt_blob

ArnaudD-FR:
Extracted from PR #360 

Add dt_blob in data_enum, this add blob support in soci-simple interface

### DIFF
--- a/docs/reference.md
+++ b/docs/reference.md
@@ -566,6 +566,27 @@ The functions above provide the *session* abstraction with the help of opaque ha
 
 Note that the only function that cannot report all errors this way is `soci_create_session`, which returns `NULL` if it was not possible to create an internal object representing the session. However, if the proxy object was created, but the connection could not be established for whatever reason, the error message can be obtained in the regular way.
 
+
+    typedef void *blob_handle;
+    blob_handle soci_create_blob(session_handle s);
+    void soci_destroy_blob(blob_handle b);
+
+    int soci_blob_get_len(blob_handle b);
+    int soci_blob_read(blob_handle b, int offset, char *buf, int toRead);
+    int soci_blob_write(blob_handle b, int offset, char const *buf, int toWrite);
+    int soci_blob_append(blob_handle b, char const *buf, int toWrite);
+    int soci_blob_trim(blob_handle b, int newLen);
+
+    int soci_blob_state(blob_handle b);
+    char const * soci_blob_error_message(blob_handle b);
+
+The functions above provide the *blob* abstraction with the help of opaque handle. The `soci_blob_state` function returns `1` if there was no error during the most recently executed function and `0` otherwise, in which case the `soci_session_error_message` can be used to obtain a human-readable error description.
+
+For easy error testing, functions `soci_blob_read`, `soci_blob_write`, `soci_blob_append`, and `soci_blob_trim` return `-1` in case of error and `soci_session_error_message` can be used to obtain a human-readable error description.
+
+Note that the only function that cannot report all errors this way is `soci_create_blob`, which returns `NULL` if it was not possible to create an internal object representing the blob.
+
+
     typedef void * statement_handle;
     statement_handle soci_create_statement(session_handle s);
     void soci_destroy_statement(statement_handle st);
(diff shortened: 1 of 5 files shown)


src/core/blob.cpp
```diff
@@ -20,31 +26,37 @@ blob::blob(session & s)
 
 blob::~blob()
 {
-    delete backEnd_;
+    if (backEnd_)
```
vadz: This change is harmless but not necessary, deleting a null pointer is a perfectly valid operation (doing nothing).

src/core/blob.cpp
```diff
@@ -20,31 +26,37 @@ blob::blob(session & s)
...
+        delete backEnd_;
 }
 
 std::size_t blob::get_len()
 {
+    assert(backEnd_);
```
vadz: I'd strongly prefer to avoid reintroducing asserts in SOCI code. In the case of this class, we should throw an exception if `backEnd_` is null from the ctor ensuring that this is never the case in any method of this class.

src/core/blob.cpp
ArnaudD-FR: Good to know :-)  I was believing `delete` works like `free` and you can not call `free` on NULL pointer.

src/core/blob.cpp
ArnaudD-FR: al right, this is true that in release mode this is not checked. I will replace them with exceptions

mloskot:
I agree with @vadz 's review. @ArnaudD-FR Thanks for the follow-up with improvements.
I'm happy to :+1: for merging.

src/core/blob.cpp
vadz: FWIW `free(NULL)` is guaranteed to do nothing as well.

vadz:
Sorry but while trying to describe the improvement in the change log I've just realized that I don't understand at all how does this work: we now have default ctor in `blob` but there doesn't seem to be any way to initialize an object created in this way, so how can this be useful? Am I missing something here or is this commit incomplete?

ArnaudD-FR:
@vadz, I totaly agree with you but I had to had it for `std::map<*, blob>` in simple_interface.

Currently there is no clean way to copy "normal" `blob` into an empty one created with default ctor. For blob copy please refer to PR #267. I didn't test it but it should do the job.

vadz:
Could we store `blob*` instead in the map (in C++11 I'd say `unique_ptr<blob>` but we're not C++11 yet, so it would have to be done with raw pointers and manual `delete`s...)? This would also obviate the need for the (inefficient, it seems) `blob` copy ctor added by #267.

ArnaudD-FR:
@vadz, I'm currently rewriting soci-simple for blob because I missed a lot of things and I was wondering if they were any support for bulk use/into of blob? With current implementation of blob class it is not possible to use them in vector, right?

vadz:
No, blobs can't be currently stored in a vector. To be honest, I don't know if it using blobs in bulk makes sense but OTOH why not...

ArnaudD-FR:
Thanks for confirmation.

ArnaudD-FR:
I've updated this PR, now user can create, manipulate, use, bind into and destroy blob using soci-simple interface. Let me know if something is wrong.

vadz:
I've never used the simple interface, so I'm almost certainly not the best person to review this, but it looks perfect to me. The one small thing I'd change: the commit message. As is, it doesn't really say much, so I'd rather use "Add blob support to the simple interface." as the summary and provide more details, if any, in the body of the message.

ArnaudD-FR:
New title and summary.

It was the first time I used the simple-interface to check if my implementation is correct and it was not so easy (C implementation didn't help). Here are the step to execute a statement:
1. create statement
2. bind used and into values
3. prepare statement
4. execute statement and fetch if necessary
5. destroy statement

You can not exchange steps 2 and 3. My first approach was to use step 3 before step 2 but a check is done and it is not possible to bind values once statement is prepared. Is it a backend requirement? I only use sqlite3.

mloskot:
@ArnaudD-FR AFAICT, it is a backend-specific requirement, but rather a unification imposed by SOCI. AFAIU, actual requirements of some of backends (perhaps all) and common sense suggested that sequence is a reasonable common denominator to all.

ArnaudD-FR:
@mloskot, thanks for your answer.

I have no other change to add to this PR, except if you have any comment.

mloskot:
Thanks!

PR #363, opened 2015-07-10
Allow setting ORACLE_HOME directly in CMake GUI

ravselj:
On windows when using CMake Oracle is not shown as an option so I have changed FindOracle.cmake to first check CMake variable ORACLE_HOME and if not set then checks for environment variable and if exists then uses its value in CMake.

This way we can set ORACLE_HOME directly in CMakeGui without setting environment variable.

I have also removed top condition where Oracle search is performed only when ORACLE_HOME is found. It should be shown in CMakeGUI the same way as all other supported back-ends so user knows that backend is available.

When compiling oracle test there was also error in VS-2012 caused by invalid header inclusion order.

### DIFF
--- a/cmake/modules/FindOracle.cmake
+++ b/cmake/modules/FindOracle.cmake
@@ -22,50 +22,51 @@
 #
 ###############################################################################
 
-# If ORACLE_HOME not defined, assume Oracle libraries not available
-if(DEFINED ENV{ORACLE_HOME})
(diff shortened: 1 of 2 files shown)


vadz:
@mloskot could you please have a look at this? I don't trust myself to do anything CMake-related...

mloskot:
Thanks!

PR #364, opened 2015-07-19
Fix connection-parameters options bug

ravselj:
I was developing and testing some additional functionality which uses options instead of connection string in SOCI and found out a bug while searching for an option. 

get_option() failed to find option by name since map matches const char* pointer value instead of option name. By changing the type to std::string the issue is resolved.

This PR also resolves compiler warning fixes in test-mysql.h. ( caused by merge of #356 )

### DIFF
--- a/include/soci/connection-parameters.h
+++ b/include/soci/connection-parameters.h
@@ -59,7 +59,7 @@ private:
     std::string connectString_;
 
     // We store all the values as strings for simplicity.
-    typedef std::map<const char*, std::string> Options;
+    typedef std::map<std::string, std::string> Options;
     Options options_;
 };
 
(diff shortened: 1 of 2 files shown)


vadz:
Oops, thanks for fixing this. I think we might keep `char const*` arguments though, for compatibility, just using `std::string` for the map key type should be enough to fix the problem, shouldn't it?

ravselj:
Of course. I have reverted the changes to both functions.

mloskot:
@ravselj Is this PR ready to merge?

ravselj:
Yes. The initial code contained some unnecessary changes( upper comment from @vadz ) which were removed by last push.

mloskot:
Thanks!

PR #366, opened 2015-07-27
"use" parameters additional functionality

ravselj:
When developing with soci I tried to use the following:

``` c++
//tried to insert null as second parameter
sql << "insert into test values( :int, :str )", use(1), use("",i_null); 
```

which was not supported because use indicator was only implemented for non const scenario.

I have also tried to call use with const char instead of std::string:

``` c++
 //This was also not supported
sql << "insert into test values( :str )", use("Hello");
```

So I wrote some additional templates and functions in use.h to support the following cases:
1.) dbNull() function to use whenever we need to insert null value into database
2.) Support for const char\* type
3.) Support for const char[N] type

In order to enable the following scenarios I needed to modify `use_container` template to support some additional cases so `bool isDataOwner` template argument is added in order to detect whether use_container holds a referenced value or holds a copy of the value passed to use function. 

With this PR additional use scenarios are possible:

``` c++
//use const char[N] directly 
sql << "insert into test values(:str)", use("Hello");

//use const char* directly
const char* value = "Hello";
sql << "insert into test values(:str)", use(value);

// use dbNull() instead of explicit indicator
sql << "insert into test values(:str)", dbNull();
// alternative to dbNull( value type is ignored in this case)
sql << "insert into test values(:str)", use("",i_null);
```

### DIFF
--- a/include/soci/type-conversion.h
+++ b/include/soci/type-conversion.h
@@ -120,7 +120,29 @@ public:
         //convert_to_base();
     }
 
+    conversion_use_type(T & value, indicator const & ind,
+            std::string const & name = std::string())
+        : use_type<base_type>(details::base_value_holder<T>::val_, ind, name)
+        , value_(value)
+        , ind_(ind)
+        , readOnly_(true) //notice const indicator
+    {
+        // TODO: likely to be removed (SHA: c166625a28f7c907318134f625ff5acea7d9a1f8)
+        //convert_to_base();
+    }
+
     conversion_use_type(T const & value, indicator & ind,
+            std::string const & name = std::string())
+        : use_type<base_type>(details::base_value_holder<T>::val_, ind, name)
+        , value_(const_cast<T &>(value))
+        , ind_(ind)
+        , readOnly_(true)
+    {
+        // TODO: likely to be removed (SHA: c166625a28f7c907318134f625ff5acea7d9a1f8)
+        //convert_to_base();
+    }
+
+    conversion_use_type(T const & value, indicator const & ind,
             std::string const & name = std::string())
         : use_type<base_type>(details::base_value_holder<T>::val_, ind, name)
         , value_(const_cast<T &>(value))
@@ -351,7 +373,21 @@ use_type_ptr do_use(T & t, indicator & ind,
 }
 
 template <typename T>
+use_type_ptr do_use(T & t, indicator const & ind,
+    std::string const & name, user_type_tag)
+{
+    return use_type_ptr(new conversion_use_type<T>(t, ind, name));
+}
+
+template <typename T>
 use_type_ptr do_use(T const & t, indicator & ind,
+    std::string const & name, user_type_tag)
+{
+    return use_type_ptr(new conversion_use_type<T>(t, ind, name));
+}
+
+template <typename T>
+use_type_ptr do_use(T const & t, indicator const & ind,
     std::string const & name, user_type_tag)
 {
     return use_type_ptr(new conversion_use_type<T>(t, ind, name));
--- a/include/soci/use-type.h
+++ b/include/soci/use-type.h
@@ -72,6 +72,21 @@ public:
         //convert_to_base();
     }
 
+    // Const indicator constructor
+    standard_use_type(void* data, exchange_type type, const indicator& ind,
+        bool readOnly, std::string const& name = std::string())
+        : data_(data)
+        , type_(type)
+        // Const cast will allow the code to compile and run but any modification to ind_ in readonly mode will result in undefined behaviour
+        , ind_(&const_cast<indicator&>(ind)) 
+        , readOnly_(readOnly)
+        , name_(name)
+        , backEnd_(NULL)
+    {
+        // FIXME
+        //convert_to_base();
+    }
+
     virtual ~standard_use_type();
     virtual void bind(statement_impl & st, int & position);
     virtual std::string get_name() const { return name_; }
@@ -164,9 +179,19 @@ public:
             static_cast<exchange_type>(exchange_traits<T>::x_type), ind, false, name)
     {}
 
+    use_type(T& t, indicator const & ind, std::string const& name = std::string())
+        : standard_use_type(&t,
+            static_cast<exchange_type>(exchange_traits<T>::x_type), ind, true, name)
+    {}
+
     use_type(T const& t, indicator& ind, std::string const& name = std::string())
         : standard_use_type(const_cast<T*>(&t),
-            static_cast<exchange_type>(exchange_traits<T>::x_type), ind, false, name)
+            static_cast<exchange_type>(exchange_traits<T>::x_type), ind, true, name)
+    {}
+
+    use_type(T const& t, indicator const & ind, std::string const& name = std::string())
+        : standard_use_type(const_cast<T*>(&t),
+            static_cast<exchange_type>(exchange_traits<T>::x_type), ind, true, name)
     {}
 };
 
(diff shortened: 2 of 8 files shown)


vadz:
I'm really not sure about whether the extra complications (and this does complicate the already not very obvious part of the SOCI code even further) is worth saving on basically a single `std::string(...)` in the code. Especially as it looks like we don't even avoid the creation of the temporary string anyhow, it's just done implicitly (I could be wrong about this though, the code is not obvious to follow...).

As for `dbNull()`, it looks useful but it should be renamed to something closer to SOCI naming conventions (straightforward equivalent would be `db_null()`, but maybe `use_null()` would be better?).

mloskot:
On 27 July 2015 at 15:22, VZ <notifications@github.com> wrote:

> I'm really not sure about whether the extra complications (and this does
> complicate the already not very obvious part of the SOCI code even further)
> is worth saving on basically a single std::string(...) in the code.
>
> As for dbNull(), it looks useful but it should be renamed to something
> closer to SOCI naming conventions (straightforward equivalent would be
> db_null(), but maybe use_null() would be better?).
>

I second @vadz opinion.

ravselj:
@vadz, @mloskot

My initial purpose was to support const indicators. It seemed natural to use const indicator value rather than always using variable for indicator.

I can easily revert changes regarding const char. As of c++ 14 there is also std::string literal support so additional templates for const char are probably not necessary but I had some brain fun during implementation :)

If you agree I'll remove all changes regarding const  char and leave support for use_null. @vadz IMO this name fits into soci naturally.

I'll rebase and push  changes shortly.

vadz:
I agree that doing `use(..., i_null)` should work, I'm just not sure what exactly is the problem with it right now?

ravselj:
>I agree that doing use(..., i_null) should work, I'm just not sure what exactly is the problem with it right now?
In current master use(...,i_null) is not possible in SOCI. It's possible to call:
```c++
indicator ind=i_null;
sql << "insert into test values(:str)", use(std::string(),ind);
```
That is basically it. use_null is directly derived from the case above. I will rebase the changes to first commit where const indicator support was added so we can continue our discussion from there.

ravselj:
@vadz ,

I have commited const indicator support with special use_null method. During testing some additional errors revealed so I needed make minor fixes in Oracle and ODBC backends. 

Whenever const value and/or indicator are used then this kind of use scenario must be checked for readOnly_ flag before setting indicator in both backends.

Please let me know if you have any other questions or comments.

include/soci/type-conversion.h
```diff
@@ -120,6 +120,17 @@ class conversion_use_type
         //convert_to_base();
     }
 
+    conversion_use_type(T & value, indicator const & ind,
+            std::string const & name = std::string())
+        : use_type<base_type>(details::base_value_holder<T>::val_, ind, name)
+        , value_(value)
+        , ind_(ind)
```
vadz: I must be missing something here... how can this ctor work when `ind` is a const reference and `ind_` a non-const one? AFAICS any attempt to use it should result in a compilation error.

include/soci/use-type.h
```diff
@@ -72,6 +72,21 @@ class SOCI_DECL standard_use_type : public use_type_base
         //convert_to_base();
     }
 
+    // Const indicator constructor
+    standard_use_type(void* data, exchange_type type, const indicator& ind,
+        bool readOnly, std::string const& name = std::string())
+        : data_(data)
+        , type_(type)
+        // Const cast will allow the code to compile and run but any modification to ind_ in readonly mode will result in undefined behaviour
```
vadz: This is not very reassuring... Especially because whether `ind_` is modified or not is controlled by `readOnly` so there seems to be a possibility of constructing this object with const indicator but `readOnly == false` which would result in disaster.

IMHO we shouldn't have `readOnly` parameter in this ctor overload at all, it should always be true...

vadz:
This is much more clear now, thanks a lot!

I think the patch should be good to apply now, we just need to update the docs and maybe the change log.

But I also think I understand know that all the problems really come from support for in/out parameters, i.e. those that are specified with `use()` but can be modified -- this is why indicators are passed by (non-const) reference in the first place. Considering how rarely are such parameters used and that they don't even seem to be supported by all backends (do we have any tests for them in our test suite? I'm not sure...), I wonder if we couldn't simplify things by using a separate `inout()` instead of `use()` for them and assume that `use()` parameters are never modified? I just don't like how complicated the current code is...

ArnaudD-FR:
I agree with @vadz , it would more clear.

include/soci/type-conversion.h
ravselj: Of course. This is never called so I missed it.

include/soci/use-type.h
ravselj: I agree. The other alternative is to add another member( see conversion_use_type.h where 2 indicators are used)

The cause of this complication is the in/out support in use scenario therefore ```readOnly``` flag was added and that complicated things.

mloskot:
@vadz 

The in/out paraneters are used with procedures.
Yes, we do have the in/out tested, see [test-oracle.cpp:350-387](https://github.com/SOCI/soci/blob/b5d65076f6a3f532bf2c6e87ff17ac6e3eae12e1/tests/oracle/test-oracle.cpp#L350-L387)

mloskot:
:+1: for merging

ravselj:
@vadz ,
in/out parameters are the cause since they do not belong into use scenario in the first place.

:+1: for simplifying. I would make this a lot easier to handle.

How would you like to proceed with this PR? I would rather implement new ```use``` and then add ```inout``` that continue with current code.

Should this be done in a new PR?

vadz:
@mloskot Thanks for the pointer, I thought I saw this somewhere but couldn't find it by grepping. I still wonder if this is supported by anything other than Oracle backend though? Both [PostgreSQL](https://github.com/SOCI/soci/blob/e5f577f72b39ce459965aeae3895ae7b02a3aa4a/tests/postgresql/test-postgresql.cpp#L165-L167) and [MySQL](https://github.com/SOCI/soci/blob/e5f577f72b39ce459965aeae3895ae7b02a3aa4a/tests/mysql/test-mysql.cpp#L55-L58) tests use separate in and out parameters for a procedure call, [Firebird one](https://github.com/SOCI/soci/blob/e5f577f72b39ce459965aeae3895ae7b02a3aa4a/tests/firebird/test-firebird.cpp#L843-L849) even uses separate syntax for them, so perhaps it _only_ works for Oracle? If so, I think breaking compatibility and forcing people to use a separate `inout()` for this would be justified as, in spite of its roots, I think that Oracle is the backend used the least with SOCI now and it seems wrong to complicate the common code so much just to cater for it.

ravselj:
I agree.

mloskot:
I'm not entirely sure about the origin of the in/out, but it smells like Oracle spirit indeed.
Regarding popularity of Oracle backend, I have no data to say one way or another. AFAIK, @msobczak uses SOCI with Oracle, so perhaps Maciej will chime in and shed more light.

Regardless, I'm neutral about breaking the compatibility here, so voting +0  :)

MonsieurNicolas:
I've run into issues many times where people call `use` with temp variables that end up being destructed before calling `execute` later in the code. Those bugs are typically quite hard to notice as they are compiler /stack dependent (use of freed stack objects).
A simple illustration is an issue I found recently:
(`st` is of type `soci::statement` and `convert` returns a `std::string`)
```
     st.exchange(use(convert(port)));  /// bug
     st.define_and_bind();
     st.execute(true);
```
That instead should be
```
     std::string converted(convert(port));
     st.exchange(use(converted));
     st.define_and_bind();
     st.execute(true);
```
As binding is done in a separate line, I think the only way to avoid this is to force `use` to only work with lvalues - which is basically in contradiction with this change set.
Maybe one way to align with the above issue is to change this pr to only work with prvalues.
(note: I am not sure how to make this work with older C++)

vadz:
@MonsieurNicolas this is a good point and I think it would be worth to explicitly break compilation with rvalues (it won't help with C++98 but at least it will help those using C++11). However I'm not really sure how is it related to this PR?

MonsieurNicolas:
It seems that this pr as it is increases the possibility of using temp variables, even though I think the goal is to just enable the use of constants.

mloskot:
What is the final word: reject or "work needed"

ravselj:
With respect to the thoughts above I think it is better to reject this PR. The proper way for this would be to rewrite all three scenarios( into,use,inout) which will break back compatibility. If I find some time to implement I will open another PR with reference to this one.

mloskot:
@ravselj Agreed. Thank you for your efforts.

PR #368, opened 2015-08-03
CMake changes

ravselj:
When ```CMAKE_DEBUG_POSTFIX``` is used soci did not load backend correctly in backend loader so I added it to ```CMake```.

For example I normally use "d" as debug postfix which results in soci_core_4_0d.dll, soci_empty_4_0d.dll, ... on windows platform. This way distinction is made for debug/release modules.

Second change is added support for MSVC multiprocess compiling( enabled by default ).


mloskot:
Thank you

Issue #370, opened 2015-08-07
Boost 1.58.0 fails static assertion when streaming optional w/o optional_io header


musopr:
Boost 1.5.8 fails a static assertion when trying to stream an optional (ref: [#10825](http://www.boost.org/users/history/version_1_58_0.html)).

It seems that including boost/optional/optional_io.hpp is required as of version 1.5.8 to prevent a bug that occurs when streaming optional objects. Including this header in soci/boost-optional.h seems to resolve the issue.

mloskot:
Thanks for the report. Pull request is welcome.

snikulov:
@musopr then submit pull request :) :+1:

PR #371, opened 2015-08-07
Fix issue SOCI/soci#370

musopr:
Okay, here it is.

### DIFF
--- a/include/soci/boost-optional.h
+++ b/include/soci/boost-optional.h
@@ -11,6 +11,7 @@
 #include "soci/type-conversion-traits.h"
 // boost
 #include <boost/optional.hpp>
+#include <boost/optional/optional_io.hpp>
 
 namespace soci
 {


vadz:
I'm not sure when did this header appear, but it's pretty old, I see that it already was in 1.47 and it's only absent from the very ancient 1.33, so I guess it should be fine.

What is the current Boost version requirement in SOCI anyhow?

musopr:
I was trying to find that out. I only thing I've seen so far is that at least version 1.35 is needed for boost::fusion::vector support, mentioned on the [Integration with Boost](http://soci.sourceforge.net/doc/3.2/boost.html) page.

musopr:
I could do a conditional include for Boost versions >= 1.58, since this is the latest version to throw an error.

snikulov:
@musopr already on master with 063e7537403fc609d042824102df1550f27691a2

PR #372, opened 2015-08-08
Include SOCI_CXX_VERSION_FLAGS when compiling with Clang

musopr:
Hi,

When compiling with Clang, SOCI_CXX_C11 is ignored whether it is ON or OFF due to SOCI_CXX_VERSION_FLAGS being set but never used.


mloskot:
Thank you

PR #375, opened 2015-08-11
Sqlite backend fetch bug

ravselj:
I discovered SQLite backend bug when I updated to the latest code and use it in my application. I had some complex code that iterates through query results and call fetch many times not just classic ```while(st.fetch()) ...``` so endless loop occurred and this issue was discovered.

The following code describes the problem:
```c++
int id;
statement st = (sql.prepare << "select id from soci_test", into(id));
st.execute();
int count = 0;
while(st.fetch())
  count++;
// Here sqlite backend returned true
bool fetchEnd = st.fetch();
```

This PR resolves the issue described and also add additional test to common tests to check for this type of error.

### DIFF
--- a/src/backends/sqlite3/statement.cpp
+++ b/src/backends/sqlite3/statement.cpp
@@ -200,8 +200,10 @@ sqlite3_statement_backend::load_rowset(int totalRows)
 statement_backend::exec_fetch_result
 sqlite3_statement_backend::load_one()
 {
-    statement_backend::exec_fetch_result retVal = ef_success;
+    if( !databaseReady_ )
+        return ef_no_data;
 
+    statement_backend::exec_fetch_result retVal = ef_success;
     int const res = sqlite3_step(stmt_);
 
     if (SQLITE_DONE == res)
@@ -218,10 +220,9 @@ sqlite3_statement_backend::load_one()
 
         std::ostringstream ss;
         ss << "sqlite3_statement_backend::loadOne: "
-           << zErrMsg;
+            << zErrMsg;
         throw sqlite3_soci_error(ss.str(), res);
     }
-
     return retVal;
 }
 
@@ -290,7 +291,8 @@ sqlite3_statement_backend::bind_and_execute(int number)
             return load_rowset(number);
         }
 
-        retVal = load_one(); //execute each bound line
+        databaseReady_=true; // Mark sqlite engine is ready to perform sqlite3_step
+        retVal = load_one(); // execute each bound line
         rowsAffectedBulkTemp += get_affected_rows();
     }
 
--- a/tests/common-tests.h
+++ b/tests/common-tests.h
@@ -518,6 +518,18 @@ TEST_CASE_METHOD(common_tests, "Basic functionality")
     int id;
     sql << "select id from soci_test", into(id);
     CHECK(id == 123);
+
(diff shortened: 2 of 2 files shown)


src/backends/sqlite3/statement.cpp
```diff
@@ -201,25 +201,29 @@ statement_backend::exec_fetch_result
 sqlite3_statement_backend::load_one()
 {
     statement_backend::exec_fetch_result retVal = ef_success;
-
-    int const res = sqlite3_step(stmt_);
-
-    if (SQLITE_DONE == res)
-    {
-        databaseReady_ = false;
+    if( !databaseReady_ )
```
ArnaudD-FR: Instead of indenting everything in else, can you just write:
``` c++
if (!databaseReady_)
    return ef_no_data;
```

It's easier to read instead of embedded if/else

ArnaudD-FR:
My point is this update is a workaround, we don't catch the root cause. I don't understand why statement has been reset between those 2 `fetch` calls.

ArnaudD-FR:
OK, forget my previous comment, we can not avoid this.

I've checked sqlite3 documentation and effectively `sqlite3_step` should not be called once it returned `SQLITE_DONE`.

So it's OK for me to add 
``` c++
if (!databaseReady_)
    return ef_no_data;
```
in `load_one`.

ravselj:
From Sqlite docs(https://www.sqlite.org/c3ref/step.html):
 >SQLITE_DONE means that the statement has finished executing successfully. ```sqlite3_step()``` should not be called again on this virtual machine without first calling ```sqlite3_reset()``` to reset the virtual machine back to its initial state.

After inspecting sqlite code I discoverd the following comment in ```sqlite3Step(Vdbe *p)```:
>   We used to require that sqlite3_reset() be called before retrying
    ** sqlite3_step() after any error or after SQLITE_DONE.  But beginning
    ** with version 3.7.0, we changed this so that sqlite3_reset() would
    ** be called automatically instead of throwing the SQLITE_MISUSE error.
    ** This "automatic-reset" change is not technically an incompatibility, 
    ** since any application that receives an SQLITE_MISUSE is broken by
    ** definition.

The second call after ```SQLITE_DONE``` to ```sqlite3_step()``` implicitly calls ```sqlite3_reset()```.

So I don't agree that this is a workaround. After ```SQLITE_DONE``` is received backend is responsible not to call ```sqlite3_step``` again by testing for ```databaseReady_```. If you look at ```load_rowset``` the same thing is properly checked.

ravselj:
:) In sync. I'll push the fixed code no problem.

snikulov:
@ravselj Could you please provide test case for this issue first? Thank you.

ravselj:
@snikulov Is it not there? Check the changes in ```common-test.h```.

snikulov:
@ravselj could you please split your PR? First - test, which expose the issue, second - fix.

ravselj:
@snikulov I have reversed the commit order so that in first commit test is added indicating the problem. Second commit is the back-end fix resolving the issue.

mloskot:
Travis CI is green, so if there's no objection, I'd like to merge this one. Are we all OK?

snikulov:
@mloskot Yes, I'm fine with it.

PR #378, opened 2015-08-17
Memory leak fix in sqlite3 backend

ravselj:
This PR fixes memory leak found in SQLite3 backend.  ```check_sqlite_err``` lacked the call to ```sqlite3_close(conn)``` whenever ```sqlite3_session_backend``` failed.

Note: here probably ```sqlite3_close_v2``` should be called since ```sqlite3_open_v2``` is used but currently Travis fails in that case.


mloskot:
Thank you

PR #380, opened 2015-08-26
Build system: Replacing CMAKE_SOURCE_DIR with CMAKE_CURRENT_SOURCE_DIR

dgrafe:
Constructing/Setting the include directories failed when integrating
SOCI as a subdirectory in an existing CMAKE project.

Replacing CMAKE_SOURCE_DIR with CMAKE_CURRENT_SOURCE_DIR. This will
make {set,get}_property work on the correct directory.


mloskot:
Thanks!

PR #381, opened 2015-08-26
sqlite3_close moved after sqlite3_errmsg is called.

ravselj:
#378 resolved memory leak but produced another bug  in sqlite session constructor resulting in invalid exception message. I have totally overlooked the call to ```sqlite3_errmsg``` after ```sqlite3_close``` in my previous PR.

Now ```sqlite3_close``` is moved right before throw so exception message is safely retrieved before connection is closed.


mloskot:
:+1:

PR #385, opened 2015-09-10
Added missing memory initialisation

atvise:
- Memory should be initialised to avoid uninitialised memory access or undefined behaviour.


vadz:
I'm not sure why is it not needed. After a (very) quick glance it looks like all these buffers are going to be immediately written to, can they really be read before they're filled?

mloskot:
I see no real issue, at least none has been practically observed.
The SQLColBind does not require initialised memory block (usage example shows exactly what SOCI does https://msdn.microsoft.com/en-us/library/ms710118%28v=vs.85%29.aspx).

I assume, it's pedantic-kind of patch. Unnecessary.

Issue #369, opened 2015-08-05
Building of very first file fails with mingw


Superlokkus:
With the most recent master checkout (70085e8a0a6d0d) building with mingw fails now at the very first file. 
Although I already got one ticket ( #338 ) about building problems with mingw, I have opened this one, because I assume this is just a minor, different, one line correction bug.
```
C:\Users\klm\Downloads\soci_build>"C:\Program Files (x86)\Dev-Cpp\MinGW64\bin\mi
ngw32-make.exe"
Scanning dependencies of target soci_core
[  1%] Building CXX object src/core/CMakeFiles/soci_core.dir/backend-loader.cpp.
obj
C:\Users\klm\Downloads\soci\src\core\backend-loader.cpp:1:0: error: -fPIC ignore
d for target (all code is position independent) [-Werror]
 //
 ^
cc1plus.exe: all warnings being treated as errors
src\core\CMakeFiles\soci_core.dir\build.make:54: recipe for target 'src/core/CMa
keFiles/soci_core.dir/backend-loader.cpp.obj' failed
mingw32-make[2]: *** [src/core/CMakeFiles/soci_core.dir/backend-loader.cpp.obj]
Error 1
CMakeFiles\Makefile2:125: recipe for target 'src/core/CMakeFiles/soci_core.dir/a
ll' failed
mingw32-make[1]: *** [src/core/CMakeFiles/soci_core.dir/all] Error 2
Makefile:126: recipe for target 'all' failed
mingw32-make: *** [all] Error 2
```

Superlokkus:
So I erased the `-fPIC` flag in the `SociConfig.cmake` file, but after that I got another build error:
```
[ 75%] Building CXX object src/backends/empty/CMakeFiles/soci_empty_static.dir/v
ector-use-type.cpp.obj
Linking CXX static library ..\..\..\lib\libsoci_empty_4_0.a
[ 75%] Built target soci_empty_static
Scanning dependencies of target soci_odbc
[ 77%] Building CXX object src/backends/odbc/CMakeFiles/soci_odbc.dir/blob.cpp.o
bj
In file included from C:/Users/klm/Downloads/soci/include/soci/odbc/soci-odbc.h:
31:0,
                 from C:\Users\klm\Downloads\soci\src\backends\odbc\blob.cpp:9:
C:/Program Files (x86)/Windows Kits/8.0/Include/um/windows.h:31:0: error: ignori
ng #pragma region Application [-Werror=unknown-pragmas]
[...]

PR #388, opened 2015-09-14
updated to latest Catch UT framework plus PR #496 to work with MinGW

snikulov:
Updated to v1.2.1 of Catch UT framework, since old one failed to build with MinGW (related to #369)
Also applied https://github.com/philsquared/Catch/pull/496 because without it, compilation fails with -Werror option set.


vadz:
Merged (as 7747cab), thank you!

PR #390, opened 2015-09-15
mingw: fix for #369

snikulov:
Hello,

Here the fix for issue #369

### DIFF
--- a/cmake/SociConfig.cmake
+++ b/cmake/SociConfig.cmake
@@ -51,7 +51,7 @@ if (MSVC)
 else()
 
   set(SOCI_GCC_CLANG_COMMON_FLAGS
-	"-pedantic -Werror -Wall -Wpointer-arith -Wcast-align -Wcast-qual -Wfloat-equal -Wredundant-decls -Wno-long-long")
+	"-Werror -Wall -Wpointer-arith -Wcast-align -Wcast-qual -Wfloat-equal -Wredundant-decls -Wno-long-long")
 
 
   if (SOCI_CXX_C11)
@@ -63,7 +63,7 @@ else()
 
   if(CMAKE_COMPILER_IS_GNUCC OR CMAKE_COMPILER_IS_GNUCXX)
 
-    set(CMAKE_CXX_FLAGS "${CMAKE_CXX_FLAGS} -fPIC ${SOCI_GCC_CLANG_COMMON_FLAGS} ${SOCI_CXX_VERSION_FLAGS} ")
+    set(CMAKE_CXX_FLAGS "${CMAKE_CXX_FLAGS} ${SOCI_GCC_CLANG_COMMON_FLAGS} ${SOCI_CXX_VERSION_FLAGS} ")
     if (CMAKE_COMPILER_IS_GNUCXX)
         if (CMAKE_SYSTEM_NAME MATCHES "FreeBSD")
             set(CMAKE_CXX_FLAGS "${CMAKE_CXX_FLAGS}")


include/soci/soci-platform.h
```diff
@@ -20,7 +20,7 @@
 #include <cstdlib>
 #include <ctime>
 
-#if defined(_MSC_VER) || defined(__MINGW32__)
+#if defined(_MSC_VER)
```
vadz: I'm pretty sure that MinGW does use `I64` when using msvcrt.dll, and doesn't support `ll`. So this test must be made conditional on MinGW version/kind, otherwise it will fix some of them while breaking the others.

vadz:
I'd gladly merge the first (CMake) part of the commit if you could extract it in a separate one because it makes sense to me, but the second one is not correct IMHO.

snikulov:
@vadz  here the log
```
[ 72%] Building CXX object src/backends/odbc/CMakeFiles/soci_odbc.dir/standard-into-type.cpp.obj
C:\WORK\GitHub\dev\soci\src\backends\odbc\standard-into-type.cpp: In member function 'virtual void soci::odbc_standard_into_type_backend::post_fetch(bool, bool, soci::indicator*)':
C:\WORK\GitHub\dev\soci\src\backends\odbc\standard-into-type.cpp:175:53: error: ISO C++98 does not support the 'I' scanf flag [-Werror=format=]
           if (sscanf(buf_, "%" LL_FMT_FLAGS "d", &ll) != 1)
                                                     ^
C:\WORK\GitHub\dev\soci\src\backends\odbc\standard-into-type.cpp:175:53: error: format '%d' expects argument of type 'int*', but argument 3 has type 'long long int*' [-Werror=format=]
C:\WORK\GitHub\dev\soci\src\backends\odbc\standard-into-type.cpp:183:53: error: ISO C++98 does not support the 'I' scanf flag [-Werror=format=]
           if (sscanf(buf_, "%" LL_FMT_FLAGS "u", &ll) != 1)
                                                     ^
C:\WORK\GitHub\dev\soci\src\backends\odbc\standard-into-type.cpp:183:53: error: format '%u' expects argument of type 'unsigned int*', but argument 3 has type 'long long unsigned int*' [-Werror=format=]
cc1plus.exe: all warnings being treated as errors
```
I using

```
c:\WORK\GitHub\dev\soci\build.gcc48>
c:\WORK\GitHub\dev\soci\build.gcc48>gcc --version
gcc (x86_64-posix-seh-rev1, Built by MinGW-W64 project) 4.8.3
Copyright (C) 2013 Free Software Foundation, Inc.
This is free software; see the source for copying conditions.  There is NO
warranty; not even for MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.


c:\WORK\GitHub\dev\soci\build.gcc48>
```

Could you please provide me with your configuration to check?

snikulov:
@vadz also please check #369 itself. You'll find the same issue in logs.

vadz:
E.g. using `g++.exe (tdm64-1) 4.9.2` and compiling this:

```
#include <stdio.h>

int main() {
    long long ll = 0x1234567890abcdefLL;
    printf("I64 = %I64d\n", ll);
    printf("ll  = %lld\n", ll);
}
```
I get
```
$ g++ -Wall i64.cpp
i64.cpp: In function 'int main()':
i64.cpp:6:30: warning: unknown conversion type character 'l' in format [-Wformat=]
     printf("ll  = %lld\n", ll);
                              ^
i64.cpp:6:30: warning: too many arguments for format [-Wformat-extra-args]
```

vadz:
In fact I also get exactly the same result using MinGW-w64 build of 4.8.0 and plain MinGW32 build of 4.8.1

Where does your MinGW build come from? I have a huge collection of MinGW versions here, going back to 3.4.5, but none of them identifies like `-posix-` something... Is it a Cygwin version maybe?

snikulov:
@vadz I use http://mingw-w64.org/doku.php/download/mingw-builds

But you've forget to add ```-pedantic``` to your build.

Full build parameters for SOCI
```-pedantic -Werror -Wall -Wpointer-arith -Wcast-align -Wcast-qual -Wfloat-equal -Wredundant-decls -Wno-long-long -std=gnu++98  -Wno-variadic-macros```

You can see it here https://travis-ci.org/snikulov/soci/jobs/80389595

snikulov:
@vadz BTW, I think that "-fPIC" can be completely and safely removed at all.
According to log from travis-ci this option added by cmake for shared libs.

vadz:
Oops, yes `-pedantic` does make the error appear. What I don't understand is that `%lld` is still flagged as an error, with or without pedantic, why doesn't this happen when building SOCI?

FWIW warnings or not, both format specifiers do work fine during run-time...

snikulov:
@vadz pedantic just force to use standard without ms-extension. 
Because -Wno-long-long inhibit the warning messages about long long usage.

vadz:
Yes, I understand what `-pedantic` does. But I do use `-Wno-long-long` and still get warnings about `%lld`. I.e. for me the only way to print out a long long without warnings with MinGW is to use `%I64d` without `-pedantic` (`-std=c++98` is ok though).

snikulov:
@vadz so what is your proposal?

vadz:
The only solution I see right now is to remove `-pedantic` and leave the code as is.

The worrying thing is that I still don't understand what changed to break this. `I64` was there since 2012 (see b82364ee) and `-pedantic` since 2010 (see 6ef88a25), has it really been broken since 3+ years?

snikulov:
@vadz Well, no problem. I can remove it.
@mloskot do you agree to lower warning letter and use non-standard ms extension with gcc on Windows platform?

mloskot:
@snikulov Yes, I agree with @vadz that for MinGW, it seems we have to `%I64d` without `-pedantic`.

snikulov:
@mloskot @vadz Ok. I've updated PR. Let's wait for Travis :)

snikulov:
@vadz @mloskot Could you please review?

vadz:
Thanks, I've merged this now, please let me know if I did anything wrong as I didn't test this at all.

snikulov:
@vadz unfortunately removing pedantic does not solved issue completely :(

I've found that http://sourceforge.net/p/mingw-w64/mailman/message/29704782/ for MinGW it's better to provide ```-Wno-pedantic-ms-format```

Looks like it was changed since 4.7.1

vadz:
OK, let's see if it works in #392.

PR #392, opened 2015-09-17
Restore using -pedantic but use -Wno-pedantic-ms-format too

vadz:
This reverts commit e607b1c731eb9058e8c627991e03c13a78ea5408 and replaces it
with the addition of -Wno-pedantic-ms-format option for MinGW as it suppresses
the warnings for using "%I64d" in format strings while still allowing us to
use -pedantic to detect other problems in the code.

See #369.

### DIFF
--- a/cmake/SociConfig.cmake
+++ b/cmake/SociConfig.cmake
@@ -51,7 +51,7 @@ if (MSVC)
 else()
 
   set(SOCI_GCC_CLANG_COMMON_FLAGS
-	"-Werror -Wall -Wpointer-arith -Wcast-align -Wcast-qual -Wfloat-equal -Wredundant-decls -Wno-long-long")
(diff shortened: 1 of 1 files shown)


snikulov:
@vadz rebase with latest master. I suspect @mloskot already fixed this warning with 8fe5d2a1df0a747f9990c39e8bcb6700a7723d9e

vadz:
Something is not right with AppVeyor, my branch does already have 8fe5d2a1df0a747f9990c39e8bcb6700a7723d9e, yet it seems to be not taken into account. Also, what's with MSVS 2014 compiler detection failure?

I'll just merge this because both errors are clearly unrelated to this change anyhow, but AppVeyor should still be fixed...

snikulov:
@vadz AppVeyor-CI log shows 
```core\connection-pool.cpp:287:12```
while @mloskot update core\connection-pool.cpp:98
So line 287 still contains uninitialized variable :)

mloskot:
I might have missed even more places :)

PR #393, opened 2015-09-17
mingw: warnings cleanup

snikulov:
Moved common code out of #ifdefs to reduce duplication.
Added default initialization for local values. 

[Build status](https://ci.appveyor.com/project/snikulov/soci)

### DIFF
--- a/cmake/SociConfig.cmake
+++ b/cmake/SociConfig.cmake
@@ -69,7 +69,7 @@ else()
             set(CMAKE_CXX_FLAGS "${CMAKE_CXX_FLAGS}")
         elseif (CMAKE_SYSTEM_NAME MATCHES "Windows")
             # We need to disable the warnings for using %I64d with MinGW.
-            set(CMAKE_CXX_FLAGS "${CMAKE_CXX_FLAGS} -Wno-pedantic-ms-format")
+            set(CMAKE_CXX_FLAGS "${CMAKE_CXX_FLAGS} -Wno-pedantic-ms-format -Wno-format")
         else()
             set(CMAKE_CXX_FLAGS "${CMAKE_CXX_FLAGS} -Wno-variadic-macros")
         endif()
--- a/src/backends/odbc/standard-use-type.cpp
+++ b/src/backends/odbc/standard-use-type.cpp
@@ -209,9 +209,9 @@ void odbc_standard_use_type_backend::bind_by_name(
 void odbc_standard_use_type_backend::pre_use(indicator const *ind)
 {
     // first deal with data
-    SQLSMALLINT sqlType;
-    SQLSMALLINT cType;
-    SQLLEN size;
+    SQLSMALLINT sqlType(0);
+    SQLSMALLINT cType(0);
+    SQLLEN size(0);
 
     void* const sqlData = prepare_for_bind(size, sqlType, cType);
 
--- a/src/backends/odbc/vector-use-type.cpp
+++ b/src/backends/odbc/vector-use-type.cpp
@@ -215,9 +215,9 @@ void odbc_vector_use_type_backend::bind_helper(int &position, void *data, exchan
     data_ = data; // for future reference
     type_ = type; // for future reference
 
-    SQLSMALLINT sqlType;
-    SQLSMALLINT cType;
-    SQLUINTEGER size;
+    SQLSMALLINT sqlType(0);
+    SQLSMALLINT cType(0);
+    SQLUINTEGER size(0);
 
     prepare_for_bind(data, size, sqlType, cType);
 
--- a/src/core/connection-pool.cpp
+++ b/src/core/connection-pool.cpp
@@ -83,27 +83,6 @@ connection_pool::~connection_pool()
     delete pimpl_;
 }
 
-session & connection_pool::at(std::size_t pos)
-{
-    if (pos >= pimpl_->sessions_.size())
-    {
-        throw soci_error("Invalid pool position");
-    }
-
-    return *(pimpl_->sessions_[pos].second);
-}
-
-std::size_t connection_pool::lease()
-{
-    // dummy default value avoids compiler warning, never leaks to client
-    std::size_t pos(0);
-
-    // no timeout, so can't fail
-    try_lease(pos, -1);
-
-    return pos;
-}
-
 bool connection_pool::try_lease(std::size_t & pos, int timeout)
 {
     struct timespec tm;
@@ -267,26 +246,6 @@ connection_pool::~connection_pool()
     delete pimpl_;
 }
 
-session & connection_pool::at(std::size_t pos)
-{
-    if (pos >= pimpl_->sessions_.size())
-    {
-        throw soci_error("Invalid pool position");
-    }
-
-    return *(pimpl_->sessions_[pos].second);
-}
-
-std::size_t connection_pool::lease()
-{
-    std::size_t pos;
-
-    // no timeout, allow unlimited blocking
-    try_lease(pos, -1);
-
-    return pos;
-}
-
 bool connection_pool::try_lease(std::size_t & pos, int timeout)
 {
     DWORD cc = WaitForSingleObject(pimpl_->sem_,
@@ -342,3 +301,26 @@ void connection_pool::give_back(std::size_t pos)
 }
 
 #endif // _WIN32
+
+session & connection_pool::at(std::size_t pos)
+{
+    if (pos >= pimpl_->sessions_.size())
+    {
+        throw soci_error("Invalid pool position");
+    }
+
+    return *(pimpl_->sessions_[pos].second);
+}
+
+std::size_t connection_pool::lease()
+{
+    // dummy default value avoids compiler warning, never leaks to client
+    std::size_t pos(0);
+
+    // no timeout, so can't fail
+    try_lease(pos, -1);
+
+    return pos;
+}
+
+


snikulov:
related to #369

cmake/SociConfig.cmake
```diff
@@ -69,7 +69,7 @@ else()
             set(CMAKE_CXX_FLAGS "${CMAKE_CXX_FLAGS}")
         elseif (CMAKE_SYSTEM_NAME MATCHES "Windows")
             # We need to disable the warnings for using %I64d with MinGW.
-            set(CMAKE_CXX_FLAGS "${CMAKE_CXX_FLAGS} -Wno-pedantic-ms-format")
+            set(CMAKE_CXX_FLAGS "${CMAKE_CXX_FLAGS} -Wno-pedantic-ms-format -Wno-format")
```
vadz: Why do we need `-Wno-format`? I thought it compiled fine now, disabling the format strings would be really unfortunate...

vadz:
As usual, it would be nice to split this commit in atomic parts... As it stands, I see nothing wrong with de-duplicating the code, probably not much wrong with initializing the variables even if it is unnecessarily (although in the past I had a lot of fun with finding code that wouldn't provoke neither "possibly uninitialized variable" warning from one compiler nor "unnecessary initialized variable/never used value" from other ones...) but would really like to avoid using `-Wno-format` which can catch bad bugs.

cmake/SociConfig.cmake
snikulov: @vadz As I already mentioned in #390 option -Wno-pedantic-ms-format make no sense for compilers with version greater or equal 4.7.1 
So, to fix this fast - I remove only for Windows this check. For Travis-CI it's still enforced.

Another option I can suggest - provide macro in soci-compiler.h (https://gcc.gnu.org/onlinedocs/cpp/Common-Predefined-Macros.html)
``` 
#define GCC_VERSION (__GNUC__ * 10000 \
                               + __GNUC_MINOR__ * 100 \
                               + __GNUC_PATCHLEVEL__)

```
and do check something like this

```
#if defined(_MSC_VER) || ( defined(__MINGW32__) && (GCC_VERSION < 40701) )
```
This should work.

vadz: Sorry, I'm completely lost, maybe it would be better to discuss this on the mailing list from the beginning as by now I have no idea what problem are we trying to solve any longer. All I can say is that:

1. `-Wno-pedantic-ms-format` suppresses the warning that is generated otherwise with `-pedantic` when using 4.8 and 4.9, so it's not clear at all why shouldn't it be used with 4.7+.
1. We already have `CHECK_GCC` macro so if we really need to check for patch level component we should just add it to the existing macro.

snikulov:
@vadz I'll resubmit as you like. Atomically.

cmake/SociConfig.cmake
jsonn: On Fri, Sep 18, 2015 at 08:28:42AM -0700, VZ wrote:
> > @@ -69,7 +69,7 @@ else()
> >              set(CMAKE_CXX_FLAGS "${CMAKE_CXX_FLAGS}")
> >          elseif (CMAKE_SYSTEM_NAME MATCHES "Windows")
> >              # We need to disable the warnings for using %I64d with MinGW.
> > -            set(CMAKE_CXX_FLAGS "${CMAKE_CXX_FLAGS} -Wno-pedantic-ms-format")
> > +            set(CMAKE_CXX_FLAGS "${CMAKE_CXX_FLAGS} -Wno-pedantic-ms-format -Wno-format")
> 
> Why do we need `-Wno-format`? I thought it compiled fine now, disabling the format strings would be really unfortunate...

Also, using %I64d is only valid for 32bit mode at least for MinGW64. It
would be nice to just depend on C++11 for Windows...

Joerg

snikulov: @jsonn You're right. But I've also checked MinGW64 v4.8.3 (i686 - 32 bit) it still fails with error

```
[ 78%] Building CXX object src/backends/odbc/CMakeFiles/soci_odbc.dir/standard-i
nto-type.cpp.obj
cd /d C:\WORK\GITHUB\soci\b.mgw32\src\backends\odbc && C:\MinGW\mingw-w64\i686-4
.8.3-posix-sjlj-rt_v3-rev2\mingw32\bin\g++.exe   -DBOOST_ALL_NO_LIB -DDBUG_OFF -
DHAVE_BOOST=1 -DHAVE_ODBC=1 -DSOCI_ABI_VERSION=\"4_0\" -DSOCI_DLL -pedantic -Wer
ror -Wall -Wpointer-arith -Wcast-align -Wcast-qual -Wfloat-equal -Wredundant-dec
ls -Wno-long-long -std=gnu++98  -Wno-pedantic-ms-format @CMakeFiles/soci_odbc.di
r/includes_CXX.rsp   -o CMakeFiles\soci_odbc.dir\standard-into-type.cpp.obj -c C
:\WORK\GITHUB\soci\src\backends\odbc\standard-into-type.cpp
C:\WORK\GITHUB\soci\src\backends\odbc\standard-into-type.cpp: In member function
 'virtual void soci::odbc_standard_into_type_backend::post_fetch(bool, bool, soc
i::indicator*)':
C:\WORK\GITHUB\soci\src\backends\odbc\standard-into-type.cpp:175:53: error: ISO
C++98 does not support the 'I' scanf flag [-Werror=format=]
           if (sscanf(buf_, "%" LL_FMT_FLAGS "d", &ll) != 1)
                                                     ^
C:\WORK\GITHUB\soci\src\backends\odbc\standard-into-type.cpp:175:53: error: form
at '%d' expects argument of type 'int*', but argument 3 has type 'long long int*
' [-Werror=format=]
C:\WORK\GITHUB\soci\src\backends\odbc\standard-into-type.cpp:183:53: error: ISO
C++98 does not support the 'I' scanf flag [-Werror=format=]
           if (sscanf(buf_, "%" LL_FMT_FLAGS "u", &ll) != 1)
                                                     ^
C:\WORK\GITHUB\soci\src\backends\odbc\standard-into-type.cpp:183:53: error: form
at '%u' expects argument of type 'unsigned int*', but argument 3 has type 'long
long unsigned int*' [-Werror=format=]
cc1plus.exe: all warnings being treated as errors
```

So, I agree. We'd better move from sprintf/printf to some C++ code instead of playing with ifdefs.

mloskot: @snikulov Obviously, we are not going to switch to C++11 soon, but I don't mind if we have private wrappers based on C++11 features for those troublemakers and we `#ifdef` activate them for MinGW exclusively.
On any other C++11 compilers, like VC++ and GCC, those wrappers can be activated with our CMake option `SOCI_CXX_C11`. By default, builds using VC++ and GCC should not require C++11.

snikulov: @mloskot what I'm trying to understand, why was @vadz using this check after exchange_type_cast in ODBC, while other back-ends does not do that way... 

Personally, I'm not big fan of MinGW, so I can disable it in Appveyor-CI permanently :)

mloskot: Hopefully, Vadim will be around to clarify.

vadz: Sorry, I'm continuing to be lost (see previous comment)... What check are we speaking about here?

What puzzles me is why `-Wno-pedantic-ms-format` doesn't work for @snikulov: here it definitely does, i.e. with it the warning I see with just `-pedantic` is not given any more when using MinGW-w64 4.9.1.

snikulov: @vadz could you please check this log https://ci.appveyor.com/project/mloskot/soci/branch/master/job/0cc8sb7ttjkbyegr ? What do you think? This code already have ```-Wno-pedantic-ms-format```

vadz: Are we sure the option is getting used there? As usual, I have no idea how to do anything with CMake, but perhaps there is a way to show the actual compilation lines with it?

If it is used, then the only explanation I have is that there was a bug in just 4.8 MinGW-W64 (as I have both 4.7 and 4.9 and they both work fine), but this would be pretty strange.

snikulov: @vadz Here the full command line in logs https://ci.appveyor.com/project/snikulov/soci/build/job/ltddtlpedtw4rth0 (added -DCMAKE_VERBOSE_MAKEFILE=ON)

Another question.
Is it possible to use same approach for odbc back-end as for mysql https://github.com/SOCI/soci/blob/master/src/backends/mysql/common.h#L57 
instead of sscanf ?

vadz: I found the reason for the mysterious discrepancy, i.e. why the same compiler with the same options compiled my test without problems but choked on SOCI code: this is due to `__USE_MINGW_ANSI_STDIO` being defined for SOCI but not for the simple example. And the reason for this is that including any C++ standard library headers ends up including `lib\gcc\i686-w64-mingw32\4.8.3\include\c++\i686-w64-mingw32\bits\os_defines.h` which contains
```
// Make sure that POSIX printf/scanf functions are activated.  As
// libstdc++ depends on POSIX-definitions of those functions, we define
// it unconditionally.
#undef __USE_MINGW_ANSI_STDIO
#define __USE_MINGW_ANSI_STDIO 1
```

So, long story short, the original fix replacing `%I64d` with `%lld` was in fact correct as the latter is what works with MinGW ANSI stdio implementation, while the former only works with the MSVC non-conforming-but-nevertheless-used-by-default-unless-a-standard-header-is-included (!) implementation.

Sorry for the confusion, I had no idea it was so twisted.

snikulov: @vadz I've submitted #398 which addressed rest of compilation issues. Could you please review?

PR #398, opened 2015-09-23
mingw: not use ms extension for MinGW G++ compilation

snikulov:
Fixed warnings on MinGW compilation (now all compilers in both CI are green).

@vadz Could you please review and comment?


mloskot:
Since this touches CMake, and @vadz prefers to not to touch CMake ;-), I'll join the review: looks good to me

snikulov:
@mloskot at least now builds are green. Need to integrate build with Boost and put tests into play.

mloskot:
Good. I'll take the liberty and merge it. Hopefully, @vadz will agree.

vadz:
Yes, absolutely, thank you.

Issue #379, opened 2015-08-24
Build VS2015 Win10 Errors


patodeborracha:
Hi,
I Try Build SOCI in VS2015 and Win10...:
https://gist.github.com/patodeborracha/a1fa994d3b511a9ef98e#file-gistfile1-txt

What i do?

Thank You!

vadz:
On Mon, 24 Aug 2015 05:03:41 -0700 patodeborracha <notifications@github.com> wrote:

p> I Try Build SOCI in VS2015 and Win10...:
p> https://gist.github.com/patodeborracha/a1fa994d3b511a9ef98e#file-gistfile1-txt
p> 
p> What i do?

 I thought this was fixed in 3.2 branch as well, but it definitely was in
the master, so I'd recommend getting the latest sources from git and using
them instead, especially if you're just starting with SOCI.

 Regards,
VZ

patodeborracha:
I'll Try it..

Thank You!

mloskot:
I haven't tried to build SOCI with VS2015 yet, so I might have missed some of the fixes in branch `release/3.2`. However, I'll be happy to apply any fixes for VS2015 to both, the master and 3.2.

snikulov:
@patodeborracha Could you please retest on latest master? 
I suspect https://github.com/SOCI/soci/commit/bb79b6074879aaef5643521e414417fcafdd7352 should fix your problem.

PR #399, opened 2015-09-23
Update README.md

snikulov:
Use table with build statuses for branches (master and release/3.2)
We also can put appveyor.yml on release/3.2 and then update readme with build status for that branch.

### DIFF
--- a/README.md
+++ b/README.md
@@ -1,6 +1,14 @@
 SOCI - The C++ Database Access Library
 ======================================
-[![Build Status](https://travis-ci.org/SOCI/soci.svg?branch=master)](https://travis-ci.org/SOCI/soci) [![Build status](https://ci.appveyor.com/api/projects/status/ree3vx3kfgs23qk6/branch/master?svg=true)](https://ci.appveyor.com/project/mloskot/soci/branch/master)
+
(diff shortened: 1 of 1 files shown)


mloskot:
@snikulov Good idea. Feel free to enable Appveyor for release/3.2.

However, if it's too much hassle, I would not waste time on release/3.2.
As per explanation on https://github.com/SOCI/soci/wiki/Branches, it is not actively maintained, no further 3.x release are planned or promised.

snikulov:
@mloskot Ok. But if we have no exact plans for 4.0.0 release date, 3.2.x still the best stable release, I suspect.

mloskot:
It depends. I'd say this way:
* master (SOCI 4): we keep it healthy and usable, so consider it as a **failry stable rolling release**, though we haven't published it officially as packaged release.
* release/3.2: is stable, but it is becoming **deprecated**, so due to lack of manpower, we are not actively working on it anymore, so we encourage people to switch to the master now - it would also help to test it and stabilise it faster

snikulov:
@mloskot no problem. Can we close #379 since it is not exactly true? 
We can see that "Build VS2015" successful and no more response from submitter available?

mloskot:
Closed. Thx

Issue #374, opened 2015-08-08
Clang warnings preventing compilation


musopr:
Hi,

From soci-sqlite3.h I'm getting a number of occurrences of:
warning: anonymous types declared in an anonymous union are an extension [-Wnested-anon-types]

And from type-conversion.h, a number of occurrences of:
warning: field 'ownInd_' is uninitialized when used here [-Wuninitialized]

I can't seem to be able to pass in their negations via -DCMAKE_CXX_FLAGS either, which seems to be due to the ordering of the flags.

I've gotten around this by editing SOCI_GCC_CLANG_COMMON_FLAGS in SociCongif.cmake, on my machine, but I'm not sure if this is the preferred way to handle this.

mloskot:
I haven't checked the actual code, but any chance to get rid of those warnings by fixing the code?

snikulov:
@mloskot Clang emit uninitialized warnings, which seems to me serious enough. 

Anonymous structures/unions was introduced not long ago with @ArnaudD-FR  in f1f0162bd9844a40b3533c4f036a370adf79f264 and we can just remove it with ```-Wno-anon-types``` or name it :)

mloskot:
@snikulov I'm in favour of naming any unonymous type.

snikulov:
@musopr should be fixed as of https://github.com/SOCI/soci/commit/91fb559f04c3ac3253208763259249a5a6f683d6
If problem still exist, please submit another issue.

musopr:
Thank you @snikulov, it compiles okay now.

PR #402, opened 2015-09-24
Name struct to avoid clang error/warning (issue #374)

ArnaudD-FR:
This is just a partial commit for issue #374. On my environment clang makes error and no warning.

So the following errors/warnings remains:
`soci4/include/soci/bind-values.h:91:39: error: C++98 requires an accessible copy constructor for class 'soci::details::use_type_vector::use_sequence<boost::fusion::vector<double, boost::optional<MyInt>, std::basic_string<char>, boost::fusion::void_,
      boost::fusion::void_, boost::fusion::void_, boost::fusion::void_, boost::fusion::void_, boost::fusion::void_, boost::fusion::void_>, void>' when binding a reference to a temporary; was private [-Werror,-Wbind-to-temporary-copy]`

and

`soci4/include/soci/type-conversion.h:91:68: error: field 'ownInd_' is uninitialized when used here [-Werror,-Wuninitialized]
        : use_type<base_type>(details::base_value_holder<T>::val_, ownInd_, name)`

### DIFF
--- a/include/soci/sqlite3/soci-sqlite3.h
+++ b/include/soci/sqlite3/soci-sqlite3.h
@@ -157,6 +157,16 @@ struct sqlite3_vector_use_type_backend : details::vector_use_type_backend
     std::string name_;
 };
 
+struct sqlite3_column_buffer
+{
+    std::size_t size_;
+    union
+    {
+        const char *constData_;
+        char *data_;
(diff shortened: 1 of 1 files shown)


mloskot:
Thanks!

Issue #337, opened 2015-05-18
Missing static ODBC object file in installation script 


Superlokkus:
When I built the current master branch with mingw on windows with only following cmake variables set in SOCI and WITH (rest default)

	SOCI
		SOCI_STATIC
		SOCI_ODBC
	WITH
		WITH_ODBC
		WITH_THREADS
		
libsoci_odbc_4_0.a is missing in the installation script thus is missing in the lib64 directory, altough it has been built.

Resulting install_manifest.txt:
```
C:/Program Files (x86)/SOCI/include/soci/backend-loader.h
C:/Program Files (x86)/SOCI/include/soci/bind-values.h
C:/Program Files (x86)/SOCI/include/soci/blob-exchange.h
C:/Program Files (x86)/SOCI/include/soci/blob.h
C:/Program Files (x86)/SOCI/include/soci/boost-fusion.h
C:/Program Files (x86)/SOCI/include/soci/boost-gregorian-date.h
C:/Program Files (x86)/SOCI/include/soci/boost-optional.h
C:/Program Files (x86)/SOCI/include/soci/boost-tuple.h
C:/Program Files (x86)/SOCI/include/soci/connection-parameters.h
C:/Program Files (x86)/SOCI/include/soci/connection-pool.h
C:/Program Files (x86)/SOCI/include/soci/error.h
C:/Program Files (x86)/SOCI/include/soci/exchange-traits.h
C:/Program Files (x86)/SOCI/include/soci/into-type.h
C:/Program Files (x86)/SOCI/include/soci/into.h
C:/Program Files (x86)/SOCI/include/soci/noreturn.h
C:/Program Files (x86)/SOCI/include/soci/once-temp-type.h
C:/Program Files (x86)/SOCI/include/soci/prepare-temp-type.h
C:/Program Files (x86)/SOCI/include/soci/procedure.h
C:/Program Files (x86)/SOCI/include/soci/query_transformation.h
C:/Program Files (x86)/SOCI/include/soci/ref-counted-prepare-info.h
C:/Program Files (x86)/SOCI/include/soci/ref-counted-statement.h
C:/Program Files (x86)/SOCI/include/soci/row-exchange.h
C:/Program Files (x86)/SOCI/include/soci/row.h
C:/Program Files (x86)/SOCI/include/soci/rowid-exchange.h
C:/Program Files (x86)/SOCI/include/soci/rowid.h
C:/Program Files (x86)/SOCI/include/soci/rowset.h
[...]

PR #403, opened 2015-09-24
fixed typo

DraconPern:
static libs not installed because of typo


mloskot:
Good catch, thanks!

mloskot:
Fixed #337

PR #410, opened 2015-10-02
Appveyor: added build for MySQL backend

snikulov:
- enabled mysql backend in build.
- disabled mysql tests, because it fails with segfault.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -86,9 +86,6 @@ if(NOT MSVC)
     set_directory_properties(PROPERTIES INCLUDE_DIRECTORIES ${DL_INCLUDE_DIR})
     add_definitions(-DHAVE_DL=1)
   endif()
-else() #MSVC
-  # This flag enables multi process compiling for Visual Studio 2005 and above
-  SET(CMAKE_CXX_FLAGS "${CMAKE_CXX_FLAGS} /MP")
 endif()
 
 if(Boost_FOUND)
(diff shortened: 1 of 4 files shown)


CMakeLists.txt
```diff
@@ -86,9 +86,6 @@ if(NOT MSVC)
     set_directory_properties(PROPERTIES INCLUDE_DIRECTORIES ${DL_INCLUDE_DIR})
     add_definitions(-DHAVE_DL=1)
   endif()
-else() #MSVC
-  # This flag enables multi process compiling for Visual Studio 2005 and above
-  SET(CMAKE_CXX_FLAGS "${CMAKE_CXX_FLAGS} /MP")
```
mloskot: It's not that `/MP` flag is critical, I'm just curious why it needs to be removed.

snikulov: @mloskot If you think it's important, will return it back. Anyway cmake do almost job for build and I just trying reduce number of non portable scripts.

pfedor:
Hi,

I have a question about


   - disabled tests, because it fails with segfault.


What is the segfault? Sorry, if there has been an email I missed it.

Thanks,

Aleksander

snikulov:
@pfedor here with enabled tests https://ci.appveyor.com/project/snikulov/soci/build/4.0.0.104/job/a24pvj13is1t8iwm
I've no time to dig into much, so I just put it with disabled tests for now.

CMakeLists.txt
mloskot: On 2 Oct 2015 18:03, "Sergei Nikulov" <notifications@github.com> wrote:
> @mloskot If you think it's important, will return it back. Anyway cmake
do almost job for build and I just trying reduce number of non portable
scripts.
>

No need to restore it.
In fact, cutting down on number of CMake script lines is a very good idea
(including any custom macros or logic scripted specifically for SOCI).

PR #411, opened 2015-10-05
fix clang build warning, fix cmake generated file that is not ignored

jeking3:


snikulov:
:-1: I see no reason to merge TODO into source.

mloskot:
Instead it would be better to fix it straight away.

mloskot:
Closing as it is better to fix it properly, see #412

jeking3:
There was already a TODO in the code to remove those methods.  Given it was allowed before, I figured I would submit my workaround that will allow clang users to make forward progress.  In any case, I'll use this workaround locally while I tackle more important things.  I don't know where the SOCI bug database is, but this needs to be entered into the bug database so someone more familiar with the project can fix the infrastructure here.  It is not a trivial change.

mloskot:
@jeking3 There is nothing with TODOs in the code. We just think it is unnecessary to add new ones via PRs (notice, those existing TODOs have been there for long time, too long).

SOCI bug database is here. We use [GitHub Issues](https://guides.github.com/features/issues/) as a bug database. So, I have created bug entry in #412.

snikulov:
@jeking3 Simple workaround for clang is remove -Werror from SociConfig.cmake 
HTH.

Issue #383, opened 2015-09-08
std::strings in nvarchar(max) is still a problem


picant:
Reading string trough ODBC version 11 still does not work for nvarchar(max) columns (yes I'm aware of issue #287). ODBC version 10 works just fine.

Environment:
windows 8.1
msvc 2013
soci version 4 (from 8-Sep-2015)

Database setup:
use history
create table test ( data nvarchar (MAX) not null );
insert into test (data) values ('Aaa')

Test source code:

```
#define SOCI_USE_BOOST
#include <soci.h>            // SOCI components
#include <soci-config.h>
#include <odbc/soci-odbc.h>
#include <afxcontrolbars.h>
using namespace soci;

//const char* connectionString = "DRIVER={SQL Server};Server=192.168.11.54\\SPSQL;Uid=sa;Pwd=Master10"; //<<<<this works
const char* connectionString = "DRIVER={ODBC Driver 11 for SQL Server};Server=192.168.11.54\\SPSQL;Uid=sa;Pwd=Master10"; //<<<<THIS DOESN'T WORK

    session sql;
    sql.open(connection_parameters("odbc", connectionString));

    sql << "use history";

    rowset<row> rs = sql.prepare << "select * from test";
    for (auto it = rs.begin(); it != rs.end(); ++it)
    {
        std::string data;
        data = it->get<std::string>("data"); //this returns empty string if ODBC 11 is used
    }
```

vadz:
Does our [existing test case](https://github.com/SOCI/soci/commit/c025cc86968914bd4bd0523dee11cdd9db991aba#diff-917ce54c9a333ec67257b8d86dd712baR23) fail as well? If not, could you please try to update the test case to show the failure in it?

picant:
I can't build your tests as I run into all kind of build problems if I use -DSOCI_TESTS=ON :(

Looking at your test code I can suggest to replace 'fixed_text nvarchar(4000)' with 'fixed_text nvarchar(max)' in your test test-odbc-mssql.cpp and check if the test still runs fine if you use ODBC driver version 11. Or even better try to replace 'nvarchar(num)' and 'varchar(num)' with 'nvarchar(max)' everywhere in your odbc tests.

vadz:
Strange, what problems are those? You may want to open another issue for them...

picant:
Done. Created issue #384 about the build problems.

toeb:
[...]

PR #413, opened 2015-10-06
AppVeyor: enable MSSQL ODBC connection

snikulov:
Enabled MSSQL tests.
Some code modification in tests was introduced.
@vadz @mloskot Could you please review?


snikulov:
@vadz BTW, here the exposure for #287 and #383. Test was failed on nvarchar.
I believe it related to http://stackoverflow.com/questions/10025032/what-is-the-meaning-of-the-prefix-n-in-t-sql-statements 
One of idea how to fix it - is use N'< here the long string >'

vadz:
Sorry, I'm really not sure what is the idea here: why do we need to create a new database just for MS SQL and just for this test? And why do we need yet another helper class, could we reuse `test_context_base` for this instead?

snikulov:
@vadz Because I'm limited with CI environment. It is not my or your local PC. 
For PostgreSQL I can create database before test run from cmd console.
With MSSQL I don't know how to do this. So I prepare environment inside the test.

If you provide guideline how to reuse test_context_base and reuse same approach approach for "// MS SQL-specific tests" I'll change my code per your suggestion.

snikulov:
@vadz Personally, I use fixtures for that. But didn't see same approach anywhere in test code.

vadz:
Normally [sqlcmd.exe](https://msdn.microsoft.com/en-us/library/ms162773.aspx) should be available in the MS SQL installation, so it should be possible to use it to create the test database and avoid doing it in the code entirely.

But maybe creating the database from the code is actually better, it's just that I don't like the inconsistency of doing it just for MS SQL but not the others. IMHO if we want to change this, we should do it for all of them. And `test_context` is basically already a fixture, so we should use it for this.

snikulov:
@vadz Ok. Thank you. :)
I also find sqlcmd but need some time to check how it'll work. Will update my pull shortly.
Creating also good, but I suspect it can be server specific, and probably harder.

snikulov:
@vadz I've closed this pull per your comments. New pr #415 submitted.
Now code is unchanged. Used sqlcmd for test preparation.

PR #415, opened 2015-10-07
AppVeyor: enable MSSQL ODBC connection

snikulov:
Re-worked #413 per @vadz comments.
Test code unchanged. Used command line tool for db creation.


bin/windows/mssql_db_create.sql
```diff
@@ -0,0 +1 @@
+IF db_id('soci_test') IS NULL BEGIN CREATE DATABASE [soci_test] END;
```
vadz: Is the test for the database existence really necessary? AFAIU we start from a clean slate every time with AppVeyor, so it shouldn't already exist -- and if it somehow does, it could presumably contain something that could interfere with our tests, so perhaps it would be better to delete and recreate it?

vadz:
Great, thank you, I think it's much better to do it like this.

mloskot:
Great effort @snikulov !

bin/windows/mssql_db_create.sql
snikulov: Just sanity check :)

snikulov:
@mloskot Thank you :) 
But now we have failed tests.
One exposes #383 and one MySQL problems on Windows.

PR #423, opened 2015-10-09
Coverity fixes

vadz:
Fixes for minor Coverity bug reports, mostly just to check whether the next report will be updated to take these fixes into account.

### DIFF
--- a/src/backends/postgresql/row-id.cpp
+++ b/src/backends/postgresql/row-id.cpp
@@ -31,7 +31,7 @@ using namespace soci::details;
 postgresql_rowid_backend::postgresql_rowid_backend(
     postgresql_session_backend & /* session */)
 {
-    // nothing to do here
+    value_ = 0;
 }
 
 postgresql_rowid_backend::~postgresql_rowid_backend()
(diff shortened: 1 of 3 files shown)


src/backends/postgresql/row-id.cpp
```diff
@@ -31,7 +31,7 @@ using namespace soci::details;
 postgresql_rowid_backend::postgresql_rowid_backend(
     postgresql_session_backend & /* session */)
 {
```
snikulov: Maybe init list? https://isocpp.org/wiki/faq/ctors#init-lists
```
postgresql_rowid_backend::postgresql_rowid_backend(postgresql_session_backend & /* session */)
    : value_()
{
}
```

vadz: Good point, initializer lists are indeed more SOCI style. I still prefer to keep an explicit 0 however for clarity.

vadz:
Committed as 2af6c5efc0d12168391f629f8fd851dd58a448a5 (AppVeyor failures are unrelated to this change).

Issue #142, opened 2013-04-24
double free 


nikunjsingh11:
Hi,

My applications crashes while trying to execute a soci statement:

double free

```
lib/libc.so.6(+0x71e16)[0x7fd293defe16]
/lib/libc.so.6(cfree+0x6c)[0x7fd293df4b8c]
/usr/local/lib/libsoci_postgresql.so.3.1(_ZN4soci36postgresql_standard_use_type_backend8clean_upEv+0x36)[0x7fd295170a60]
/usr/local/lib/libsoci_postgresql.so.3.1(_ZN4soci36postgresql_standard_use_type_backend8post_useEbPNS_9indicatorE+0x2c)[0x7fd295170a28]
/usr/local/lib/libsoci_core.so.3.1(_ZN4soci7details17standard_use_type8post_useEb+0x3f)[0x7fd2953f1a81]
/usr/local/lib/libsoci_core.so.3.1(_ZN4soci7details14statement_impl8post_useEb+0x5a)[0x7fd2954302dc]
/usr/local/lib/libsoci_core.so.3.1(_ZN4soci7details14statement_impl7executeEb+0x3c6)[0x7fd29542f69e]
./marketFeeder(_ZN4soci9statement7executeEb+0x26)[0x48ef68]
```

Is there a way I can avoid the above failure.

vadz:
This seems to have been fixed by #424, so closing.

PR #424, opened 2015-10-12
type_ptr update

snikulov:
Make type_ptr more safe.
Current implementation contains default copy constructor/assignment which can lead to double free issue (for example #142).

### DIFF
--- a/include/soci/type-ptr.h
+++ b/include/soci/type-ptr.h
@@ -8,6 +8,8 @@
 #ifndef SOCI_TYPE_PTR_H_INCLUDED
 #define SOCI_TYPE_PTR_H_INCLUDED
 
+#include <algorithm>
+
 namespace soci { namespace details {
 
 template <typename T>
@@ -18,9 +20,38 @@ public:
     ~type_ptr() { delete p_; }
 
     T * get() const { return p_; }
-    void release() const { p_ = 0; }
+    T * release() const
+    {
+        T * tmp = 0;
+        std::swap(tmp, p_);
+        return tmp;
+    }
+
+    // rhs will loose it ownership
+    type_ptr& operator=(const type_ptr& rhs)
+    {
+        // move sematics
+        std::swap(this->p_, rhs.p_);
+        if (rhs.p_)
+        {
+            delete rhs.p_;
+            rhs.p_ = 0;
+        }
+    }
+
+    // rhs will loose it ownership
+    type_ptr(const type_ptr& rhs)
+        : p_(rhs.p_)
+    {
+        if (rhs.p_)
+        {
+            // avoid double free
+            rhs.p_ = 0;
+        }
+    }
 
 private:
+
     mutable T * p_;
 };
 
--- a/include/soci/bind-values.h
+++ b/include/soci/bind-values.h
@@ -30,7 +30,7 @@ public:
             delete *iter;
     }
 
-    void exchange(use_type_ptr const& u) { push_back(u.get()); u.release(); }
+    void exchange(use_type_ptr const& u) { push_back(u.release()); }
 
     template <typename T, typename Indicator>
     void exchange(use_container<T, Indicator> const &uc)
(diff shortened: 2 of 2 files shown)


snikulov:
@mloskot @vadz Could you please review and comment?
Current implementation allows following usage https://github.com/snikulov/soci/commit/8d2f10181d1125d7c61ff418511b46452aa79c3c which can produce double free.
Make it non-copyable will break compilation, since C++98 does not allow move semantics. 
Since we don't use C++11, I disable compiler defined copy/assign operations.

include/soci/type-ptr.h
```diff
@@ -18,9 +20,43 @@ class type_ptr
...
+        : p_(rhs.p_)
+    {
+        if (rhs.p_)
+        {
+            // avoid double free
+            rhs.p_ = 0;
+        }
+    }
+
+    operator bool() const
```
mloskot: https://en.wikibooks.org/wiki/More_C%2B%2B_Idioms/Safe_bool

mloskot:
My question would be, do we need `type_ptr<Person> pt3 = pt1;` use case?
`type_ptr` is a kind of shared smart pointer, but it is very very minimal, I guess, such use case and/or move semantics was not considered as required.

include/soci/type-ptr.h
snikulov: @mloskot I'll remove it.

snikulov:
@mloskot I just try conform to following guideline "Make interfaces easy to use correctly and hard to use incorrectly" http://www.aristeia.com/Papers/IEEE_Software_JulAug_2004_revised.htm
Current implementation can be used incorrectly, even if we don't want this.

mloskot:
@snikulov  I agree with you and that is a good idea, in general.

My point was that AFAIK, type_ptr is hidden implementation detail in SOCI (it lives in `soci::details::` namespace).
It means, we as SOCI developers are sole users of `type_ptr`.
AFAIU, core developers can take the risk and should know how to use it correctly.

Is there any actual issue to solve, apart from inconveniences coming from limited `type_ptr` features?

snikulov:
I try reduce number of reasons for possible memory leaks and double free, which still present in issue list.
Because, I suspect, type_ptr not sole use of developers. It leaks into user interface through typedef indirectly.

vadz:
I'm not sure how does this type "leak" into the public API, but I'd say it's probably a problem if it does.

As for the changes themselves, `release()` is IMHO completely uncontroversial. I'm much less sure about implementing the custom assignment operator and copy ctor though: if we do it like this, why not just `std::auto_ptr<>` explicitly, as it seems to become the same thing? At the very least, assignment should IMO be forbidden at all, do we need it anywhere?

include/soci/type-ptr.h
```diff
@@ -18,9 +20,38 @@ class type_ptr
...
+        // move sematics
+        std::swap(this->p_, rhs.p_);
+        if (rhs.p_)
+        {
+            delete rhs.p_;
+            rhs.p_ = 0;
+        }
+    }
+
+    // rhs will loose it ownership
```
vadz: s/loose/lose/ (also above if it doesn't go away)

snikulov:
@vadz Because auto_ptr is deprecated for modern compilers.
So we just break SOCI for example on MSVC, where we unable to disable c++11 with set of switches. 

BTW, I suspect, I have similar issue on MSVC with nothrow for destructor. MSVC will terminate program even if nothrow(false) set for destructor ``` ~once_temp_type() SOCI_ONCE_TEMP_TYPE_NOEXCEPT; ```.

vadz:
Sorry, "use `auto_ptr`" was shorthand for "we should use `unique_ptr<>` where available (check for `__cplusplus` and `_MSC_VER` separately) or `auto_ptr` otherwise".

snikulov:
@vadz @mloskot So, what's the decision? 
Should we remove type_ptr completely?

mloskot:
I'll summary my point:
* The proposed addition to `soci::details::type_ptr` does not make it ideal, less shoot-yourself-in-foot prone.
* "std::auto_ptr or std::unique_ptr in C++11 mode", as I mentioned on Gitter
* Do we really need assignment? Indeed, I see adding it is not going to solve any issues, so we "ain't gonna need it" :)
* If `soci::details::type_ptr` does not deliver any added feature, comparing to `std::auto/unique_ptr`, then it is a good idea to replace it with the standard smart pointer(s).

PR #425, opened 2015-10-13
moved SOCI_CXX_C11 define to common section

snikulov:
Added definition for all compilers when generator setting ```-DSOCI_CXX_C11``` provided.
Once type used only when SOCI_CXX_C11 defined.


mloskot:
:+1:

PR #428, opened 2015-10-20
reset count of affected rows (sqlite3)

MonsieurNicolas:
when reusing prepared statements, the count was not reset, causing subsequent calls to `execute` to set `get_affected_rows` wrong.

Here is sample code that was failing:

```C++
session << "CREATE TABLE IF NOT EXISTS kv("
        "k   CHARACTER(32) PRIMARY KEY,"
        "v       TEXT);";

    session
        << "INSERT INTO kv (k, v) VALUES ('a', '2')";

    soci::statement st(session);
    st.alloc();

    std::string k = "a";
    std::string v = "1";

    st.clean_up(false);

    st.prepare("UPDATE kv SET v = :v WHERE k = :k;");
    st.exchange(soci::use(v));
    st.exchange(soci::use(k));
    st.define_and_bind();
    st.execute(true);

    if (st.get_affected_rows() != 1)
    {
        throw std::runtime_error("could not update");
    }

    std::string k2 = "b";

    st.clean_up(false);

    st.exchange(soci::use(v));
    st.exchange(soci::use(k2));
    st.define_and_bind();
    st.execute(true);

    // update should not do anything as 'b' does not exist
    if (st.get_affected_rows() == 1)
    {
        throw std::runtime_error("should not happen");
    }
```

### DIFF
--- a/src/backends/sqlite3/statement.cpp
+++ b/src/backends/sqlite3/statement.cpp
@@ -233,6 +233,8 @@ sqlite3_statement_backend::bind_and_execute(int number)
 
     long long rowsAffectedBulkTemp = 0;
 
+    rowsAffectedBulk_ = -1;
+
     int const rows = static_cast<int>(useData_.size());
     for (int row = 0; row < rows; ++row)
     {
--- a/tests/common-tests.h
+++ b/tests/common-tests.h
@@ -3832,6 +3832,10 @@ TEST_CASE_METHOD(common_tests, "Get affected rows", "[core][affected-rows]")
     st1.execute(true);
     CHECK(st1.get_affected_rows() == 1);
 
+    // attempts to run the query again, no rows should be affected
+    st1.execute(true);
+    CHECK(st1.get_affected_rows() == 0);
+
     statement st2 = (sql.prepare <<
         "update soci_test set val = val + 1");
     st2.execute(true);


MonsieurNicolas:
failures seems to be unrelated to this change: I see 
```
4/14 Test  #4: soci_mysql_test_static ..............***Exception: SegFault
```

```
C:\projects\soci\tests\odbc\test-odbc-mssql.cpp(23): FAILED:
  Error binding input parameter #1: [Microsoft][SQL Server Native Client 10
  .0]Invalid precision value (SQL state HY104) while executing "insert into
  soci_test(long_text) values(:str)" with :str="Line #0
```

snikulov:
Yes, these are long standing failures on Windows platform.

mloskot:
AFAICT, the patch is sound, isn't it? :+1: thanks!

snikulov:
@mloskot AFAIK, there is common test for get_affected_rows(...) here https://github.com/SOCI/soci/blob/master/tests/common-tests.h#L3813 which is not failed on master.

Perhaps, it should be added another test to expose issue described.

mloskot:
@snikulov Yes, good point. @MonsieurNicolas could you update the PR adding new test for your use case?

MonsieurNicolas:
I added the test, but it looks like db2 has a different bug in there: an update that doesn't update any row throws

mloskot:
Ok, lets get it in.
Thanks @MonsieurNicolas

mloskot:
It seems, this also reveales a subtle bug in DB2. To be checked

Issue #433, opened 2015-10-29
Fix discrepency between SOCI_USE_BOOST and HAVE_BOOST


mloskot:
The issue has been revealed in #429 
Between 3.2.2 and 3.2.3, role of `HAVE_BOOST` has changed and it disconnected from `SOCI_USE_BOOST` legacy.

In SOCI 3.2.3 (and current master too) `HAVE_BOOST` controls the Boost support in SOCI.
The `SOCI_USE_BOOST` is still dangling in the all-in-one header `soci.h` and in fact, it should be automatically defined if `HAVE_BOOST` is defined - that is missing, unfortunately.

Sources and CMake need to rely on single definition that will also be required while building SOCI clients.

Documentation needs updating as it mentions the old `SOCI_USE_BOOST` only.

ArnaudD-FR:
Even if role of `HAVE_BOOST` has changed I don't think keeping it is a good idea. This macro name is too generic. `SOCI_USE_BOOST` has a better name.

I checked on SOCI 3.2 branch and both macro exist, that maybe why I made this mistake and introduced this regression.

We will have to to support both macro for a while or use version 4.0 to correctly handle this.

ArnaudD-FR:
Lets continue with my latest message from issue #429.

I think that SOCI users shouldn't care about having correct definitions when they build a program using SOCI library and it can cause error if they forget one.

So the idea is to automatically put all definitions in a single file explaining how the library as been built.

Currently my commit ArnaudD-FR@882bff3a2903c9a1688ce8028734ba2bc421a6a5 is an example and only show  the SOCI_USE_BOOST macro definition but this is the main idea.

mloskot:
On 29 October 2015 at 21:53, ArnaudD-FR <notifications@github.com> wrote:

> Even if role of HAVE_BOOST has changed I don't think keeping it is a good
> idea. This macro name is too generic. SOCI_USE_BOOST has a better name.
>

I was thinking of replacing both with new macro SOCI_HAVE_BOOST and place
it in generated config header.
The HAVE indicates clearly, particular build/package of SOCI has Boost
support built-in.


> We will have to to support both macro for a while or use version 4.0 to
> correctly handle this.
>

As far as I'm concerned, SOCI 4.0 can wipe old macros and introduce new
ones.

For SOCI 3.x, I'm not in favour of doing anything, or perhaps just update
docs in release/3.2 branch to indicate requirement of HAVE_BOOST while
building a client.

[...]

PR #435, opened 2015-11-03
Specify definitions in soci-config.h

ArnaudD-FR:
* replace all commmand line definition as -DHAS_BOOST, -DHAS_SQLITE3 by
   soci-config.h


ArnaudD-FR:
related to issue Issue #433

CMakeLists.txt
```diff
@@ -89,25 +89,18 @@ if(NOT MSVC)
 endif()
 
 if(Boost_FOUND)
-  get_property(SOCI_COMPILE_DEFINITIONS
-    DIRECTORY ${CMAKE_CURRENT_SOURCE_DIR}
-    PROPERTY COMPILE_DEFINITIONS)
-
-  list(APPEND SOCI_COMPILE_DEFINITIONS "HAVE_BOOST=1")
-  list(APPEND SOCI_COMPILE_DEFINITIONS "BOOST_ALL_NO_LIB")
+  set(SOCI_HAVE_BOOST Y)
```
snikulov: Can we use ON/OFF for Boolean variables as described here https://cmake.org/cmake/help/v3.4/prop_cache/TYPE.html?highlight=bool? I know it is more verbose, but it is also more readable.

CMakeLists.txt
```diff
@@ -89,25 +89,18 @@ if(NOT MSVC)
 endif()
 
 if(Boost_FOUND)
-  get_property(SOCI_COMPILE_DEFINITIONS
-    DIRECTORY ${CMAKE_CURRENT_SOURCE_DIR}
-    PROPERTY COMPILE_DEFINITIONS)
-
-  list(APPEND SOCI_COMPILE_DEFINITIONS "HAVE_BOOST=1")
-  list(APPEND SOCI_COMPILE_DEFINITIONS "BOOST_ALL_NO_LIB")
```
snikulov: This removal is breaks Windows linkage.

CMakeLists.txt
```diff
@@ -163,5 +166,22 @@ if(SOCI_TESTS)
   add_subdirectory(tests)
 endif()
 
+###############################################################################
+# build config file
+###############################################################################
```
snikulov: Cool idea to build config from any SOCI_HAVE variable defined in scope, but I prefer be more explicit in my intentions.

So, IMO, soci-config.h.in should contains all explicit variables which should be defined using either
```#cmakedefine SOCI_HAVE_VAR ${SOCI_HAVE_VAR}```
or
```#cmakedefine SOCI_HAVE_VAR @SOCI_HAVE_VAR@```

which allows to document any defines in template and in result file.

@mloskot What do you think? I'm not insist on it, but usable for documentation purposes.

CMakeLists.txt
ArnaudD-FR: No problem

CMakeLists.txt
ArnaudD-FR: On windows I also got an error on link but when calling cmake if I set `Boost_USE_STATIC_LIBS=ON` everything is fine.

CMakeLists.txt
ArnaudD-FR: I started to write has you said but I feel that someday it will be unsynchronised. That's why I switch to this "auto" mode.

But this is a minor risk, and this is easy to switch. So @mloskot what do you prefer?

CMakeLists.txt
snikulov: Well, I know about static. It is second options. 
But current changes breaks Boost dynamic lib linkage which was fine before. 
@mloskot If it's ok - I'll be fine with this.

ArnaudD-FR: I just checked the meaning of `BOOST_ALL_NO_LIB`, I was thinking it was soci specific. So I keep it

snikulov: :+1:  Thanks! It fixed now.

CMakeLists.txt
mloskot: You both present no very strong valid points. I have no strong opinion but since we already have been hit by bug due to missed maintenance of such definitions I lean towards @ArnaudD-FR 's auto mode. But I'd like to see

1. Clear naming convention E. g. `SOCI_HAVE_*` only for generated configuration variables.

2. Way to dump or list all such variables automatically so it is easy to write them on docs without searching through sources, missing any.

I can write docs in markdown for them.

ArnaudD-FR: Now I handle only cache variable starting with `SOCI_HAVE_`. I'm using CACHE_HELPSTRING property to add some help in generated header

mloskot: On 5 Nov 2015 15:06, "ArnaudD-FR" <notifications@github.com> wrote:
>
> In CMakeLists.txt:
>
> > @@ -163,5 +166,22 @@ if(SOCI_TESTS)
> >    add_subdirectory(tests)
> >  endif()
> >
> >
+###############################################################################
> > +# build config file
> >
+###############################################################################
>
> Now I handle only cache variable starting with SOCI_HAVE_. I'm using
CACHE_HELPSTRING property to add some help in generated header
>

Looks good to me. I'll merge when I'm back home or @snikulov feel free to
do it if you see no objections.

Thank you!

snikulov:
Merged.

PR #436, opened 2015-11-05
CMake generates soci/version.h

ArnaudD-FR:
During my last cmake review I discovered that SOCI version was defined in both root CMakeLists.txt and include/soci/version.h

This is a quick fix to avoid forgetting one of them: include/soci/version.h is now generated based on what is defined in root CMakeLists.txt


snikulov:
LGTM.

mloskot:
Sweet fix!

PR #438, opened 2015-11-07
DB2: Switch session from SQLConnect ot SQLDriverConnect

mloskot:
SQLConnect is very basic, relies on DSN and is not convenient
if user wants to provide all connection parameters as command line
argument, without editing db2cli.ini, configuring DSN.

SQLDriverConnect alternative enables remote database connection
in quick and easy fashion, via complete connection string.

This is required to run DB2 tests on Vagrant VMs, where SOCI is built on `vmsoci` and DB2 instance is on separate `vmdb2` and tests connect via network.

### DIFF
--- a/docs/backends/db2.md
+++ b/docs/backends/db2.md
@@ -85,6 +85,19 @@ TODO
 
 None.
 
-### <a name="config"></a> Configuration options
+### <a name="configuration"></a> Configuration options
 
-None.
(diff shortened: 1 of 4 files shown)


mloskot:
The PR is sound as per [Build 860](https://travis-ci.org/SOCI/soci/builds/89826554)

PR #443, opened 2016-02-03
postgres library built as static would not install unless also buildi…

BlakBat:
postgres library built as static would not install unless also building shared version


snikulov:
Nice catch :+1: 
@mloskot it was fixed on master with 4fab0bb33a0593f0e78a3443d57b9529403a9b93

PR #444, opened 2016-02-03
Include soci/soci-config.h only if it is available

vadz:
Also include it only once, from soci/soci-platform.h, as both of these files
serve similar purposes.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -147,6 +147,7 @@ set(CONFIG_INCLUDE_DIR ${CMAKE_CURRENT_BINARY_DIR}/include)
 install(DIRECTORY ${CONFIG_INCLUDE_DIR}/soci DESTINATION ${INCLUDEDIR})
 set(CONFIG_FILE_IN "include/soci/soci-config.h.in")
 set(CONFIG_FILE_OUT "${CONFIG_INCLUDE_DIR}/soci/soci-config.h")
+add_definitions(-DHAVE_SOCI_CONFIG_H)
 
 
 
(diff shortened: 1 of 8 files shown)


ArnaudD-FR:
Hi vadz, thanks for the patch but purpose to add `soci/soci-config.h` (PR #435) was to remove any SOCI definition from soci users point of view. In your patch SOCI users have to define `HAVE_SOCI_CONFIG_H` macro when they want to use the configuration file generated by CMake.

Adding `add_definitions(-DHAVE_SOCI_CONFIG_H)` in `CMakeLists.txt` makes this definition visible only in SOCI project, not for users depending on SOCI.

As you want to use SOCI without CMake, why don't you just create a `soci/soci-config.h` with all your definitions? like `wx/setup.h` with wxWidgets?

The main problem in your case will come from `soci/version.h`, maybe the best is, as you suggested in mailing list, to read this file from CMake, this can be done easily with CMake.

vadz:
OK, you're right, sorry for not thinking this through, defining `SOCI_NO_CONFIG_H` on my end indeed seems to be the only option and maybe it's not worth it finally... I do have my own `soci-config.h` already, of course, just to be able to build the latest SOCI, I just hoped to get rid of it but, again, you're right and it's not such a big deal and `version.h` is more of a problem. Unfortunately I have no idea how to extract values from it from CMake...

P.S. AppVeyor seems to be broken :-(

ArnaudD-FR:
I will try to find some time to provide a patch this week for `version.h`.

PR #446, opened 2016-02-04
CMake read version from version.h

ArnaudD-FR:
As discussed in mailing list, code should be sufficient. So lets read SOCI version from `include/soci/version.h` instead of defining it in root CMakeLists.txt


mloskot:
LGTM, thank you!

I'll leave the final merge decision to @vadz

vadz:
Thanks a lot @ArnaudD-FR!

Issue #441, opened 2016-01-12
SOCI doesn't define SOCI_HAVE_CXX_C11 for VS2015


Cleroth:
VS2015 seems to support every C++11 in SOCI, but still doesn't get that defined. I just spent an entire day figuring out why I wasn't able to catch exceptions from SOCI queries. The reason? SOCI is throwing from a destructor even when it's not supported (or thinks it's not supported).

mloskot:
Regardless of compiler features, SOCI builds in C++98 mode by default.

Cleroth:
Isn't it pretty useless if you can't catch any exceptions in C++98? Any query error results in abort()/terminate() being called.

mloskot:
`SOCI_HAVE_CXX_C11` enables or disables C++11 features *on-demand*.

In case of VS2015, the problem is it does not support C++98 only mode as GCC or clang do, so the `noexcept(false)` needs to be implicitly enabled, indeed. VS2015 support is incomplete, apparently.

Cleroth:
I see. Honestly I'm not sure if the problem occurs in `master`, as it seems it's a bit different in that aspect from the latest release (3.2.3). If the problem still happens, there really should be something done about it as it makes SOCI pretty much unusable with VS2015.

All I did to fix it was the noexcept declaration to this:
    #if __cplusplus >= 201103L || (defined(_MSC_FULL_VER) && _MSC_FULL_VER >= 190023026)
    #define SOCI_ONCE_TEMP_TYPE_NOEXCEPT noexcept(false)

Since VS's __cplusplus is still defined to C++98 (because its C++11 features are not complete), but the noexcept is fully implemented, then this should work just fine (and it does for me).

mloskot:
Please, consider submitting a pull request. Thanks!

Cleroth:
I only changed it on 3.2.3, not on trunk. I don't really know how to make it so that `SOCI_HAVE_CXX_C11` is set for VS (I don't have much knowledge in CMake and such configs).
VS could probably benefit from C++11 on the all of SOCI but I don't know how to do that. It really isn't as important as the breaking case of exceptions being thrown in a destructor and terminating the program, though. So what should I do? Just commit that one line that enables noexcept for VS?

ckaminski:
I know this is old and fixed but I'd like to add the following for the Googlebot: 

    terminate called after throwing an instance of 'soci::postgresql_soci_error'

[...]

PR #448, opened 2016-02-04
Handle MSVS 2015 (and 2013, to some extent) as C++11 compiler

vadz:
Notably this fixes the problem of crashing as soon as `once_temp_type` dtor throws, see #441.

### DIFF
--- a/include/soci/once-temp-type.h
+++ b/include/soci/once-temp-type.h
@@ -8,15 +8,9 @@
 #ifndef SOCI_ONCE_TEMP_TYPE_H_INCLUDED
 #define SOCI_ONCE_TEMP_TYPE_H_INCLUDED
 
-#include "soci/soci-config.h"
+#include "soci/soci-platform.h"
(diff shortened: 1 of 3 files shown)


mloskot:
Looks good, thanks!

vadz:
Does anybody understand what's going on with Travis? The PR builds succeeded but master is now failing for some reason...

snikulov:
@vadz you pull request was not successful according to ```X``` against ffc564b hash.
While compilation is good, test for clang produced binaries are broken.

vadz:
The failing Travis build 869 corresponds to `ffc564b70d96ea13fa64c134e0b5b74fe0afd720` on master, the build for the PR had succeeded before (it was green when I merged and even right now it still says "1 check failed", which corresponds to the AppVeyor failure, and implies that Travis had succeeded).

Anyhow, I could reproduce this failure locally with clang, so I'll debug this soon...

PR #449, opened 2016-02-05
fixed #447

### LINKED ISSUE #447: AppVeyor can not find sh.exe?
Looks like AV setup got broken, but I'm not sure what is going on:
```
del /Q "%ProgramFiles(x86)%\Git\bin\sh.exe"
The system cannot find the path specified.
Command exited with code 1
```
Do we really aim to delete the executable?

Sergei, could you have a look or confirm what needs to be done and I'll try?

snikulov:
Updated scripts to check if sh.exe exists and delete it before MinGW build generation.

### DIFF
--- a/appveyor.yml
+++ b/appveyor.yml
@@ -41,23 +41,36 @@ cache:
 
 install:
     - ps: |
-          if (!(Test-Path C:\projects\mingw))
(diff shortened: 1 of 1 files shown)


mloskot:
Thank you!

snikulov:
@mloskot Anyway it not makes AV green.

PR #450, opened 2016-02-05
Avoid recursive inclusion between soci-{config,platform}.h

vadz:
This resulted in SOCI_HAVE_CXX_C11 not being defined in soci-platform.h if
soci-config.h was the first header to be included, as soci-platform.h is
included from it before defining this symbol, but uses it.


snikulov:
Now Linux build fixed :+1:

PR #451, opened 2016-02-05
Include soci/soci-platform.h instead of soci-config.h

vadz:
This is less confusing and ensures that common macros such as
SOCI_NOT_COPYABLE() are always defined.

Also remove an unneeded inclusion from session.cpp as the header is already
included from session.h anyhow.


mloskot:
:+1:

PR #452, opened 2016-02-05
Define cxx_details::auto_ptr<> to get rid of preprocessor checks

vadz:
Defining cxx_details::auto_ptr<> as either std::auto_ptr<> or
std::unique_ptr<> once instead of checking for SOCI_HAVE_CXX_C11 simplifies
the code and also allows to use unique_ptr<> for MSVS 2010+ which has it even
when SOCI_HAVE_CXX_C11 is not defined.

Also use cxx_details::auto_ptr<> instead of std::auto_ptr<> in the tests to
fix the build in C++11 mode with clang 3.8 due to -Wdeprecated-declarations
warnings given by it for std::auto_ptr<> (and -Werror used by CMake and which
apparently can't be overridden from command line...).

### DIFF
--- a/include/soci/rowset.h
+++ b/include/soci/rowset.h
@@ -8,7 +8,7 @@
 #ifndef SOCI_ROWSET_H_INCLUDED
 #define SOCI_ROWSET_H_INCLUDED
 
-#include "soci/soci-config.h"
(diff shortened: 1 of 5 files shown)


mloskot:
:+1:

PR #453, opened 2016-02-11
updated catch.hpp version to 1.3.3

snikulov:
- Updated Catch C++ UT framework to latest release
- Disabled compilation failure for GCC case with parentheses (```-Wno-error=parentheses```)

### DIFF
--- a/cmake/SociConfig.cmake
+++ b/cmake/SociConfig.cmake
@@ -49,7 +49,7 @@ if (MSVC)
 else()
 
   set(SOCI_GCC_CLANG_COMMON_FLAGS
-	"-pedantic -Werror -Wall -Wpointer-arith -Wcast-align -Wcast-qual -Wfloat-equal -Wredundant-decls -Wno-long-long")
+	"-pedantic -Werror -Wno-error=parentheses -Wall -Wpointer-arith -Wcast-align -Wcast-qual -Wfloat-equal -Wredundant-decls -Wno-long-long")
 
 
   if (SOCI_CXX_C11)
--- a/tests/catch.hpp
+++ b/tests/catch.hpp
@@ -1,6 +1,6 @@
 /*
- *  Catch v1.2.1
- *  Generated: 2015-06-30 18:23:27.961086
+ *  Catch v1.3.3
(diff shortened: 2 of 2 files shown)


tests/mysql/test-mysql.cpp
```diff
@@ -297,7 +297,7 @@ void test_num(const char* s, bool valid, T value)
         {
             char const * expectedPrefix = "Cannot convert data";
             CAPTURE(e.what());
-            CHECK(strncmp(e.what(), expectedPrefix, strlen(expectedPrefix)) == 0);
+            CHECK( (strncmp(e.what(), expectedPrefix, strlen(expectedPrefix)) == 0) );
```
vadz: We have a few occurrences of `strncmp(s, foo, strlen(foo)) == 0` in the tests, I think it would make sense to add a helper `starts_with(s, foo)` function which could be used instead, this would make the code more clear and less error-prone and also avoid the problem with the parentheses (I'm surprised that it happens only here, what about the other places `strncmp(...) == 0` is used?)

snikulov: No. It happens multiple times.
I just add -Wno-error=parenthesis to postpone this for another fix.

snikulov: BTW, I also think about something like
```
template< typename T>
bool test_eq(T a, T b)
{
    return a == b;
}
```
or just use std::equals_to()(one, two);

vadz: `test_eq()` would break CATCH expression decomposition. In the particular case of `strncmp() == 0` it doesn't matter anyhow because decomposition doesn't work with it anyhow, but for anything else it would be a regression.

From the point of view of decomposition, the best solution is actually
```
CHECK( s.substr(0, strlen(expected)) == expected );
```
And if we do this, we could get rid of `CAPTURE(s)` before it.

snikulov:
@vadz OK to merge?

vadz:
Maybe undo the parentheses addition, just to avoid inconsistencies between the different places?

But yes, sure, OK to merge anyhow.

snikulov:
@vadz OK. Will remove it before merge.

PR #454, opened 2016-02-11
added Address Sanitizer to expose memory errors

snikulov:
- added option to enable Address Sanitizer (```-DSOCI_ASAN=ON/OFF```) OFF by default.
- added SOCI_BUILD_BACKEND=asan for Travis-CI.

Notes:
- Current check is allowed to fail.
- gcc v4.6.x installed on Travis is not supporting this feature, so it not used
- clang is builds with this feature and exposes issue with reconnect() after close() session for MySQL described in #136. This also the reason why Windows Build segfaults. Calling close on session is leave library with dangling pointer to factory https://github.com/SOCI/soci/blob/master/tests/mysql/test-mysql.cpp#L540


vadz:
Using ASAN is a very good idea, but I don't understand why does it need to be a "backend"? This doesn't seem to make much sense and seems to need a lot of new scripts.

Why not just have `env` option `WITH_ASAN` or something like this instead?

snikulov:
Well, to get meaningful from back-trace it should be build with 
```-DCMAKE_BUILD_TYPE=Debug```

If it's OK if all backends will be build in Debug mode, then I can remove specific target.

vadz:
I am almost sure that ASAN requires debug info (`-g` and maybe `-fno-omit-frame-pointer`?) but can still work with release/production build (`-O2`) otherwise while I think CMake debug build also uses `-O0`, doesn't it?

Ideal would be to use `-g -O2 -fsanitize=address`, but if we can't, I don't think there is any big problem with just using debug build neither.

snikulov:
Currently it ```-fsanitize=address -g``` for my asan target and no ```-g``` on release builds

vadz:
Strange, do we disable optimizations for Travis builds to speed up compilation time somewhere?

Anyhow, my main point is that we shouldn't have a separate ASAN target but use ASAN with all (or as many as we can) backends. IMO it's not a problem at all to use `-g` for all of them.

snikulov:
Ok. As you wish.
I'll remove target and will use 
```option(SOCI_ASAN "Enable address sanitizer on GCC v4.8+/Clang v 3.1+" ON)```
**ON** by default

snikulov:
@vadz BTW, I can up GCC to 4.8 on Travis.
Should I do that, or we'll stay with 4.6 ?

vadz:
Sorry, do you mean that you'll turn ASAN on by default for all builds? If so, I definitely don't think this would be the right thing to do, we should turn it on explicitly for Travis builds only. If you meant to turn it on for all Travis builds, then yes, I agree.

As for 4.8: I think it could make sense to have (some?) builds using 4.6 and some using 4.8 with C++11. But I'd prefer to do it in a separate change if possible. TIA!

snikulov:
@vadz OK. Done. Removed asan target.
Explicitly enabled for all Travis-CI builds (but possible only on Clang, because of old Gcc).
Travis-CI will now be RED, because of error found in MySQL/ODBC tests.

vadz:
Breaking Travis is not nice as we won't notice any other breakage until this is fixed. IMO we need to either skip the currently failing tests (with a big fat "FIXME" comment in the code) or *temporarily* avoid using ASAN for MySQL/ODBC.

snikulov:
@vadz Well, doing it green by disabling checking not solve the issue.
It gives you false sense of security. 
I prefer to see proper state with, then just thinking that all OK.

vadz:
I understand your point of view but the trouble is that we don't know when is this going to be fixed (I don't use `reconnect()` myself so I can't promise to look at this) and until it is, all the other problems/regressions risk passing unnoticed. It's really more about testing future PRs (which are all going to fail) than the state of the current code itself which is, indeed, broken.

snikulov:
@vadz we can remove .close() from test. This will fix the issue with mysql test.
BTW, I prefer to move our internal chat into gitter room. What do you think?

vadz:
I prefer to remove it from/disable it in the test until it's fixed (still with a big comment explaining why is it disabled).

gitter: Sorry, I don't use it (yet?) and, anyhow, I think it's useful to keep the record of what happened here in the ticket itself. And for anything longer there is always the mailing list...

mloskot:
I assume there is work needed around the  tests and the big comment too.

Once updated, please rebase on master to update this PR and trigger Travis CI builds would be enough, and if builds are green, I'm OK to merge.

snikulov:
@mloskot Could you please merge? Then I just fix travis.yml with allowed failures.
Unfortunately I removed SOCI clone with this branch, so I unable to rebase, but in fact this change still in sync with upstream.

mloskot:
@snikulov done.

You could merge too, since you have full access too :)

PR #455, opened 2016-02-12
compilation fix for msvc2012

snikulov:
Introduced in  #452.
MSVC 2012 not supported type aliases ``` using auto_ptr =  ... ```.
So I restrict it to more recent compiler MSVC 2013.

### DIFF
--- a/include/soci/soci-platform.h
+++ b/include/soci/soci-platform.h
@@ -98,7 +98,7 @@ namespace soci
 namespace cxx_details
 {
 
-#if defined(SOCI_HAVE_CXX_C11) || (defined(_MSC_VER) && _MSC_VER >= 1600)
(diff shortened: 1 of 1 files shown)


vadz:
Thanks for fixing this! Just for clarity though: `_MSC_VER == 1800` is MSVC 12 a.k.a. MSVS 2013, so the commit message is not completely correct.

snikulov:
Not sure what are you talking about. Commit states that it was fix for msvc 2012, not 12.

vadz:
It's just the naming. There is no such thing as MSVC 2012, there is only MSVS ("Studio" as opposed to the "Compiler") 2012. The versions of MSVS (VC) are: 2003 (7), 2005 (8), 2008 (9), 2010 (10), 2012 (11), 2013 (12), 2015 (14).

PR #462, opened 2016-04-08
Fix inclusion via SOCI_POSTGRESQL_SOURCE

julien-lecomte:
On branch "release/3.2" there's an include issue when compiling via:

```
cmake -G "Unix Makefiles" -DCMAKE_BUILD_TYPE=Debug -DSOCI_SHARED=OFF -DSOCI_STATIC=ON -DWITH_BOOST=ON -DWITH_ORACLE=OFF -DWITH_DB2=OFF -DWITH_FIREBIRD=OFF -DWITH_MYSQL=OFF -DSOCI_TESTS=OFF -DWITH_ODBC=OFF -DWITH_SQLITE3=OFF -DWITH_POSTGRESQL=ON -DWITH_EMPTY=ON -DCMAKE_VERBOSE_MAKEFILE=ON -DCMAKE_INSTALL_PREFIX=/usr/local/ ../soci/src
```

After `make` and `sudo make install`, inclusion of `soci/postgresql/soci-postgresql.h` will try to include `soci-backend.h` but the latter is installed as `soci/soci-backend.h`

This "fix" is done via on the preprocessor value `SOCI_POSTGRESQL_SOURCE`

Sample program to check for the inclusion error:

```
#include <soci/soci.h>
#include <soci/postgresql/soci-postgresql.h>
int main() { return 0; }
```

### DIFF
--- a/src/backends/postgresql/soci-postgresql.h
+++ b/src/backends/postgresql/soci-postgresql.h
@@ -24,7 +24,11 @@
 # define SOCI_POSTGRESQL_DECL
 #endif
 
+#ifdef SOCI_POSTGRESQL_SOURCE
 #include <soci-backend.h>
+#else
+#include <soci/soci-backend.h>
+#endif
 #include <libpq-fe.h>
 #include <vector>
 


mloskot:
@BlakBat Thanks, but could you have a look at the PostgreSQL build failing here https://travis-ci.org/SOCI/soci/builds/121714211 and see if it is related to your PR?

vadz:
I don't think we're going to apply this to 3.2 and the problem doesn't seem to exist in 4.0, so I believe we can close it.

PR #474, opened 2016-05-16
Adding example of a oracle connection to a remote server without need of tnsnames.ora

wdavilaneto:
Adding example of a oracle connection to a remote server without need of tnsnames.ora


snikulov:
Merged. Thank you!

PR #475, opened 2016-05-17
coverity // Fixed uninited local std::tm structs

snikulov:
Fixes for coverity issues related uninitialized local structs.


mloskot:
:+1: Thanks

PR #479, opened 2016-06-03
fixed #478

### LINKED ISSUE #478: delete already deleted pimpl if connection_pool is assigned
Using this code will crash:
```c++
connection_pool pool(1);
pool = connection_pool(10);
```

This happens because the default assignment operator assigns the pimpl_ pointer of the connection_pool class. So it gets deleted twice.

snikulov:
Disabled copy constructor/assignment operator by default to avoid usage as described in #478

### DIFF
--- a/include/soci/connection-pool.h
+++ b/include/soci/connection-pool.h
@@ -32,6 +32,8 @@ public:
 private:
     struct connection_pool_impl;
     connection_pool_impl * pimpl_;
+
+    SOCI_NOT_COPYABLE(connection_pool)
 };
 
 }


snikulov:
@mloskot @vadz Could you please review? Do you agree with fix or should we allow connection_pool copying?

mloskot:
Looks good to me.

Alternative, if necessary for some users, `connection_pool` could be copyable with shared/ref-counted reference to the implementation. 

However, in SOCI, we have worked out a convention that public major classes are non-copyable.

vadz:
I agree, IMHO there is no need to make it copyable. Thanks!

PR #480, opened 2016-06-07
Portable metadata queries.

msobczak:
This PR is an interface extension for portable metadata queries.
Assuming 'sql' is a session object, this is possible:

    // get all table names in user's schema
    std::string name;
    soci::statement st = (sql.prepare_table_names(), into(name));
    // st.execute/while-fetch/etc.

    // get column descriptions for the given table name:
    soci::column_info ci;
    soci::statement st = (sql.prepare_column_descriptions(table_name), into(ci))
    // st.execute/while-fetch/etc.

(bulk versions for vector<string> are also supported, see the diff)
where column_info is defined as:

struct column_info
{
    std::string name;
    data_type type;
    std::size_t length; // meaningful for text columns only
    std::size_t precision;
    std::size_t scale;
    bool nullable;
};

The implementation assumes standard SQL information schema and PostgreSQL is used as a reference. Other backends do not need to do anything if they follow the standard. Oracle has a separate implementation since it is not standard-compliant in this regard - it is enough for the backend to override a set of backend functions that compose schema queries to redefine the standard behaviour.

This PR introduces one new header file. It *might* be necessary to fix build scripts appropriately.

### DIFF
--- a/include/soci/column-info.h
+++ b/include/soci/column-info.h
@@ -1,0 +1,117 @@
+//
+// Copyright (C) 2016 Maciej Sobczak
+// Distributed under the Boost Software License, Version 1.0.
+// (See accompanying file LICENSE_1_0.txt or copy at
+// http://www.boost.org/LICENSE_1_0.txt)
+//
+
+#ifndef SOCI_COLUMN_INFO_H_INCLUDED
+#define SOCI_COLUMN_INFO_H_INCLUDED
+
+#include "soci/soci-backend.h"
+#include "soci/type-conversion.h"
+#include "soci/values.h"
+
+namespace soci
+{
+
+struct SOCI_DECL column_info
+{
+    std::string name;
+    data_type type;
+    std::size_t length; // meaningful for text columns only
+    std::size_t precision;
+    std::size_t scale;
+    bool nullable;
+};
+
+std::size_t get_numeric_value(const values & v,
+    const std::string & field_name)
+{
+    data_type dt = v.get_properties(field_name).get_data_type();
+    switch (dt)
+    {
+    case dt_double:
+        return static_cast<std::size_t>(
+            v.get<double>(field_name, 0.0));
+    case dt_integer:
+        return static_cast<std::size_t>(
+            v.get<int>(field_name, 0));
+    case dt_long_long:
+        return static_cast<std::size_t>(
+            v.get<long long>(field_name, 0ll));
+    case dt_unsigned_long_long:
+        return static_cast<std::size_t>(
+            v.get<unsigned long long>(field_name, 0ull));
+        break;
+    default:
+        return 0u;
+    }        
+}
+
+template <>
+struct type_conversion<column_info>
+{
+    typedef values base_type;
+    
+    static void from_base(values const & v, indicator /* ind */, column_info & ci)
+    {
+        ci.name = v.get<std::string>("COLUMN_NAME");
+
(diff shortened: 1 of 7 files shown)


include/soci/column-info.h
```diff
@@ -0,0 +1,117 @@
...
+{
+    std::string name;
+    data_type type;
+    std::size_t length; // meaningful for text columns only
+    std::size_t precision;
+    std::size_t scale;
+    bool nullable;
+};
+
+std::size_t get_numeric_value(const values & v,
```
vadz: Is this going to be a public API? If not, it should be in the `details` sub-namespace.

include/soci/soci-backend.h
```diff
@@ -234,6 +234,28 @@ class session_backend
...
+    // queried in a portable way - backends that are standard compliant
+    // do not need to override the following methods, which are intended
+    // to return a proper query for basic metadata statements.
+    virtual std::string get_table_names_query() const
+    {
+        return "select table_name as \"TABLE_NAME\""
+            " from information_schema.tables"
+            " where table_schema = 'public'";
+    }
+    virtual std::string get_column_descriptions_query() const
```
vadz: Should have a comment documenting that this query takes a parameter (`t`) as it's easy to miss it otherwise.

include/soci/column-info.h
vadz: Is this going to be a public API? If not, it should be in the `details` sub-namespace.

vadz:
Thanks, I'm sure this is going to be useful, but as I'm not sure how/what for exactly, I can't really comment on the API.

Of course, as always, it would be nice (indispensable?) to have the tests and documentation for this feature. TIA!

include/soci/column-info.h
msobczak: This is a public API.

include/soci/soci-backend.h
msobczak: Comment added in:

https://github.com/msobczak/soci/commit/a05b61f1f5ffe85e077f6739cb13a83d6c2bbdd1

mloskot:
Apologies for late comment, but this is related and important since some of our CI builds are failing now:

https://travis-ci.org/SOCI/soci/builds/137353265
```
standard-use-type.cpp:(.text+0x0): 
  multiple definition of `soci::get_numeric_value(soci::values const&, 
    std::basic_string<char, std::char_traits<char>, std::allocator<char> > const&)'
```

Shouldn't `get_numeric_value` be inline?

msobczak:
Fixed in https://github.com/SOCI/soci/commit/a0917b7c81c4e0d33b7fc78609d5637a8aacd5ff

The get_numeric_value is a helper function that was accidentally put at the namespace scope, while it belongs to the type_conversion instantiation, where it is uniquely used.

cstiborg:
I'm aware this is 8 years ago, but as this code hasn't been changed since and on the odd chance that someone remembers I'd like to ask about the decision to have a hardcoded filter on table_schema = 'public'.
This decision means if you use a schema for your tables... you lose. I understand that sqlite3 doesn't support schemas and mysql/mariadb is the odd one out with its concept of database. I also understand why is it necessary to make sure that the table is unique per table. However, I'm struggling to see why you can't have a dynamic schema, possibly just as default argument to the prepare methods taking the value of 'public'.
Am I missing anything or was it just enough at the time?

PR #482, opened 2016-06-13
Added singlerows mode for PostgreSQL.

msobczak:
I would like to propose a new feature for PostgreSQL - query execution in the single-row mode.
From the docs:

In addition to standard PostgreSQL connection parameters, the following can be set:

* `singlerow` or `singlerows`

For example:

    session sql(postgresql, "dbname=mydatabase singlerows=true");

If the `singlerows` parameter is set to `true` or `yes`, then queries will be executed in the single-row mode, which prevents the client library from loading full query result set into memory and instead fetches rows one by one, as they are requested by the statement's fetch() function. This mode can be of interest to those users who want to make their client applications more responsive (with more fine-grained operation) by avoiding potentially long blocking times when complete query results are loaded to client's memory.

Note that this feature is performance-oriented and as such has no influence on the base functionality - except that bulk queries are not supported as they do not make any sense in such operation mode.

### DIFF
--- a/docs/backends/postgresql.md
+++ b/docs/backends/postgresql.md
@@ -60,6 +60,20 @@ To establish a connection to the PostgreSQL database, create a `session` object 
 
 The set of parameters used in the connection string for PostgreSQL is the same as accepted by the `[PQconnectdb](http://www.postgresql.org/docs/8.3/interactive/libpq.html#LIBPQ-CONNECT)` function from the `libpq` library.
 
+In addition to standard PostgreSQL connection parameters, the following can be set:
+
+* `singlerow` or `singlerows`
+
+For example:
+
+    session sql(postgresql, "dbname=mydatabase singlerows=true");
+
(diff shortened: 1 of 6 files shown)


mloskot:
Although this looks good to me, it would be nice if the CI builds stay green.

vadz:
I wonder if there is a better way to separate the single mode handling from the normal case because having to use `if`s to check for it everywhere makes the code significantly messier. E.g. could we have 2 backend classes with different implementations instead?

msobczak:
Vadim - I was thinking about separating the code paths and there are three options:
1. create separate backend class (statement backend would be enough), perhaps with some common base in the middle,
2. create separate implementations of only those functions that are actually affected (prepare/execute/fetch - for example prepare_single, prepare_multi, etc.), and dispatch to them based on the flag
3. inject ifs where appropriate.

Unfortunately for all three, the amount of affected code is actually very small when compared to the whole code. The bulk of the code is common and deals with preparation of query parameters, data buffers, error handling, etc. For this reason, creating separate classes would be a definitive overkill and even a common base class would not reduce the duplication of preparation-related code in the affected functions. More importantly, it would disrupt the structure of the backend, which is now uniform across other databases (that is, you know what to expect in each file).
Having separate versions of each function was more attractive, but again the amount of preparation and buffer handling code would lead to massive duplication.
I consider a bunch of ifs to be the least intrusive and actually minimal in terms of editing and future maintenance (note that the original flow is preserved), even though at first look they look like being "everywhere".

PR #484, opened 2016-06-20
Added support for portable DDL statements.

msobczak:
This PR proposes a portable API for basic DDL statements.
The idea is best described by the diff of the docs/statement.md file.

Since PostgreSQL is the most standard with regard to its SQL dialect, it was chosen as a basis. Other backends do not have to do anything if they are compliant, otherwise they need to override methods from the base backend interface in order to generate appropriate target SQL code. This was already done for Oracle.

### DIFF
--- a/docs/statements.md
+++ b/docs/statements.md
@@ -5,6 +5,7 @@
 * [Bulk operations](#bulk)
 * [Stored procedures](#procedures)
 * [Transactions](#transactions)
+* [Portable DDL](#ddl)
 * [Metadata queries](#metadata)
 * [Basic logging support](#logging)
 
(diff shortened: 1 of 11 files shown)


mloskot:
Handy :+1:

docs/statements.md
```diff
@@ -290,6 +291,58 @@ A typical usage pattern for this class might be:
...
+
+SOCI supports some basic methods to construct portable DDL queries. That is, instead of writing explicit SQL statement for creating or modifying tables, it is possible to use dedicated SOCI functions, which prepare appropriate DDL statements behind the scenes, thus enabling the user application to create basic database structures in a way that is portable across different database servers. Note that the actual support for these functions depends on the actual backend implementation.
+
+It is possible to create a new table in a single statement:
+
+   sql.create_table("t1").column("i", soci::dt_integer).column("j", soci::dt_integer);
+
+Above, table "t1" will be created with two columns ("i", "j") of type integer.
+
+It is also possible to build such a statement piece by piece:
```
vadz: Just to avoid any confusion, I'd make it more clear that this piece-by-piece example is not exactly the same as the one liner above.

include/soci/once-temp-type.h
```diff
@@ -109,6 +109,48 @@ class prepare_type
 
 } // namespace details
 
+// Note: ddl_type is intended to be used just as once_type_type,
```
vadz: Typo `s/once_type_type/once_temp_type/`

include/soci/once-temp-type.h
```diff
@@ -109,6 +109,48 @@ class prepare_type
...
+    ddl_type & primary_key(const std::string & name,
+        const std::string & columnNames);
+    ddl_type & foreign_key(const std::string & name,
+        const std::string & columnNames,
+        const std::string & refTableName,
+        const std::string & refColumnNames);
+
+    ddl_type & operator()(const std::string & arbitrarySql);
+
+    void set_tail(const std::string & tail);
```
vadz: Would be nice to have a comment explaining what does this do, it's not obvious from just looking at the header.

include/soci/oracle/soci-oracle.h
```diff
@@ -278,6 +279,77 @@ struct oracle_session_backend : details::session_backend
             " where table_name = :t";
     }
     
+    virtual std::string create_column(const std::string & columnName, data_type dt,
+        int precision, int scale)
+    {
+        //  Oracle-specific SQL syntax:
+        
+        std::string res = columnName + " ";
```
vadz: Perhaps we should always put quotes around the column name to avoid possible conflicts with SQL keywords?

vadz:
This definitely looks very nice, thank you!

The only real comment I have is that we really should have some unit tests for the new functionality IMO.

include/soci/oracle/soci-oracle.h
jsonn: How does that interact with case sensitiveness?

msobczak: Let the users add quotes on their own, they have impact on later queries.
Note also that this feature is not intended to enable random junk or something that come from user input as column names (this does not make sense in DDL anyway), so we can assume that column names will be chosen wisely and most likely by the application developers.

You other comments were fixed in the new commit.

PR #486, opened 2016-06-29
Added failover callback for Oracle and PostgreSQL.

msobczak:
Hi all,

In this PR I would like to propose the failover capability, implemented for Oracle (where it is supported by the native OCI interface, especially in RAC configurations) and PostgreSQL (where it is emulated by the backend).
The idea is that when the database connection gets broken, the backend can notify the user about that fact and allow him to guide the automated reconnection process. It is possible to instruct the backend to retry to the same or to another destination, as well as replay necessary session configuration sequence.

The comments for functions in new file callbacks.h explain it all.

(no, don't ask for unit tests...)

### DIFF
--- a/include/soci/callbacks.h
+++ b/include/soci/callbacks.h
@@ -1,0 +1,47 @@
+//
+// Copyright (C) 2015 Maciej Sobczak
+// Distributed under the Boost Software License, Version 1.0.
+// (See accompanying file LICENSE_1_0.txt or copy at
+// http://www.boost.org/LICENSE_1_0.txt)
+//
+
+#ifndef SOCI_CALLBACKS_H_INCLUDED
+#define SOCI_CALLBACKS_H_INCLUDED
+
+namespace soci
+{
+
+class session;
+
+// Simple callback interface for reporting failover events.
+// The meaning of each operation is intended to be portable,
+// but the behaviour details and parameters can be backend-specific.
+class SOCI_DECL failover_callback
+{
+public:
+
+    // Called when the failover operation has started,
+    // after discovering connectivity problems.
+    virtual void started() {}
+
+    // Called after successful failover and creating a new connection;
+    // the sql parameter denotes the new connection and allows the user
+    // to replay any initial sequence of commands (like session configuration).
+    virtual void finished(session & /* sql */) {}
+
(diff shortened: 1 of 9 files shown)


include/soci/callbacks.h
```diff
@@ -0,0 +1,47 @@
...
+// Simple callback interface for reporting failover events.
+// The meaning of each operation is intended to be portable,
+// but the behaviour details and parameters can be backend-specific.
+class SOCI_DECL failover_callback
+{
+public:
+
+    // Called when the failover operation has started,
+    // after discovering connectivity problems.
+    virtual void started() {}
```
vadz: It looks like this and other methods are always called inside `try/catch(...)`, so IMHO it would make sense to make them protected and have public `call_started()` and so on which would call the real callback and eat the exceptions from it.

This would reduce the amount of `try/catch`s in the rest of the code and ensure that they can't be forgotten, letting the exceptions escape.

vadz:
Thanks, this looks like another useful addition!

No real comments, but it looks like the PostgreSQL implementation is almost completely generic, just `PQstatus() == BAD_CONNECTION` part is really backend-specific, so perhaps we could abstract just this check and make the rest generic and reusable from the other backends? I think it would be useful to have this functionality available for most of them (with the exception of SQLite).

And while I understand that unit testing this is difficult, it would be still great to have at least a small example (in the docs as we don't have any real examples?).

include/soci/callbacks.h
msobczak: No. The failover_callback is a classical interface, which does not enforce any policy with regard to exception handling. It is the choice of the backend whether the calls to this interface should be protected by try/catch block or not and I can imagine a backend where this is not necessary. I can also imagine a user who decides to use the same interface to implement some forwarding or notification mechanism at the application level, and it is also not the responsibility of the interface to enforce such policies.

msobczak:
Vadz - I'm not sure if the whole concept of failover is generic enough for anything reusable. Oracle and PostgreSQL do it in completely different ways and for SQLite it does not make any sense at all - just these three examples show that a reusable component might not see any reuse in practice.
It would also make the library structure and dependencies more complex.

PR #489, opened 2016-07-13
MSSQL ODBC // fixed #383 - long lines insert/select

### LINKED ISSUE #383: std::strings in nvarchar(max) is still a problem

snikulov:
Fixed failed tests for ODBC long lines usage.

### DIFF
--- a/include/soci/odbc/soci-odbc.h
+++ b/include/soci/odbc/soci-odbc.h
@@ -33,6 +33,10 @@
 #include <sqlext.h> // ODBC
 #include <string.h> // strcpy()
 
+#ifndef SQL_SS_LENGTH_UNLIMITED
+#define SQL_SS_LENGTH_UNLIMITED 0
+#endif
+
 namespace soci
 {
 
@@ -40,6 +44,10 @@ namespace details
 {
     // TODO: Do we want to make it a part of public interface? --mloskot
     std::size_t const odbc_max_buffer_length = 100 * 1024 * 1024;
+
+    // select max size from following MSDN article
+    // https://msdn.microsoft.com/en-us/library/ms130896.aspx
+    SQLLEN const ODBC_MAX_COL_SIZE = 8000;
 
     // This cast is only used to avoid compiler warnings when passing strings
     // to ODBC functions, the returned string may *not* be really modified.
--- a/src/backends/odbc/standard-use-type.cpp
+++ b/src/backends/odbc/standard-use-type.cpp
@@ -93,15 +93,6 @@ void* odbc_standard_use_type_backend::prepare_for_bind(
         memcpy(buf_, s.c_str(), size);
         buf_[size++] = '\0';
         indHolder_ = SQL_NTS;
-
-        // Strings of greater length are silently truncated at 8000 limit by MS
-        // SQL unless SQL_SS_LENGTH_UNLIMITED (which is defined as 0, but not
-        // available in all headers) is used.
-        if (size > 8000)
-        {
-          sqlType = SQL_LONGVARCHAR;
-          size = 0 /* SQL_SS_LENGTH_UNLIMITED */;
-        }
     }
     break;
     case x_stdtm:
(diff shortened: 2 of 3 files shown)


src/backends/odbc/standard-use-type.cpp
```diff
@@ -212,14 +211,20 @@ void odbc_standard_use_type_backend::pre_use(indicator const *ind)
     SQLSMALLINT sqlType(0);
     SQLSMALLINT cType(0);
     SQLLEN size(0);
+    SQLLEN bufLen(0);
 
     void* const sqlData = prepare_for_bind(size, sqlType, cType);
-
+    if (sqlType == SQL_LONGVARCHAR)
```
vadz: Even if we want to do it like this, it doesn't make much sense to do it like this: why explicitly set `sqlType` to `SQL_LONGVARCHAR` inside `prepare_for_bind()` only to reset it to `SQL_VARCHAR` here? Instead, the code in `prepare_for_bind()` added by c025cc86 should be simply reverted.

But I am not sure if doing this would be correct, after all it wasn't added without reason and did fix #287, so I'm afraid doing this might reintroduce that bug. It looks like there might be a difference in behaviour between ODBC 10 and 11 here, has this patch (and the "long string" ODBC unit test) been tested with both 10 and 11?

snikulov: Because with SQL_LONGVARCHAR it still have issue.
I've checked on both 10 on AppVeyor, and 11 - locally on Win 10, sql express 2016 - all works, fine as you can see.

Another option come into my mind - operates only with sizeof and move check for long ```size > 8000``` after ```prepare_for_bind```.

Anyway it solves the long standing issue - even if you don't like it.

snikulov:
@vadz Updated per your comments.
Now it looks better.

src/backends/odbc/standard-use-type.cpp
```diff
@@ -212,14 +203,19 @@ void odbc_standard_use_type_backend::pre_use(indicator const *ind)
...
     SQLSMALLINT cType(0);
     SQLLEN size(0);
+    SQLLEN bufLen(0);
 
     void* const sqlData = prepare_for_bind(size, sqlType, cType);
-
+    if (size > ODBC_MAX_COL_SIZE)
+    {
+        bufLen = size;
+        size = 0;
```
vadz: I think it would be more clear if the comment about this 0 actually being `SQL_SS_LENGTH_UNLIMITED` were kept.

vadz:
Of course it doesn't matter whether or not I "like" it, if it works for both versions, then this is the right thing to do. I just wasn't sure it was still working for v10 too, thanks for testing and verifying that it did!

FWIW I still don't fully understand what's going on here, e.g. [this MSDN article](https://msdn.microsoft.com/en-us/library/ms131462.aspx) only mentions that `ColumnSize` must be set to 0 for `varchar(max)`, but says nothing about `BufferLength`. But this wouldn't be the first time MSDN documentation wasn't 100% accurate and complete...

Anyhow, I don't have any other remarks about this, thank you for fixing this bug!

src/backends/odbc/standard-use-type.cpp
snikulov: I've updated to use ```SQL_SS_LENGTH_UNLIMITED``` and since it not defined on my MSVC2015/Win10, I also added this defined in our soci-odbc.h

snikulov:
@vadz @mloskot final changes with updates rebased into single change.
Could you please review? If no objections, I'll merge it to master.

include/soci/odbc/soci-odbc.h
```diff
@@ -33,6 +33,10 @@
 #include <sqlext.h> // ODBC
 #include <string.h> // strcpy()
 
+#ifndef SQL_SS_LENGTH_UNLIMITED
```
mloskot: Since this define is SQLNCLI driver-specific, it is not defined in `sql.h`, but in `sqlncli.h`, similarly to `SQL_SS_UDT` and others. Part of the ODBC mess.

In `sqlncli.h` it is defined as follows:
```
// SQL_SS_LENGTH_UNLIMITED is used to describe the max length of
// VARCHAR(max), VARBINARY(max), NVARCHAR(max), and XML columns
#define SQL_SS_LENGTH_UNLIMITED             0
```

The PR looks good to me. Thanks!

vadz:
Just to repeat, I don't have any further comments, thanks!

snikulov:
Ok. Merged on master.

PR #490, opened 2016-07-14
Updated failed appveyour.yml

snikulov:
Just try to fix mingw build on appveyour.

### DIFF
--- a/appveyor.yml
+++ b/appveyor.yml
@@ -2,18 +2,18 @@ version: 4.0.0.{build}
 
 environment:
     matrix:
-      - PRJ_GEN: "Visual Studio 11 2012 Win64"
-        BDIR: msvc2012
-        PRJ_CFG: Release
-        BOOST_ROOT: C:\Libraries\boost_1_58_0
-        POSTGRESQL_ROOT: C:\Program Files\PostgreSQL\9.4
-        MYSQL_DIR: C:\Program Files\MySql\MySQL Server 5.7
-      - PRJ_GEN: "Visual Studio 12 2013 Win64"
-        BDIR: msvc2013
-        PRJ_CFG: Release
-        BOOST_ROOT: C:\Libraries\boost_1_58_0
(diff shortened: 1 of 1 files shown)


mloskot:
@snikulov AppVeyor provides two installations of MinGW:
* MinGW 4.9.3 (32-bit) 
* MinGW-w64 5.3.0

See https://www.appveyor.com/docs/installed-software#mingw-msys-cygwin
So you shouldn't need to install it.

Here is AppVeyor configuration in action that uses both.
https://github.com/lexicalunit/nanodbc/blob/master/appveyor.yml

snikulov:
@mloskot Cool, thank you. Will check it soon. Anyway I should do something with this :)

mloskot:
@snikulov Don't miss this hack, required if CMake is used.

```
# For MinGW make to work correctly sh.exe must NOT be in your path.
- set PATH=%PATH:C:\Program Files\Git\usr\bin;=%
```

snikulov:
@mloskot I have a better option for that :)

snikulov:
@mloskot I fugured out that MinGW installed on appveyour is 32bit while MySQL/PostgreSQL is 64bit.
Not sure if 64bit client dlls will work with 32bit SOCI.
Will get back to installation from sf.net.

mloskot:
TBH I wouldn't bother about 32 bit and just do one build 64bit with
mingw-w64

snikulov:
Even mingw-w64 is 32bit.
Per link you've provided -

> MinGW-w64:
> MinGW root directory: C:\mingw-w64\i686-5.3.0-posix-dwarf-rt_v4-rev0

```i686``` in the name of MinGW-w64 clearly states - it 32bit, while it should be ```x86_64```.

mloskot:
I thought it is 32 bit binary but cross-compiler for 64-bit targets. I have now experience with MinGW, so rookie errors.

PR #495, opened 2016-07-15
Removed padding extra char to the end of each std string in odbc vectro-use-type

candychiu:
Removed padding extra char to the end of each std string.

### DIFF
--- a/src/backends/odbc/vector-use-type.cpp
+++ b/src/backends/odbc/vector-use-type.cpp
@@ -165,7 +165,7 @@ void odbc_vector_use_type_backend::prepare_for_bind(void *&data, SQLUINTEGER &si
             prepare_indicators(vecSize);
             for (std::size_t i = 0; i != vecSize; ++i)
             {
-                std::size_t sz = v[i].length() + 1;  // add one for null
+                std::size_t sz = v[i].length();
                 indHolderVec_[i] = static_cast<long>(sz);
                 maxSize = sz > maxSize ? sz : maxSize;
             }


snikulov:
@candychiu Could you please investigate failures in unit test produced with your change?

Here is [build](https://ci.appveyor.com/project/mloskot/soci/branch/master/job/4hosm41ndbtvqk94) from master branch
```
The following tests passed:
	soci_empty_test
	soci_empty_test_static
	soci_odbc_test_mssql
	soci_odbc_test_mssql_static
	soci_odbc_test_mysql
	soci_odbc_test_mysql_static
	soci_odbc_test_postgresql
	soci_odbc_test_postgresql_static
	soci_postgresql_test
	soci_postgresql_test_static
	soci_sqlite3_test
	soci_sqlite3_test_static
86% tests passed, 2 tests failed out of 14
Total Test time (real) =  83.41 sec
The following tests FAILED:
	  3 - soci_mysql_test (SEGFAULT)
	  4 - soci_mysql_test_static (SEGFAULT)
Errors while running CTest
```
Here the [build](https://ci.appveyor.com/project/mloskot/soci/build/4.0.0.58/job/w5ri5b2699y68na6) produced with your PR

```
The following tests passed:
	soci_empty_test
	soci_empty_test_static
	soci_odbc_test_mysql
	soci_odbc_test_mysql_static
	soci_odbc_test_postgresql
	soci_odbc_test_postgresql_static
	soci_postgresql_test
	soci_postgresql_test_static
	soci_sqlite3_test
	soci_sqlite3_test_static
71% tests passed, 4 tests failed out of 14
Total Test time (real) =  51.35 sec
The following tests FAILED:
	  3 - soci_mysql_test (SEGFAULT)
	  4 - soci_mysql_test_static (SEGFAULT)
	  7 - soci_odbc_test_mssql (Failed)
	  8 - soci_odbc_test_mssql_static (Failed)
Errors while running CTest
```

candychiu:
@snikulov @vadz 

When I bulk insert with the orig code, the following test, which is run in a sql GUI client, fails.

`select str, len(str) from soci`

Speficially, len(str) is 1 more than the length of the source string.  The problem goes away after I put in the code change.  However, the soci test now fails.  The reader returns a string with length equivalent to the length of the longest string in the bulk insert.  The original reader and writer are "compatible" even though the writer might not have written the right data into the database.  

I tried to compare the len(str) to outStr[i].length() before and after the edit:

`cout << '[' << strs[i] << "] " << lengths[i] << ' ' << strs[i].length() << endl;`
                       [actual string]               database length           std::string length
before               [hello]                                6                                     5
after                  [hello       ]                         5                                     12

It feels like SQL Server and std::string doesn't agree on how a string terminates?  I also tried setting `indHolderVec_[i] = SQL_NTS;` which didn't work.

vadz:
Just FYI, I'm working on this and will try to get this fixed before 4.0.0.

vadz:
I finally understand what happens here and current code is quite amazingly broken: first, it writes one byte too many when inserting string data into the table, so all strings inserted in this way end up with the length one greater than needed because the trailing NUL byte is actually stored in the database. However correcting this breaks reading data because the reading code relies (implicitly and almost certainly accidentally) on having this NUL byte in the data returned by the database which, of course, should have never been the case.

My only doubt at this moment is whether we can actually get the length of the string from ODBC or whether we need to trim all trailing spaces ourselves.

vadz:
Superseded by #561.

Issue #256, opened 2014-05-30
Avoid exception from destructor while stack unwinding


mloskot:
Maciej notified me about discussion on SOCI at Linkedin C++ Professionals Group, where folks discussed that SOCI still suffers from potential of throwing from destructor while another exception is handled.

Maciej suggested to fix it by checking ```std::uncaught_exception()``` in ```ref_counted_statement_base::dec_ref()```, and if it returns true, the ```final_action()``` should not be called to avoid ***any*** interactions with database.

PR #500, opened 2016-09-01
Throwing an exception out of a destructor is forbidden

nefelim:
I faced application abort because of thrown exception in destructor.

### DIFF
--- a/src/core/once-temp-type.cpp
+++ b/src/core/once-temp-type.cpp
@@ -37,14 +37,14 @@ once_temp_type & once_temp_type::operator=(once_temp_type const & o)
 
 once_temp_type::~once_temp_type() SOCI_NOEXCEPT_FALSE
 {
-    try
+    //try
     {
         rcst_->dec_ref();
     }
-    catch(...)
+    /*catch(...)
     {
         //Throwing an exception out of a destructor is forbidden
-    }
+    }*/
 }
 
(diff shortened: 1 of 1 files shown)


mloskot:
Sounds familiar https://github.com/SOCI/soci/pull/424#issuecomment-147404419
Please, have a look at the discussion to #256 and #257.

Please, provide a minimal example that shows how you call/use SOCI.

snikulov:
@mloskot @msobczak Looks like @nefelim have found the dark corner of SOCI design about throwing exception from destructor :)
https://github.com/SOCI/soci/issues/256


@nefelim Looks at test results. Why do you think they broken?

mloskot:
@snikulov Yes, so I suggested @nefelim to first familiarise himself with the archived discussion (which rooted with Roger Orr's message on the mailing list, https://sourceforge.net/p/soci/mailman/message/26451676/, also linked in that Issue comments).

Then, @nefelim can judge if he agrees or not, with SOCI design decisions. 
Perhaps, he found a new use case, apart from the convoluted one (https://github.com/SOCI/soci/issues/256#issuecomment-47965263) that we should consider.

Hence, I also asked @nefelim to provide better picture of what he is actually doing.

snikulov:
@mloskot I think this one should be closed.

nefelim:
I use SQLite3 db with "pragma foreign_key=ON" and has receive foreign key violation in case same is #256. Thank you for links.

PR #502, opened 2016-09-02
Add support c++11 on build with gcc

nefelim:
1. SOCI always builded with GCC compiler with c++03 flags
2. Аdd support c++11 for GCC to build SOCI

### DIFF
--- a/cmake/SociConfig.cmake
+++ b/cmake/SociConfig.cmake
@@ -29,6 +29,11 @@ endif(WIN32)
 
 set (SOCI_CXX_C11 OFF CACHE BOOL "Build to the C++11 standard")
 
+if(CMAKE_COMPILER_IS_GNUCC OR CMAKE_COMPILER_IS_GNUCXX)
+    if(NOT CMAKE_CXX_COMPILER_VERSION LESS 4.7)
+      set(SOCI_CXX_C11 ON)
+    endif()
+endif()
 
 #
 # Force compilation flags and set desired warnings level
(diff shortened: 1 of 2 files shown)


vadz:
I don't think it's right to use C++11 by default with g++ 4.8+. People should be able to specify the version of the language they want explicitly, e.g. by doing `CXXFLAGS=-std=c++14` or whatever, and the default should be whichever version the compiler uses without any options, which will remain C++03 for g++ 4.x.

nefelim:
1. >> I don't think it's right to use C++11 by default with g++ 4.8+
Please read https://gcc.gnu.org/gcc-4.7/cxx0x_status.html
2. With default options throwing an exception out of a destructor is close application. I has enable nothrow(false) option for destructor.

snikulov:
@nefelim Have you tried to enable CXX11 with ```-DSOCI_CXX_C11=ON``` ?

nefelim:
@snikulov cmake script has been enable this define automatically in lines:
`+if(SOCI_CXX_C11)
+  add_definitions(-DSOCI_CXX_C11)
+endif()`

Support c++11 enabled in gcc 4.7 and more
`+if(CMAKE_COMPILER_IS_GNUCC OR CMAKE_COMPILER_IS_GNUCXX)
+    if(NOT CMAKE_CXX_COMPILER_VERSION LESS 4.7)
+      set(SOCI_CXX_C11 ON)
+    endif()
+endif()` 

I have check changes in my machine. Now my application not closed after throwing exception in destructor (its solution for me).
Unittests in build server completed too.
My SOCI configuration: http://pastebin.com/2pFxYGt8

snikulov:
@nefelim So, can we close this PR?

nefelim:
> @nefelim So, can we close this PR?
@snikulov I can't close it. It is not my responsibility. In my project I use fork from SOCI.

mloskot:
@nefelim You should be able to close it yourself.

nefelim:
@mloskot 10x :)

PR #506, opened 2016-09-27
fix iOS compile error

dorgonman:
soci/src/backends/sqlite3/common.cpp:58:36: error:
      implicit conversion loses integer precision: 'long' to 'int'
      [-Werror,-Wshorten-64-to-32]
    details::mktime_from_ymdhms(t, year, month, day, hour, minute, second);

just a hotfix for compile error

### DIFF
--- a/src/backends/sqlite3/common.cpp
+++ b/src/backends/sqlite3/common.cpp
@@ -55,5 +55,5 @@ void soci::details::sqlite3::parse_std_tm(char const *buf, std::tm &t)
         second = parse10(p1, p2, errMsg);
     }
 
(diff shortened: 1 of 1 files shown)


vadz:
I think it would be better to modify `parse10()` to return an int, I see no reason for it to return long as none of the date components needs more than int range. In fact, I'd change this function to check for both `v <= INT_MAX` and `v >= 0` because negative numbers shouldn't be allowed here neither.

dorgonman:
yes, I will be better to change parse10() to return int, 
but this is just a hotfix for compile error, I don't want to change too much code.

PR #509, opened 2016-10-10
Added bigstring (XML and CLOB) support.

msobczak:
This PR adds support for big strings and XML values for Oracle and PostgreSQL.

The problem with large strings is that in Oracle they have to be bound differently from regular std::string values and the target database type has to be different as well (CLOB instead of varchar).
Interestingly, large string values sometimes happen to be XML values, which have their own dedicated target types as well, in both Oracle and PostgreSQL.
In order to accommodate these differences in a more-or-less portable way, two wrapper types were added and the rest of the code was extended appropriately.
See the added test cases and the basic docs additional to see the general idea behind these features.

### DIFF
--- a/include/soci/column-info.h
+++ b/include/soci/column-info.h
@@ -63,7 +63,8 @@ struct type_conversion<column_info>
         ci.scale = get_numeric_value(v, "NUMERIC_SCALE");
 
         const std::string & type_name = v.get<std::string>("DATA_TYPE");
-        if (type_name == "text" ||
+        if (type_name == "text" || type_name == "TEXT" ||
+            type_name == "clob" || type_name == "CLOB" ||
             type_name.find("char") != std::string::npos ||
             type_name.find("CHAR") != std::string::npos)
         {
@@ -96,10 +97,15 @@ struct type_conversion<column_info>
         }
         else if (type_name.find("blob") != std::string::npos ||
             type_name.find("BLOB") != std::string::npos ||
-            type_name.find("bytea") != std::string::npos ||
-            type_name.find("BYTEA") != std::string::npos)
+            type_name.find("oid") != std::string::npos ||
+            type_name.find("OID") != std::string::npos)
         {
             ci.type = dt_blob;
+        }
+        else if (type_name.find("xml") != std::string::npos ||
+            type_name.find("XML") != std::string::npos)
+        {
+            ci.type = dt_xml;
         }
         else
         {
--- a/src/backends/postgresql/standard-use-type.cpp
+++ b/src/backends/postgresql/standard-use-type.cpp
@@ -9,6 +9,7 @@
 #include "soci/postgresql/soci-postgresql.h"
 #include "soci/blob.h"
 #include "soci/rowid.h"
+#include "soci/type-wrappers.h"
 #include "soci/soci-platform.h"
 #include "soci-dtocstr.h"
 #include "soci-exchange-cast.h"
@@ -159,7 +160,25 @@ void postgresql_standard_use_type_backend::pre_use(indicator const * ind)
                 snprintf(buf_, bufSize, "%lu", bbe->oid_);
             }
             break;
-
+        case x_xmltype:
+            {
+                xml_type * xml = static_cast<xml_type *>(data_);
+                std::string * s = &(xml->value);
+                
+                buf_ = new char[s->size() + 1];
+                std::strcpy(buf_, s->c_str());
+            }
+            break;
+        case x_longstring:
+            {
+                long_string * ls = static_cast<long_string *>(data_);
+                std::string * s = &(ls->value);
+                
+                buf_ = new char[s->size() + 1];
+                std::strcpy(buf_, s->c_str());
+            }
+            break;
+            
         default:
             throw soci_error("Use element used with non-supported type.");
         }
--- a/src/backends/sqlite3/statement.cpp
+++ b/src/backends/sqlite3/statement.cpp
@@ -182,6 +182,9 @@ sqlite3_statement_backend::load_rowset(int totalRows)
                             col.buffer_.data_ = (col.buffer_.size_ > 0 ? new char[col.buffer_.size_] : NULL);
                             memcpy(col.buffer_.data_, sqlite3_column_blob(stmt_, c), col.buffer_.size_);
                             break;
+
+                        case dt_xml:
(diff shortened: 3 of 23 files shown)


snikulov:
@msobczak Could you please fix Travis CI build? SQLite3 backend now fails with ```-Werror``` option.

mloskot:
I skimmed the set of changes and I have no comments.
Great addition.

mloskot:
AFAIR, MySQL tests have been segfault'ing for long time, so not really relevant to fix them for this PR. 
I think, it only matters that CI builds confirm this PR does not break anything new.

snikulov:
@mloskot Yes. MySQL known issue #136

src/backends/postgresql/standard-use-type.cpp
```diff
@@ -159,7 +160,25 @@ void postgresql_standard_use_type_backend::pre_use(indicator const * ind)
...
+                std::strcpy(buf_, s->c_str());
+            }
+            break;
+        case x_longstring:
+            {
+                long_string * ls = static_cast<long_string *>(data_);
+                std::string * s = &(ls->value);
+                
+                buf_ = new char[s->size() + 1];
+                std::strcpy(buf_, s->c_str());
```
vadz: Are we sure that there can be no embedded `NUL` bytes in long strings? We probably are, but OTOH using `memcpy()` instead of `strcpy()` would work in any case and wouldn't really be any more difficult in this case...

src/backends/sqlite3/statement.cpp
```diff
@@ -182,6 +182,9 @@ sqlite3_statement_backend::load_rowset(int totalRows)
                             col.buffer_.data_ = (col.buffer_.size_ > 0 ? new char[col.buffer_.size_] : NULL);
                             memcpy(col.buffer_.data_, sqlite3_column_blob(stmt_, c), col.buffer_.size_);
                             break;
+
+                        case dt_xml:
+                            throw soci_error("XML data type is not supported");
```
vadz: Wouldn't it make more sense to gracefully handle it as strings? Otherwise it seems it would be difficult to write code using `dt_xml` and working with several databases.

tests/sqlite3/Makefile.basic
```diff
@@ -0,0 +1,12 @@
+COMPILER = g++
```
vadz: Is it really a good idea to maintain and add these hand-written makefiles? I'm not a big fan of CMake myself, but IMO it's really better to have one, single build system rather than two in parallel...

include/soci/column-info.h
```diff
@@ -96,11 +97,16 @@ struct type_conversion<column_info>
         }
         else if (type_name.find("blob") != std::string::npos ||
             type_name.find("BLOB") != std::string::npos ||
-            type_name.find("bytea") != std::string::npos ||
-            type_name.find("BYTEA") != std::string::npos)
+            type_name.find("oid") != std::string::npos ||
```
vadz: I'm not sure why did `bytea` have to be changed to `oid`, is it related to the XML/long string changes? If so, I'd be curious to understand how.

Independently of my curiosity, I also wonder if this can/will have any compatibility implications?

src/backends/postgresql/standard-use-type.cpp
msobczak: This is a non-issue for several reasons.
1. Arbitrary binary data, with or without embedded zeros, can be stored as BLOBs. There is no reason to misuse strings for the same use case.
2. There is nothing special in long strings with regard to embedded zeros. The same question can be discussed in the context of short strings, too - and the last decade is an evidence that such a feature is not missing.
3. PostgreSQL client API, which is used below this layer, is string based and I'm not even sure if it would work with embedded zeros. Please investigate.

src/backends/sqlite3/statement.cpp
msobczak: Please provide a comprehensive handling of these new types for SQLite3 in a separate PR. The above lines are just placeholders that make the switch coverage complete for the enumeration type that is not SQLite3-specific.

tests/sqlite3/Makefile.basic
msobczak: These Makefiles are useful on test systems that have minimal set of packages installed.
As to having one, single build system - these Makefiles played this role from the beginning of the project, even before CMake was introduced.

msobczak:
General note (code comments were answered in separate replies):

> My main concern is that it seems to (please correct me if I'm wrong) not supported at all in the other backends

This is not a problem for the users of those other backends, as they can pretend that the new feature does not exist. It does not disrupt the existing application code.

> IMO this will make life difficult for people like me

There is no risk of making anybody's life difficult as the existing applications are not affected.
The new feature is of immediate benefit to the users of Oracle and PostgreSQL; it is also possible to propose separate PRs with implementations for other backends.

vadz:
Just to be sure, I do realize that this change is backwards compatible (BLOB changes, which I still don't really understand, aside) and doesn't break any existing code as it can just continue not using the new types. All I'm saying is that it seems it would take exactly as much (or maybe even less as you wouldn't have to write `throw` statements) effort to fall back to normal string handling for the other backends rather than give an error and that this would make the new feature even more useful.

snikulov:
@vadz 
> Just to be sure, I do realize that this change is backward compatible
It should not be stopping point for 4.x, IMHO.

msobczak:
> All I'm saying is that it seems it would take exactly as much

In the case of Oracle and PostgreSQL, 5 or 6 files had to be changed (+ tests + docs). There is no reason to claim that it will take significantly less for SQLite3. Please feel free to propose a comprehensive SQLite3 implementation in a separate PR.

vadz:
@msobczak I have the impression that we're speaking completely past each other :-( I'm absolutely _not_ saying that implementing it for SQLite3 or whatever else would be trivial. All I say is that if XML string type were handled in exactly the same way as normal string type which does seem to be trivial (isn't it?), this would allow people like me, who use other backends than Oracle and PostgreSQL, to start using it right now because it would be a gain when using PostgreSQL and not have any drawbacks for everything else, where I do use plain strings for storing XML currently.

It seems so obvious to me that falling back on plain strings is better than throwing an exception in the other backends that either I must be explaining it extremely unclearly or I must be missing something here. If I did manage to explain it clearly enough this time, I'd still be very curious to know why do you think the current approach is better.

PR #517, opened 2016-12-01
Add JSON support to the MySQL backend

ayllon:
This is a small change, based on what the PostgreSQL code looks like. There is no #define for the type, though, and for old library versions the enum is not available either, so I have put the constant instead, as with MYSQL_TYPE_NEWDECIMAL


mloskot:
Looks good to me

PR #521, opened 2017-01-14
Compatibility with C++11 on release branch 3.2

denisarnaud:
With C++11, SOCI v3.2.3 no longer compiles. Hence, on the newest releases of Linux distributions (eg, those shipping with g++ 6), SOCI can no longer be delivered. That patch allows to deliver SOCI v3.2.x (v3.2.4?) with G++ 6.
See for instance the [build of SOCI-3.2.3-9 on Fedora Rawhide](https://koji.fedoraproject.org/koji/buildinfo?buildID=833787).

### DIFF
--- a/src/CMakeLists.txt
+++ b/src/CMakeLists.txt
@@ -39,6 +39,10 @@ soci_version(MAJOR 3 MINOR 2 PATCH 3)
 include(SociSystemInfo)
 include(SociConfig)
 
+if(CMAKE_COMPILER_IS_GNUCXX)
+  add_definitions(-std=gnu++0x)
+endif()
+
 boost_report_value(SOCI_PLATFORM_NAME)
 boost_report_value(SOCI_COMPILER_NAME)
 
--- a/src/core/rowset.h
+++ b/src/core/rowset.h
@@ -154,8 +154,8 @@ private:
 
     unsigned int refs_;
 
-    const std::auto_ptr<statement> st_;
-    const std::auto_ptr<T> define_;
+    const std::unique_ptr<statement> st_;
+    const std::unique_ptr<T> define_;
 
     // Non-copyable
     rowset_impl(rowset_impl const &);
--- a/src/core/session.cpp
+++ b/src/core/session.cpp
@@ -235,11 +235,11 @@ std::string session::get_query() const
 }
 
 void session::set_query_transformation_(
-        std::auto_ptr<details::query_transformation_function> qtf)
+        std::unique_ptr<details::query_transformation_function> qtf)
 {
     if (isFromPool_)
     {
-        pool_->at(poolPosition_).set_query_transformation_(qtf);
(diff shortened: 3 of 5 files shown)


jsonn:
This code breaks using SOCI in C++98 environments and is therefore not acceptable. I don't understand the purpose of the patch at all. Using std::auto_ptr might be deprecated, but it is certainly valid C++11.

vadz:
I agree with @jsonn, we don't want to _require_ C++11 in SOCI, even if we do want it to compile without warnings with C++11 compilers (and, soon, compile at all with C++17 ones which really don't have `std::auto_ptr` any more). So we need to use conditional compilation to select the appropriate class to use. And if anybody wants to reuse/adapt a [header](https://github.com/vslavik/xmlwrapp/blob/master/src/libxml/cpp11.h) addressing similar problem in the xmlwrapp library, please feel free to do it.

denisarnaud:
As stated above, the idea is to allow SOCI to compile again on recent Linux distributions such as Fedora 24+, which use g++ v6.
The newest code of SOCI (master branch) compiles correctly, and the ideal would therefore be to release newer versions of SOCI, for instance v3.2.4, or even 4.0.1. That point was raised last September (2016), as seen on the mailing list archive: http://sourceforge.net/p/soci/mailman/message/35342624/

However, as no new release has been made, and the version v3.2.3 of SOCI still fails to build with g++6, a patch must be made, for instance a backport of the the ec7d2df commit. But that latter was performed on [a too far away branch of SOCI](https://github.com/SOCI/soci/tree/master) for it to be applicable on the [3.2 branch](http://github.com/SOCI/soci/tree/release/3.2).

So, that pull request (#521) is a minimalist attempt to solve the g++6/C++11 compatibility issue, only on the [3.2 branch](http://github.com/SOCI/soci/tree/release/3.2), following some [standard recommendations on StackOverflow](http://stackoverflow.com/questions/3451099/stdauto-ptr-to-stdunique-ptr). It certainly breaks MS Visual builds. So, if someone could add the corresponding MSVC safeguards, it would be perfect. The same for C++98. Or, again, optimally, ec7d2df should be backported on the [3.2 branch](http://github.com/SOCI/soci/tree/release/3.2).

denisarnaud:
I agree with @vadz: if you come up with a satisfying patch for the 3.2 branch, please do not hesitate.

snikulov:
@denisarnaud @mloskot I want new release! :-) Brand new v4.0.0!

mloskot:
@snikulov Do you mean to release the master as v4.0.0?

snikulov:
@mloskot yes. I think it is time to move on :-)

mloskot:
I agree. I'm leaving for the weekend. Let's wait for some more feedback and cut the release during the upcoming 1-2 weeks.

vadz:
AFAICS this can't/shouldn't be applied, so closing it. We will release 4.0.0 soon and, of course, if you have any problems building it with any recent compilers, please report them. TIA!

PR #524, opened 2017-03-01
Firebird blob bug

ravselj:
I have discovered bug in Firebird BLOB backend save method. It is currently not possible to write data larger than max `unsigned short` 0xFFFF. The problem is that Firebird API accepts only `unsigned short` as data size and expects that `isc_put_segment` is called in loop until all data has been written meaning only 0xFFFF can be written at once.

PR consists of 2 commits. 

First one add additional Firebird test to indicate the problem. I tried to write byte sequence of size 0x10000 (out of unsigned short range). Cast to `unsigned short` results in 0 and therefore no data is written to blob.

Second commit resolves the issue.

### DIFF
--- a/src/backends/firebird/blob.cpp
+++ b/src/backends/firebird/blob.cpp
@@ -247,13 +247,26 @@ void firebird_blob_backend::save()
     if (data_.size() > 0)
     {
         // write data
-        if (isc_put_segment(stat, &bhp_,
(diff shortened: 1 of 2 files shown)


mloskot:
I'm not experienced with Firebird, so I can't comment on the details of the fix.

The patch only changes Firebird backend, includes comments and test.
The CI builds are green (AppVeyor failed partially, for MySQL only).
I'm happy to merge.

Thank you!

PR #535, opened 2017-04-17
Fixing uninitialized gotData_

csteifel:
Set gotData_ to false on new procedures and make sure it is copied
correctly for copied procedures.


vadz:
Thanks for fixing this!

Issue #473, opened 2016-05-16
Handle OCI_SUCCESS_WITH_INFO


kyudin:
Greetings. Why soci oracle backend does not handle OCI_SUCCESS_WITH_INFO result?
Just throw it as real error, but not as "Unknown error code" will help much.

```
    case OCI_ERROR:
    case OCI_SUCCESS_WITH_INFO:
        OCIErrorGet(errhp, 1, 0, &errcode,
             errbuf, sizeof(errbuf), OCI_HTYPE_ERROR);
        msg = reinterpret_cast<char*>(errbuf);
        errNum = static_cast<int>(errcode);
        break;
```

Actually probably it will be better to not treat OCI_SUCCESS_WITH_INFO as an exception, since actually it is not error, but just warning.
Thanks.

vadz:
I see `OCI_SUCCESS_WITH_INFO` already being treated specially in src/backends/oracle/statement.cpp, so I'm not sure where exactly can we get "Unknown error" for it. I don't know much about OCI anyhow and so probably won't be able to do anything about it anyhow, but if someone is going to look at this, it would be certainly helpful to have a simple way of reproducing the problem, could you please provide one? Perhaps even as a (currently failing) new test in the test suite?

kyudin:
vadz, actually `OCI_SUCCESS_WITH_INFO` can happen in any OCI call. For example yesterday we have this problem in production system, when I correct code for check what error actually was, we see, that info was: password will be expire in 10 days, that was on session open.

snikulov:
@kyudin could you please provide information, which version/branch you using?

kyudin:
@snikulov We use release 3.2.

snikulov:
@vadz I see what @kyudin mean.
Every condition check ``res != OCI_SUCCESS`` will produce ``throw soci_error(...)``
Which, perhaps, not good in case when res equals OCI_SUCCESS_WITH_INFO...

vadz:
[...]

PR #536, opened 2017-05-05
Pull request for issues 511 and 473

ddowling:
Pull request to address issues:
https://github.com/SOCI/soci/issues/511
https://github.com/SOCI/soci/issues/473

To correctly handle the OCI_SUCCESS_WITH_INFO return code we are likely going to need to ability to handle cases where the backend does not completely fail an operation with an assert but wants to log some warning or information message back to the user of the API. This happens on Oracle database connections when the users password is about to expire. With this patch the call will still fail but rather than getting a "soci error: Unknown error code" exception the caller will get an exception with a message saying the password is about to expire.

### DIFF
--- a/src/backends/oracle/error.cpp
+++ b/src/backends/oracle/error.cpp
@@ -58,6 +58,7 @@ void soci::details::oracle::get_error_details(sword res, OCIError *errhp,
         msg = "soci error: No data";
         break;
     case OCI_ERROR:
+    case OCI_SUCCESS_WITH_INFO:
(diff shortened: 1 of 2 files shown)


mloskot:
Thanks!

FYI, AppVeyor failures of MySQL only tests are known issue. Let's ignore those and consider CI builds green.

vadz:
@mloskot I think it would be better to disable AppVeyor MySQL tests then for now, wouldn't it?

vadz:
Thanks for the the PR!

PR #541, opened 2017-06-26
disable autosetting of SOCI_CXX_C11 to OFF

ckaminski:
When building SOCI as a sub-project, this file overrides the flag value set by the parent project. 

My parent project was built as C++11, but this default was causing SOCI to build as C++98.

### DIFF
--- a/cmake/SociConfig.cmake
+++ b/cmake/SociConfig.cmake
@@ -27,7 +27,9 @@ endif(WIN32)
 # C++11 Option
 #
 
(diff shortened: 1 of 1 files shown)


vadz:
I have no idea if this is the right way to do it in CMakese, could any resident CMake experts please chime in?

The idea of inheriting C++11 setting from the parent project does look correct to me, of course.

mloskot:
LGTM

mloskot:
Thanks

PR #544, opened 2017-07-06
moved to latest Catch UT Framework

snikulov:
- Updated to latest Catch v1.9.6 and fixed some obstacles 
- Added -Wextra options to compilation

### DIFF
--- a/cmake/SociConfig.cmake
+++ b/cmake/SociConfig.cmake
@@ -49,7 +49,7 @@ if (MSVC)
 else()
 
   set(SOCI_GCC_CLANG_COMMON_FLAGS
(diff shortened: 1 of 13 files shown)


mloskot:
@snikulov Good idea, thanks!

Looks like `-Wextra` combined with `-Werror` are causing failures on Travis CI, aren't they?

snikulov:
@mloskot yeah... issue with struct initialization on g++ prior v4.7 (ref https://gcc.gnu.org/bugzilla/show_bug.cgi?id=53119)

snikulov:
@mloskot now it, at least, as was before...
But what can be if C++ intermixed with C structs without constructors :)

PR #547, opened 2017-07-12
* added common parse() for connection parameters

snikulov:
Proposal:
- re-use same function to parse a connection string into parameters map.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -4,7 +4,7 @@
 # Copyright (c) 2013 Mateusz Loskot <mateusz@loskot.net>
 #
 language: cpp
-
+sudo: required
 compiler:
     - g++
     - clang
--- a/include/soci/connection-parameters.h
+++ b/include/soci/connection-parameters.h
@@ -32,26 +32,33 @@ public:
 
     // Retrieve the backend and the connection strings specified in the ctor.
     backend_factory const * get_factory() const { return factory_; }
-    void set_connect_string(const std::string & connectString) { connectString_ = connectString; }
+    void set_connect_string(std::string const & connectString)
+    {
+        connectString_ = connectString;
+        parse();
+    }
     std::string const & get_connect_string() const { return connectString_; }
 
     // Set the value of the given option, overwriting any previous value.
-    void set_option(const char * name, std::string const & value)
+    void set_option(std::string const & name, std::string const & value)
     {
         options_[name] = value;
     }
 
+    // common method to parse param=value from connection string
+    void parse();
+
     // Return true if the option with the given name was found and fill the
     // provided parameter with its value.
-    bool get_option(const char * name, std::string & value) const
+    bool get_option(std::string const & key, std::string & value) const
     {
-        Options::const_iterator const it = options_.find(name);
-        if (it == options_.end())
-            return false;
-
-        value = it->second;
-
-        return true;
+        Options::const_iterator const it = options_.find(key);
+        if (it != options_.end())
+        {
+            value = it->second;
+            return true;
+        }
+        return false;
     }
 
 private:
--- a/src/backends/mysql/session.cpp
+++ b/src/backends/mysql/session.cpp
@@ -30,107 +30,6 @@ using std::string;
 namespace
 { // anonymous
 
-void skip_white(std::string::const_iterator *i,
-    std::string::const_iterator const & end, bool endok)
-{
-    for (;;)
-    {
-        if (*i == end)
-        {
-            if (endok)
-            {
-                return;
-            }
-            else
-            {
-                throw soci_error("Unexpected end of connection string.");
-            }
-        }
-        if (std::isspace(**i))
-        {
-            ++*i;
-        }
-        else
-        {
-            return;
-        }
-    }
-}
-
-std::string param_name(std::string::const_iterator *i,
-    std::string::const_iterator const & end)
-{
-    std::string val("");
-    for (;;)
-    {
-        if (*i == end or (not std::isalpha(**i) and **i != '_'))
-        {
-            break;
-        }
-        val += **i;
-        ++*i;
-    }
-    return val;
-}
-
-string param_value(string::const_iterator *i,
-    string::const_iterator const & end)
-{
-    string err = "Malformed connection string.";
-    bool quot;
-    if (**i == '\'')
-    {
-        quot = true;
-        ++*i;
-    }
-    else
-    {
-        quot = false;
-    }
-    string val("");
-    for (;;)
-    {
-        if (*i == end)
-        {
-            if (quot)
-            {
-                throw soci_error(err);
-            }
-            else
-            {
-                break;
-            }
-        }
-        if (**i == '\'')
-        {
-            if (quot)
-            {
-                ++*i;
-                break;
-            }
-            else
-            {
-                throw soci_error(err);
-            }
-        }
-        if (not quot and std::isspace(**i))
-        {
-            break;
-        }
-        if (**i == '\\')
-        {
-            ++*i;
-            if (*i == end)
-            {
-                throw soci_error(err);
-            }
-        }
-        val += **i;
-        ++*i;
-    }
-    return val;
-}
-
 bool valid_int(const string & s)
 {
     char *tail;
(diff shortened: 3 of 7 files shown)


tests/common-tests.h
```diff
@@ -4171,6 +4171,7 @@ TEST_CASE_METHOD(common_tests, "Blank padding", "[core][insert][exception]")
     CHECK(tvarchar == test1);
 }
 
+
```
vadz: Probably not an intentional change?

tests/mysql/test-mysql.cpp
```diff
@@ -24,6 +24,16 @@
 std::string connectString;
 backend_factory const &backEnd = *soci::factory_mysql();
 
+TEST_CASE("MySQL connection parameters", "[mysql][connection_parameters]")
+{
+    soci::connection_parameters myparam("mysql", "dbname=test user=root password='test123'");
```
vadz: It would be nice to test quoted values with spaces in them as this uses a different part of the code.

src/core/connection-parameters.cpp
```diff
@@ -59,4 +61,38 @@ connection_parameters::connection_parameters(std::string const & fullConnectStri
...
+        std::getline(ssconn, key, '=');
+        std::getline(ssconn, val, ' ');
+
+        char delim = 0;
+        if (val.size()>0 && (val[0]=='\"' || val[0]=='\''))
+        {
+            delim = val[0];
+            std::string quotedVal = val.erase(0, 1);
+
+            if (quotedVal[quotedVal.size()-1] ==  delim)
```
vadz: You need to check that `quotedVal` is not empty here. Perhaps by testing for `val.size() >= 2` above, or by checking for `!quotedVal.empty()` explicitly here.

Also, personally, I think `*s.rbegin()` is more clear and less error-prone than `s[s.size() - 1]`.

src/core/connection-parameters.cpp
```diff
@@ -59,4 +61,38 @@ connection_parameters::connection_parameters(std::string const & fullConnectStri
...
+void connection_parameters::parse()
+{
+    std::stringstream ssconn(connectString_);
+    while (!ssconn.eof() && ssconn.str().find('=') != std::string::npos)
+    {
+        std::string key, val;
+        std::getline(ssconn, key, '=');
+        std::getline(ssconn, val, ' ');
+
+        char delim = 0;
```
vadz: This could be declared inside the `if` below, reducing its scope and making it possible to initialize it at the same time.

src/core/connection-parameters.cpp
```diff
@@ -59,4 +61,38 @@ connection_parameters::connection_parameters(std::string const & fullConnectStri
...
+            delim = val[0];
+            std::string quotedVal = val.erase(0, 1);
+
+            if (quotedVal[quotedVal.size()-1] ==  delim)
+            {
+                quotedVal.erase(val.size()-1);
+            }
+            else // space inside value string
+            {
+                std::getline(ssconn, val, delim);
```
vadz: This doesn't take into account escaped delimiters which, I think, are possible, e.g. `password="in \"quotes\""`.

.travis.yml
```diff
@@ -4,7 +4,7 @@
 # Copyright (c) 2013 Mateusz Loskot <mateusz@loskot.net>
 #
 language: cpp
-
+sudo: required
```
snikulov: It was side fix targeted issue discovered here https://travis-ci.org/snikulov/soci/jobs/252953339

include/soci/connection-parameters.h
```diff
@@ -32,26 +32,33 @@ class SOCI_DECL connection_parameters
...
+    void set_connect_string(std::string const & connectString)
+    {
+        connectString_ = connectString;
+        parse();
+    }
     std::string const & get_connect_string() const { return connectString_; }
 
     // Set the value of the given option, overwriting any previous value.
-    void set_option(const char * name, std::string const & value)
+    void set_option(std::string const & name, std::string const & value)
```
snikulov: @vadz it was intentional change to replace the naked pointer to type T param as used in C++ std::map to avoid conversion from const char* -> temporary T in our case (std::string).

src/backends/mysql/session.cpp
```diff
@@ -148,155 +47,62 @@ bool valid_int(const string & s)
...
-        if (par == "port" and not *port_p)
-        {
-            if (not valid_int(val))
-            {
-                throw soci_error(err);
-            }
-            *port = std::atoi(val.c_str());
-            if (*port < 0)
+            port = std::atoi(portstr.c_str());
+            if (port < 0)
```
snikulov: @vadz previous behavior was not changed see here https://github.com/SOCI/soci/blob/master/src/backends/mysql/session.cpp#L201
port by default initialized to 0 by me and it is the correct value for the default connection - ref to mysql_real_connect API docs here https://dev.mysql.com/doc/refman/5.7/en/mysql-real-connect.html

tests/common-tests.h
snikulov: @vadz No. At the first step, I've added test case here.
But, unfortunately, connection_parameters coupled with the backend, which fails common part tests other than MySQL. So I moved it to the specific test file.

src/backends/mysql/session.cpp
```diff
@@ -148,155 +47,62 @@ bool valid_int(const string & s)
...
-        {
-            *ssl_cert = val;
-            *ssl_cert_p = true;
-        }
-        else if (par == "sslkey" and not *ssl_key_p)
+        else
         {
-            *ssl_key = val;
-            *ssl_key_p = true;
+            throw soci_error("Wrong port value");
```
snikulov: Yep. Will fix it today night.

tests/mysql/test-mysql.cpp
snikulov: @vadz Feel free to add more test cases :) Current test suite does not complain.

src/core/connection-parameters.cpp
snikulov: @vadz copy-paste code from one of the parser/splitter (sqlite3 session to be precise). Not my invention.
So if you think it should be changed - feel free to do so.

src/core/connection-parameters.cpp
snikulov: see above comment. Now it in a single place for update.

.travis.yml
mloskot: @vadz I think that the missing `sudo` requirement was an overlook. All of Travis CI shell scripts require `sudo` and I'm actually surprised `install_*.sh` scripts worked without `sudo: required`.
Am I missing something?

include/soci/connection-parameters.h
mloskot: I'm not sure I understand it. Regardless if `value` below is `char*` or `string const&`, copy is made:
```
map<string, string> m;
m[key] = value;
```

snikulov:
@mloskot let me disagree http://en.cppreference.com/w/cpp/container/map/operator_at  T operator[](const T&) should work in this case.

snikulov:
@mloskot I'm talking about key here, not return value.

mloskot:
> @mloskot I'm talking about key here, not return value.

@snikulov Sorry for confusion, I should have said in https://github.com/SOCI/soci/pull/547#discussion_r127269132 that I mean 

Regardless if value below is `char*` or `string const&`, **copy of value is made**

include/soci/connection-parameters.h
mloskot: To be more precise, here is my understanding of the case like this
```
    string key{"k"}, value{"v"};
    map<string, string> m;
    m[key] = value;
```

AFAIU, key passed by `string const&` is still eventually emplaced in `m` by `std::forward<string>(key)` which in fact makes a copy. The `value` is copied too, obviously.

In this particular case, passing key by `char*` or `string&` makes no difference to me, both are copied. Please, correct me if I'm wrong.

vadz:
Sorry if it was unclear, but my main point was that even if these changes are necessary, it would be better to make them as separate commits, e.g. one commit for adding `sudo: required`, another commit changing `const char*` to `std::string` (although I am still not sure why is it needed) etc.

On a related note, it would also be great if we could agree to use the standard format for the commit messages, i.e., in brief, write them in imperative mood with 50 char summary and body wrapped to 72 columns and explaining what was done in the commit and why (see e.g. https://chris.beams.io/posts/git-commit/ for more details).

src/backends/mysql/session.cpp
vadz: Sorry, I didn't notice this didn't change. It's still wrong though, `atoi()` returns 0 if you pass it `"foobar"`, so we should test for it ideally.

vadz:
And on the topic of "having a single version of the broken code instead of multiple ones": I don't think this is quite correct, I believe you copied the wrong version (sqlite3 one, apparently) instead of a better one (mysql one). I.e. it looks like mysql wouldn't crash on `foo="` parameter before, but would now. I don't care about MySQL one way or the other, but this does seem like regression to me (admittedly, SQLite is unchanged because it was already broken).

mloskot:
@vadz 
>  it would be better to make them as separate commits

I agree.

> it would also be great if we could agree to use the standard format for the commit messages, i.e., in brief, write them in imperative mood with 50 char summary and body wrapped to 72 columns and explaining what was done in the commit and why

I agree. I try to follow this guidelines most of the time, especially in ad-hoc PRs I submit to variety projects - good commit message can help defending the change :-)

snikulov:
@mloskot @vadz ok. thank you for a guideline.

mloskot:
BTW, pretty much de-facto gold standard for commit messages is http://tbaggery.com/2008/04/19/a-note-about-git-commit-messages.html

PR #548, opened 2017-07-18
Consistently check for connection in the session object methods

vadz:
This was done for most, but not all, of them.

Add missing checks to ensure consistent and safer (exception instead of
a null pointer dereference) behaviour in all cases.

---

Just some cleanup before submitting more changes to the same code.

### DIFF
--- a/src/core/session.cpp
+++ b/src/core/session.cpp
@@ -361,16 +361,22 @@ bool session::get_last_insert_id(std::string const & sequence, long & value)
 
 details::once_temp_type session::get_table_names()
 {
+    ensureConnected(backEnd_);
+
     return once << backEnd_->get_table_names_query();
 }
 
 details::prepare_temp_type session::prepare_table_names()
 {
+    ensureConnected(backEnd_);
+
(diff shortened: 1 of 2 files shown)


src/core/session.cpp
```diff
@@ -361,16 +361,22 @@ bool session::get_last_insert_id(std::string const & sequence, long & value)
 
 details::once_temp_type session::get_table_names()
 {
+    ensureConnected(backEnd_);
```
mloskot: Perhaps good time to rename to `ensure_connected` too

mloskot:
LGTM

src/core/session.cpp
vadz: OK, but let's do it after I merge (if all goes well...) the real change (coming soon).

vadz:
BTW, should we disable MySQL tests in AppVeyor? They seem to be reliably crashing, but I have no idea why nor how to debug it. I also don't know why does MinGW build fail :-(

mloskot:
@vadz Let's disable for now (#550)

PR #549, opened 2017-07-18
Add get_dummy_from_{table,clause}() methods

vadz:
These functions are helpful to allow writing queries not using any
tables portably, i.e. they basically abstract the difference between
"select ... from dual" in Oracle and just "select ..." in almost all the
other backends.

### DIFF
--- a/include/soci/session.h
+++ b/include/soci/session.h
@@ -150,6 +150,19 @@ public:
     std::string empty_blob();
     std::string nvl();
 
+    // And some functions to help with writing portable DML statements.
+
+    // Get the name of the dummy table that needs to be used in the FROM clause
+    // of a SELECT statement not operating on any tables, e.g. "dual" for
+    // Oracle. The returned string is empty if no such table is needed.
+    std::string get_dummy_from_table() const;
+
+    // Returns a possibly empty string that needs to be used as a FROM clause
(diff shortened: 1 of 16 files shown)


include/soci/session.h
```diff
@@ -150,6 +150,19 @@ class SOCI_DECL session
     std::string empty_blob();
     std::string nvl();
 
+    // And some functions to help with writing portable DML statements.
+
+    // Get the name of the dummy table that needs to be used in the FROM clause
+    // of a SELECT statement not operating on any tables, e.g. "dual" for
+    // Oracle. The returned string is empty if no such table is needed.
+    std::string get_dummy_from_table() const;
```
mloskot: Wouldn't `get_dummy_table` be more obvious?

vadz: I expect this to be rarely used, i.e. I think most people would just use `get_dummy_from_clause()` anyhow (and I even didn't want to make this one public at all initially, although I finally ended up by doing it because why not, basically...), so I decided that keeping the name similar to the other function was more important than making it shorter.

vadz:
Applied as d91c2228ced2b6ba45df23dd02c838cfca9c9224

include/soci/session.h
mloskot: Makes sense

Issue #528, opened 2017-03-18
unresolved external symbol sqlite3_soci_error::result when linking soci_sqlite3_test.exe on VS2015 U3


sinall:
> Link:
>   C:\Program Files (x86)\Microsoft Visual Studio 14.0\VC\bin\link.exe /ERRORREPORT:QUEUE /OUT:"C:\Users\xxx\Document
>   s\Visual_Studio_2015\Projects\soci-build\bin\Debug\soci_sqlite3_test.exe" /INCREMENTAL /NOLOGO kernel32.lib user32.li
>   b gdi32.lib winspool.lib shell32.lib ole32.lib oleaut32.lib uuid.lib comdlg32.lib advapi32.lib ..\..\..\lib\Debug\soc
>   i_sqlite3_3_2.lib C:\Users\xxx\Documents\Visual_Studio_2015\Projects\Test2\Debug\x86\sqlcipher.lib C:\Users\xxx
>   \Documents\Visual_Studio_2015\Projects\Test2\Debug\x86\sqlcipher.lib ..\..\..\lib\Debug\soci_core_3_2.lib /MANIFEST /
>   MANIFESTUAC:"level='asInvoker' uiAccess='false'" /manifest:embed /DEBUG /PDB:"C:/Users/xxx/Documents/Visual_Studio
>   _2015/Projects/soci-build/bin/Debug/soci_sqlite3_test.pdb" /SUBSYSTEM:CONSOLE /TLBID:1 /DYNAMICBASE /NXCOMPAT /IMPLIB
>   :"C:/Users/xxx/Documents/Visual_Studio_2015/Projects/soci-build/lib/Debug/soci_sqlite3_test.lib" /MACHINE:X86 /SAF
>   ESEH  /machine:X86 "soci_sqlite3_test.dir\Debug\test-sqlite3.obj"
> test-sqlite3.obj : error LNK2019: unresolved external symbol "public: int __thiscall soci::sqlite3_soci_error::result(v
> oid)const " (?result@sqlite3_soci_error@soci@@QBEHXZ) referenced in function __catch$?test6@@YAXXZ$0 [C:\Users\xxx\D
> ocuments\Visual_Studio_2015\Projects\soci-build\backends\sqlite3\test\soci_sqlite3_test.vcxproj]
> C:\Users\xxx\Documents\Visual_Studio_2015\Projects\soci-build\bin\Debug\soci_sqlite3_test.exe : fatal error LNK1120:
[...]

PR #551, opened 2017-07-19
Define SOCI_SQLITE3_SOURCE in all SQLite3 backend sources

vadz:
Not doing it in session.cpp resulted in link errors in MinGW build as
SOCI_SQLITE3_DECL wasn't defined correctly in this case and the DLL
tried to import sqlite3_soci_error class instead of exporting it.

See #528.

### DIFF
--- a/src/backends/sqlite3/blob.cpp
+++ b/src/backends/sqlite3/blob.cpp
@@ -5,6 +5,7 @@
 // http://www.boost.org/LICENSE_1_0.txt)
 //
 
+#define SOCI_SQLITE3_SOURCE
 #include "soci/sqlite3/soci-sqlite3.h"
 
 #include <algorithm>
--- a/src/backends/sqlite3/row-id.cpp
+++ b/src/backends/sqlite3/row-id.cpp
@@ -5,6 +5,7 @@
 // http://www.boost.org/LICENSE_1_0.txt)
 //
 
+#define SOCI_SQLITE3_SOURCE
 #include "soci/sqlite3/soci-sqlite3.h"
 
 #ifdef _MSC_VER
--- a/src/backends/sqlite3/session.cpp
+++ b/src/backends/sqlite3/session.cpp
@@ -5,7 +5,7 @@
 // http://www.boost.org/LICENSE_1_0.txt)
 //
 
-
+#define SOCI_SQLITE3_SOURCE
 #include "soci/sqlite3/soci-sqlite3.h"
 
 #include "soci/connection-parameters.h"
--- a/src/backends/sqlite3/standard-into-type.cpp
+++ b/src/backends/sqlite3/standard-into-type.cpp
@@ -5,6 +5,7 @@
 // http://www.boost.org/LICENSE_1_0.txt)
 //
 
+#define SOCI_SQLITE3_SOURCE
 #include "soci/soci-platform.h"
 #include "soci/sqlite3/soci-sqlite3.h"
 #include "soci/rowid.h"
--- a/src/backends/sqlite3/standard-use-type.cpp
+++ b/src/backends/sqlite3/standard-use-type.cpp
@@ -5,6 +5,7 @@
 // http://www.boost.org/LICENSE_1_0.txt)
 //
 
+#define SOCI_SQLITE3_SOURCE
 #include "soci/soci-platform.h"
 #include "soci/sqlite3/soci-sqlite3.h"
 #include "soci/rowid.h"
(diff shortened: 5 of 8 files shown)


mloskot:
LGTM

mloskot:
@snikulov I'm taking Chandler Carruth's advise on that "don't let yourself be fooled into popularity of build systems" from a lecture I can't remember (https://twitter.com/mloskot/status/327070224044138499).

Chandler talked about portable C/C++ source code, all you need to compile is source code, and without the baggage of complexity of configuration specific to a particular build system. New SOCI maintainer may prefer to ditch CMake, users/developers may prefer to use eg. makefiles, and it is great if they have such freedom cheap.

That is my preference about writing portable multi-platform C++ code.
I'll leave it to @vadz to decide for this PR though.

snikulov:
@mloskot anyway it can be defined in single place, for ex. in #include "soci/sqlite3/soci-sqlite3.h"

mloskot:
Yes, but such is an implementation detail that can be easily refactored, if needed or if author of a PR finds time for refactoring, etc.

Having limited man power involved in the project, a better solution can easily become an enemy of just good solution.

vadz:
I thought about doing it in CMakefiles, but didn't because:

1. Other files in `src/backends/sqlite3` already did it like this.
2. Other files in other backends (e.g. MySQL) also did it like this.
3. I agree with @mloskot that it's better to put things in the code rather than build system if possible.
4. I hate CMake and prefer to avoid touching it as much as possible.

Notice that we can't just do this in `soci-sqlite3.h` itself because it can be included from outside the library too, while this symbol must only be defined when building the library itself.

Anyhow, I'm going to merge this because it's one of the changes needed to make the CI builds work and I'd really like to see them green before 4.0.0, but we could always change this and do it at the build system level if there are any real advantages to it later.

vadz:
Applied as 2f65f02e8fe70f72f277860d9c2adda466e5d5f0

mloskot:
@vadz just curious, why not close PRs via (appended) merge commit with:

```
Closes #551
```

vadz:
No special reason, I could have amended the commit message and added this line to it, but I just cherry-picked the commit locally and pushed it: this is faster/simpler than using Github web UI for me.

mloskot:
Right. I'm not suggesting to use the green button, no.

Since there is a PR, I just like to see  in logs
```
Merge pull request #XXX from contributor/branch
```
especially if it is from occasional `contributor` (not your case, obviously).

vadz:
Generally speaking, I try to avoid single commit merges as I don't think a merge commit really adds anything in this case, so I do a real merge if there are 2 or more commits but just rebase/cherry-pick if there is just one. Do you feel strongly about making merge commits even in the former case?

P.S. AppVeyor is finally green!

mloskot:
I agree there is need for merge commits in your case - it makes similar commit history if you did not open PR for your own change, which would be also fine I think (we don't need PR ourselves for every change, do we).
If, however,
* an external contributor submits PR
* PR is a brand new feature
* PR fixes a non-trivial bug
then it makes sense to keep track of PR merges, including PR numbers in the logs, IMHO.

So, I just asked to learn what's your take on that.

vadz:
I open PRs for my own changes to

1. See if there are any comments.
2. Test that they don't break the build.

So I think that they're always worth opening except for the really trivial changes not needing CI checks.

As for tracking the PR number, I agree that it should be mentioned in the commit message, the question was just about making or not the merge commits. I know there are people who forbid _any_ non-merge commits to master, and I even can see the point of it, but I don't think we need to be as fanatic about this, do we.

mloskot:
>  I know there are people who forbid any non-merge commits to master, and I even can see the point of it, but I don't think we need to be as fanatic about this, do we.

Although there is lots of policies around and although I appreciate having a recipe handy and expect things behave in certain way, I'm not trying to set any here or be fanatic really. I asked question about merge commits, because that automatically means inter-reference between commit (log) and PR on GH,
moreover, if PR fixes Issue and commit in PR mentions Issue ref, you have merge linked to Issue.
And, we get that for free using GitHub-style mentions ('Fixes', 'Closes'), no need to come back to GH and manually comment with SHA.

Sieving through commits, branches and PRs while collecting changelog or release notes is a hefty yet boring task. It is very helpful to be able to grep through `git log` to list all PRs and Issues.  I care about linear history less than being able to GTD.

TL;TR: All I'm asking for is the interlinking, however achieved

PR #552, opened 2017-07-19
Disable MySQL unit test currently always resulting in a crash

vadz:
Currently prepared statements do _not_ survive session::reconnect() due
to the issue #136, so running this test simply results in crashes.

Disable it for now, it should be reenabled when #156 or another solution
is applied.

---

So this isn't a solution in any way, shape or form, but at least it should allow tests to pass on AppVeyor instead of *always* failing.


vadz:
Applied as d7c986c6be97f512cb18fad3176a97c8de1ed188

PR #553, opened 2017-07-19
Suppress memory leak report in glibc NSS under Ubuntu

vadz:
There is absolutely nothing we can do about this leak, reported in both
MySQL and PostgreSQL tests for getpwuid() called from inside the
corresponding client libraries, so just suppress it.


vadz:
Applied as 4f1fbfcd4698dc8a4fcacf650635bb8fae75ada8, this does help a bit, even if it's not enough to fix Valgrind run yet.

PR #558, opened 2017-07-19
Try adding back fast_finish=true to Travis CI config

vadz:
Somehow the changes of f938c0133c9f213cc31939a36a160e036e09614a resulted
in Travis moving all the builds from Ubuntu Precise to Ubuntu Trusty,
which is not supported by many of our scripts.

Try restoring fast_finish option, removed by this commit, to see if it
could be due to it.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -5,6 +5,12 @@
 #
 language: cpp
 
+# Several scripts hard code Precise and don't currently work in any other
+# environment (e.g. the default Trusty).
+dist: precise
+
+sudo: required
+
 compiler:
     - g++
     - clang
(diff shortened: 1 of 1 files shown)


vadz:
@mloskot I'm going to merge this for now if you don't object as otherwise Travis CI is broken for all new PRs, which is inconvenient at best.

I still have no idea why did the changes in f938c01 result in switching to Trusty for all Travis builds, when they all ran under Precise previously...

mloskot:
@vadz I have no idea either. I don't see any changes that could have made the switch.

However, it seems to me the reason is coincidence of your change with updates on Travis CI:

> As of July, 18th 2017, we’re switching the default Linux distribution on Travis CI from Ubuntu Precise 12.04 LTS to Ubuntu Trusty 14.04 LTS

https://docs.travis-ci.com/user/precise-to-trusty-migration-guide/

You committed on 19 Jul, that is hours after Travis CI switched to Trusty.

PR #561, opened 2017-07-21
Fix handling of strings in ODBC backend bulk operations

vadz:
See #495.

### DIFF
--- a/docs/beyond.md
+++ b/docs/beyond.md
@@ -16,8 +16,12 @@ It can be useful to know how many rows were affected by the last SQL statement, 
 
 ---
 ##### Portability note:
-This method is currently not supported by the Oracle backend. It is however
-supported when using Oracle database via ODBC backend.
+This method behaviour in case of partially executed update, i.e. when some
+records were updated or inserted while some other have failed to be updated or
+inserted, depends on the exact backend and, in the case of ODBC backend, on the
+exact ODBC driver used. It can return `-1`, meaning that the number of rows is
+unknown, the number of rows actually updated or the total number of affected
+rows.
 ---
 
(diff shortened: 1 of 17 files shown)


mloskot:
This looks good, though I've learned ODBC drivers might be buggy or fussy. Is that what is happening with DB2 (CLI is de-facto ODBC model, isn't it?).

vadz:
DB2 failure is annoying, but I guess it either doesn't like empty strings in VARCHAR columns or allows putting only 19 characters in `VARCHAR(20)` (which would be really weird, but who am I to second guess IBM).

But what really bothers me are the failures in AppVeyor builds, I have no idea why does MS SQL ODBC test fail there, as it passes here and I don't see how could it be affected by these changes. SQLite test is apparently due to not supporting `size_t` output vector values, but I'll just work around it instead of trying to really deal with this for now.

vadz:
PS: DB2 code just has the same bug as ODBC (now I understand that this is what you probably meant in the first comment, sorry for not realizing it before), so I'll have to correct it there too. I wonder if anybody actually still uses DB2 in this day and age...

vadz:
OK, the only remaining test failure is the one in the ODBC test with `get_affected_rows()` returning 0 but, looking at the code, why does the test actually expect this to work? Our own documentation (in `backends.md`) says that this function can only be

> Called to determine the actual number of rows affected by data modifying statement.

and `get_number_of_rows()` can be used for `SELECT`s instead. So isn't the test simply wrong?

Or, alternatively, if `get_affected_rows()` is supposed to work for all statements, why do we even have `get_number_of_rows()` at all then?

PR #564, opened 2017-07-25
Some fix for SQLite3 backend

Arenoros:


vadz:
Thanks!

BTW, it looks like we don't have any tests for these methods. It would be great if some could be added, this would allow us to be sure that they really work and, maybe even more importantly, avoid breaking them in the future.

PR #566, opened 2017-08-06
Update structure of SOCI 4.0.0 documentation

mloskot:
Add MkDocs configuration to build documentation pages.

------

Next, decide where to host, keep on SF.net or move to GitHub Pages  with redirect from SF.net?


vadz:
I don't know if there is any advantage in moving to GH Pages, but just a question: how should the docs be built now? I had never used MkDocs before, it would be nice to have a README in the docs directory explaining what to do (and not to do, e.g. apparently raw HTML shouldn't be used any longer?), even if it's very short and trivial.

TIA!

mloskot:
> I don't know if there is any advantage in moving to GH Pages

I've been thinking about generating all docs with Travis CI build (eg. in the Empty backend build job) and, not necessarily, but it might be easier to deploy generated docs and website from to GH pages, than SF.net.

First, I will try to configure it for SF.net.

> how should the docs be built now? I had never used MkDocs before, it would be nice to have a README in the docs directory

Yes, indeed. Once I get the docs + website workflow, I'll add such README with instructions.

Meanwhile, if you fancy trying the new docs out:

```
pip install mkdocs
cd soci
mkdocs serve
```

PR #570, opened 2017-08-22
Sanitizer fix & cmake support for build in superprojects

mcencora:


include/soci/type-conversion.h
```diff
@@ -28,7 +28,12 @@ namespace details
 template <typename T>
 struct base_value_holder
 {
+    base_value_holder()
```
vadz: In principle, this code is not type-dependent, so it might be extracted into some `base_value_holder_type_indep` subclass from which `base_value_holder` would inherit to reduce "template bloat". I admit I'm not really sure if it's worth doing it in this case, but it might be.

include/soci/type-conversion.h
```diff
@@ -42,17 +47,15 @@ class conversion_into_type
     typedef typename type_conversion<T>::base_type base_type;
 
     conversion_into_type(T & value)
-        : into_type<base_type>(details::base_value_holder<T>::val_, ownInd_)
+        : into_type<base_type>(details::base_value_holder<T>::val_, details::base_value_holder<T>::ownInd_)
```
vadz: Why do we need to use `details::` here, inside details namespace? AFAICS it shouldn't be necessary and omitting it would make the code shorter and more readable.

include/soci/type-conversion.h
mcencora: That won't change anything here, since number of template instantiations will be same after moving this one member to a base non-template class, and there is no functionality to be moved to such non-template base - it is only for data storage.

include/soci/type-conversion.h
mcencora: indeed, I just added the extra namespace qualification to be consistent with rest of the code in this file.
Next commit removes them all

include/soci/type-conversion.h
vadz: Yes, you're right, sorry, no idea what I was think.

vadz:
Thanks for your contribution! I don't know much about CMake, but the changes look correct to me (except that I am not sure if this functionality is available in 2.8.10 that we currently require, shouldn't we bump this to 2.8.12?), so I took the risk of merging them too.

Just as a minor note, could you please submit separate PRs for unrelated changes in the future, e.g. in this case I think it would have been better to have one PR with the CMake changes and another one with the rest. TIA!

mloskot:
(Sorry for delay, I was/am on holidays)

Regarding CMake, I think we could safely bump the requirement to CMake 3.

vadz:
FWIW CMake 3 is _not_ easy to get if you have to use RHEL 6 (as I had, just recently). You can find prebuilt 64 bit binary, but if you're using a 32 bit system, you need to build it from source which, in turn, doesn't work out of the box at least for the latest CMake version (I didn't try the earlier ones because I got fed up with it and just built the library I had to build manually instead of using CMake).

PR #572, opened 2017-09-03
fix some of warnings from fresh Coverity report

alexott:
I just got fresh Coverity report & decided to fix new warnings about not initialized variables


mloskot:
Thanks a lot @alexott

PR #573, opened 2017-09-03
Avoid using overloaded virtual methods in vector type backends

vadz:
This is mostly done in order to allow SOCI headers to compile cleanly
with -Woverloaded-virtual (or corresponding MSVC warning C4266) in
effect, but it also allows to turn the same warning on for building SOCI
itself, which can be useful.

---

I'm submitting this because I'd like to avoid warnings when including `soci/odbc/soci-odbc.h` in my own project which enables this warning without having to put pragmas before/after including it and I think I might be not the only one doing this.

If renaming these methods is too contentious (I don't think there are any compatibility considerations in play here, but I could be missing something), there is also a possibility of just adding `using vector_into_type_backend::define_by_pos;` etc to all the derived classes instead.


mloskot:
👍 I think it is a good idea

PR #576, opened 2017-09-10
Add support for EditorConfig

alexott:
http://editorconfig.org/ helps developers define and maintain consistent coding styles
between different editors and IDEs.  Some IDEs have this support out of box, while some
are requiring separate plugin to be installed.

I've tried to create configuration based on the project's style wiki page & checking the
different files in project.

### DIFF
--- a/.editorconfig
+++ b/.editorconfig
@@ -1,3 +1,4 @@
+# This is the EditorConfig (http://editorconfig.org/) coding style file for SOCI.
 root = true
 
 [*]
@@ -6,6 +7,7 @@ insert_final_newline = true
 charset = utf-8
 indent_style = space
 indent_size = 4
+trim_trailing_whitespace = true
 
 [{Vagrantfile,CMakeLists.txt,*.cmake}]
 indent_size = 2
--- a/.editorconfig
+++ b/.editorconfig
@@ -1,0 +1,11 @@
+root = true
+
+[*]
+end_of_line = lf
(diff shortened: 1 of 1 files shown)


.editorconfig
```diff
@@ -0,0 +1,11 @@
+root = true
+
+[*]
+end_of_line = lf
+insert_final_newline = true
+charset = utf-8
+indent_style = space
+indent_size = 4
+
```
vadz: I also like `trim_trailing_whitespace = true`, perhaps we should add it (and also remove all the trailing whitespace in the current sources?)?

.editorconfig
```diff
@@ -0,0 +1,11 @@
+root = true
```
vadz: Could having something like

    # This is the EditorConfig (http://editorconfig.org/) coding style file for SOCI.

here helpful?

alexott:
Ok, I've updated config with these comments.

I'll write separate email to dev list about it - why it useful, where to take plugins, etc. & will ask people about bulk removal of trailing whitespace, just to avoid conflicts with people's work.

mloskot:
Although I like the  idea very much and i us  EditorConfig in almost all my
projects, I'm not sure if LF requirement is 'portable' and it might cause
annoyances when git is configured  with autocrlf as eg. Git/GitHub
recommend on Windows.

Nevertheless, I agree this should be merged and we can always improve
later.

PR #578, opened 2017-09-12
Add CLOB and XML support to Firebird backend

vadz:
This PR contains a refactoring of the existing tests and Firebird-specific changes.

I'd appreciate any comments about the need to add `blob_` to `firebird_standard_use_type_backend` in order to handle CLOBs. This seems wrong to me because it's so different from the way BLOBs are handled, but I just don't see any way to do it otherwise except changing the API and replacing the simple `long_string` struct with `clob` one similar to the existing `blob`, i.e. having its proper backend.

### DIFF
--- a/include/soci/firebird/soci-firebird.h
+++ b/include/soci/firebird/soci-firebird.h
@@ -85,6 +85,11 @@ struct firebird_standard_into_type_backend : details::standard_into_type_backend
 
     char *buf_;
     short indISCHolder_;
+
+private:
+    // Copy contents of a BLOB (represented by its id) in buf_ into the given
+    // string.
+    void copy_from_blob(std::string& out);
 };
 
 struct firebird_vector_into_type_backend : details::vector_into_type_backend
@@ -141,6 +146,11 @@ struct firebird_standard_use_type_backend : details::standard_use_type_backend
 
     char *buf_;
     short indISCHolder_;
+
+private:
+    // Allocate a temporary blob, fill it with the data from the provided
+    // string and copy its ID into buf_.
+    void copy_to_blob(const std::string& in);
 
     // This is used for types mapping to CLOB.
     firebird_blob_backend* blob_;
--- a/include/soci/firebird/soci-firebird.h
+++ b/include/soci/firebird/soci-firebird.h
@@ -59,6 +59,7 @@ enum BuffersType
     eStandard, eVector
 };
 
+struct firebird_blob_backend;
 struct firebird_statement_backend;
 struct firebird_standard_into_type_backend : details::standard_into_type_backend
 {
@@ -117,7 +118,8 @@ struct firebird_vector_into_type_backend : details::vector_into_type_backend
 struct firebird_standard_use_type_backend : details::standard_use_type_backend
 {
     firebird_standard_use_type_backend(firebird_statement_backend &st)
-        : statement_(st), data_(NULL), type_(), position_(0), buf_(NULL), indISCHolder_(0)
+        : statement_(st), data_(NULL), type_(), position_(0), buf_(NULL), indISCHolder_(0),
+          blob_(NULL)
     {}
 
     virtual void bind_by_pos(int &position,
@@ -139,6 +141,9 @@ struct firebird_standard_use_type_backend : details::standard_use_type_backend
 
     char *buf_;
     short indISCHolder_;
+
+    // This is used for types mapping to CLOB.
+    firebird_blob_backend* blob_;
 };
 
 struct firebird_vector_use_type_backend : details::vector_use_type_backend
--- a/src/backends/firebird/standard-into-type.cpp
+++ b/src/backends/firebird/standard-into-type.cpp
@@ -126,26 +126,39 @@ void firebird_standard_into_type_backend::exchangeData()
             {
                 long_string* const dst = static_cast<long_string*>(data_);
 
-                firebird_blob_backend blob(statement_.session_);
-                blob.assign(*reinterpret_cast<ISC_QUAD*>(buf_));
+                copy_from_blob(dst->value);
+            }
+            break;
 
-                std::size_t const len_total = blob.get_len();
-                dst->value.resize(len_total);
+        case x_xmltype:
+            {
+                xml_type* const dst = static_cast<xml_type*>(data_);
 
-                std::size_t len_read = blob.read(0, &dst->value[0], len_total);
-                if (len_read != len_total)
-                {
-                    std::ostringstream os;
-                    os << "Read " << len_read << " bytes instead of expected "
-                       << len_total << " from Firebird text blob object";
-                    throw soci_error(os.str());
-                }
+                copy_from_blob(dst->value);
             }
             break;
 
         default:
             throw soci_error("Into element used with non-supported type.");
     } // switch
+}
+
+void firebird_standard_into_type_backend::copy_from_blob(std::string& out)
+{
+    firebird_blob_backend blob(statement_.session_);
+    blob.assign(*reinterpret_cast<ISC_QUAD*>(buf_));
+
+    std::size_t const len_total = blob.get_len();
+    out.resize(len_total);
+
+    std::size_t const len_read = blob.read(0, &out[0], len_total);
+    if (len_read != len_total)
+    {
+        std::ostringstream os;
+        os << "Read " << len_read << " bytes instead of expected "
+           << len_total << " from Firebird text blob object";
+        throw soci_error(os.str());
+    }
 }
 
 void firebird_standard_into_type_backend::clean_up()
--- a/src/backends/firebird/standard-into-type.cpp
+++ b/src/backends/firebird/standard-into-type.cpp
@@ -11,6 +11,8 @@
 #include "firebird/common.h"
 #include "soci/soci.h"
 
+#include <sstream>
+
 using namespace soci;
 using namespace soci::details;
 using namespace soci::details::firebird;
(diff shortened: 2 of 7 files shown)


include/soci/firebird/soci-firebird.h
```diff
@@ -84,6 +85,10 @@ struct firebird_standard_into_type_backend : details::standard_into_type_backend
 
     char *buf_;
     short indISCHolder_;
+
+private:
+    // Copy contents of a BLOB in buf_ into the given string.
+    void init_from_blob(std::string& out);
```
mloskot: IMO, `copy_from_blob` name would be clearer and would spare the comment

include/soci/firebird/soci-firebird.h
```diff
@@ -139,6 +145,14 @@ struct firebird_standard_use_type_backend : details::standard_use_type_backend
 
     char *buf_;
     short indISCHolder_;
+
+private:
+    // Allocate a temporary blob, fill it with the data from the provided
+    // string and copy its ID into buf_.
+    void init_as_blob(const std::string& in);
```
mloskot: Same here, `copy_as_blob`. Especially, those `init` functions are not called from constructor.

src/backends/firebird/standard-into-type.cpp
```diff
@@ -119,11 +121,46 @@ void firebird_standard_into_type_backend::exchangeData()
                 blob->assign(*reinterpret_cast<ISC_QUAD*>(buf_));
             }
             break;
+
+        case x_longstring:
+            {
+                long_string* const dst = static_cast<long_string*>(data_);
+
```
mloskot: Do we need those empty lines?

It's matter of taste and perhaps belongs to the latest style and formatting and `clang-format` discussion.
I just find SOCI code a bit (too) sparse :-)

src/backends/firebird/standard-into-type.cpp
```diff
@@ -119,11 +121,46 @@ void firebird_standard_into_type_backend::exchangeData()
...
 
+void firebird_standard_into_type_backend::init_from_blob(std::string& out)
+{
+    firebird_blob_backend blob(statement_.session_);
+    blob.assign(*reinterpret_cast<ISC_QUAD*>(buf_));
+
+    std::size_t const len_total = blob.get_len();
+    out.resize(len_total);
+
+    std::size_t len_read = blob.read(0, &out[0], len_total);
```
mloskot: `std::size_t const len_read` ?

tests/common-tests.h
```diff
@@ -4260,6 +4279,97 @@ TEST_CASE_METHOD(common_tests, "String length", "[core][string][length]")
     CHECK(vout[2].length() == 20);
 }
 
+// Helper function used in two tests below.
+static std::string make_long_xml_string()
+{
+    std::string s;
+    s.reserve(10000);
```
mloskot: Is this arbitrary size or is there any rationale behind?

mloskot:
@vadz If I knew a bit about Firebird, I'd help about the `clob` and `blob` situation. Sorry.

include/soci/firebird/soci-firebird.h
vadz: OK, I'll change the function names to use `copy_` instead of `init_`. I'd rather keep the comment though, this code is not really obvious to follow.

src/backends/firebird/standard-into-type.cpp
vadz: I mostly try to emulate the existing style and the `case x_blob` just above does use them.

But, to be honest, I prefer too many blank lines to not enough of them, personally...

src/backends/firebird/standard-into-type.cpp
vadz: Indeed, thanks.

tests/common-tests.h
vadz: The string constructed below is `200*26 + 12` bytes long, so I guess I could use its exact length instead, but I just lazily rounded up...

src/backends/firebird/standard-into-type.cpp
mloskot: Sure

include/soci/firebird/soci-firebird.h
mloskot: Sure

tests/common-tests.h
mloskot: If I use 'magic' numbers, I'd put a tiny comment

```
s.reserve(10000); // arbitrary large
```

That makes it clear to the reader that this is not a special number, documented anywhere or required by an (presumably) thrid-party API, etc. Less questions raised, simpler to comprehend.

vadz:
@mloskot Thanks for the review!

This is not really about Firebird though, but more a general API/design question: to me both approaches make sense, but using both of them at the same time -- not so much.

To make it more explicit, the two approaches I'm speaking about are

1. The one used for BLOBs: here we have a `soci::blob` class which has `soci::details::blob_backend` inside it and this backend contains database-specific BLOB handle. We also have `exchange_traits<soci::blob>` declared in `soci/blob-exchange.h`.
1. The one used for CLOBs (and for XML): here we have just a trivial `soci::long_string` wrapper and `exchange_traits<long_string>` for it declared in `soci/exchange-traits.h` (I give the file names just to show that there is inconsistency here too). But there is no backend-specific classes for CLOBs, which means that if we need any extra data, we have to store it in `$database_use_type_backend` instead, as I did for Firebird (there is no need to do this for "into" because we can just create a temporary BLOB, but for "use" direction we really need to store it somewhere until the statement finishes executing).

Again, I can see pro and cons of both approaches, but I don't like using both of them. And as the first one is present since a long time, I think that we should seriously consider changing the CLOB API to follow it. I really hope @msobczak can comment on this.

tests/common-tests.h
sinall: Maybe could use 3 constants for those 3 number in expression 200*26 + 12.

mloskot:
@vadz Sorry, I know nothing about the design rationale. I hope @msobczak will help.

msobczak:
Hi there,
The reason for a different implementation of xml_type/long_string (from type-wrappers.h) and older blob type is due to their inherently different API and different handling in major backends (Oracle/PostgreSQL).
BLOB objects behave like file handles and have file-like API in both Oracle and PostgreSQL - that is, they are operated via functions like open/close/read/write. This makes them heavy in the implementation sense and they need their own backend handles in every target server.
Long strings and XML objects are different, as they are supposed to behave like strings. In fact, in PostgreSQL they *are* just strings and if SOCI were written only for PostgreSQL, there would be no need to make a separate type for them. But in Oracle they deserve a different handling due to Oracle's archaic limitation to 4000 characters on regular strings; internally, long strings are CLOBs and they are operated on via LOB API. But this is an implementation detail and from the library user perspective, they are again just strings, just potentially longer than 4k (in Oracle there are also some special XML-oriented functions, but they exist in the SQL space only and have no impact on the C API). This means that long strings in Oracle cannot exist as std::strings in the user space, because regular std::strings are operated via functions that have the 4k limitation and longer strings need different handling. The purpose of the type wrapper is only to hide the std::string type from the into/use dispatching mechanism and make it distinguishable from the backend.
So, the picture is this:
1. we have BLOB, which is not really a data type, it's a handle to the DB-side file object, which deserves a dedicated class with file-like API,
2. we have short and long strings, which do not deserve distinct API, but need to be distinguished by the backend so that different methods are used internally to push and fetch the string content.

That's why a lightweight type wrapper was introduced to handle strings, instead of more heavyweight construct that was justified for BLOBs.

Now, if you discover that some database (Firebird?) has a different methods of accessing long strings (CLOBs?), then please have a look at src/backends/oracle/standard-into-type.cpp, and see how x_longstring and x_xmltype are dispatched. These operations do rely on some backend handles, but they are already available at that level and do not need to be carried with the string object itself. In other words, the fact that long_string does not offer any specific API to the user, means that it does not need to be tied to the backend all the time. On the other hand, the BLOB class has its own backend, because it offers a stand-alone API to the user that can be operated on outside of the into/use patterns - that is, BLOB can be manipulated outside of the statement object and for this reason needs to be tied to the necessary backend handles.
I hope this justifies the implementation. I have no idea how Firebird works, but I can guess that the above explanation has sufficient analogies to be transferred to the Firebird space.

vadz:
Thanks for the explanation, but it's not completely satisfying. I agree that PostgreSQL is nice to work with and we don't need anything special when using it, but the API should be generic enough to allow implementation for all backends. And Oracle actually seems to be exactly the same as Firebird (or vice versa) in this respect: in both backends, CLOBs are handled in the same way as BLOBs. And I just hadn't realized it before, but you added `ociData_` to Oracle code exactly for the same reason I've added `blob_` to Firebird in this PR, so the implementations are pretty similar except that I reuse the existing `firebird_blob_backend` code while you partially duplicate it, e.g. `oracle_standard_into_type_backend::read_from_lob()` is basically the same as the combination of `oracle_blob_backend::get_len()` and `oracle_blob_backend::read()`. Looking at Oracle code again, it really seems like the same could be done in it to avoid code duplication and make it more type-safe (currently `ociData_` is `void*` when it's actually always an `OCILobLocator*`, as in `oracle_blob_backend`).

Again, this does work, but there is really no reason it couldn't be made to work for BLOBs too: nothing prevents us from (hypothetically) defining `struct long_data { std::string value; }` and handling it in a similar way to `long_string`. So it all returns to the question of consistency.

I still think that it would be more clear to introduce `clob` class similar to `blob` (and mostly reusing the same implementation, of course). But if you object to this, we'll keep the current approach. I'd just like this decision to be taken before I implement CLOB/XML support for ODBC driver...

mloskot:
Some time ago I was considering `clob` class. I think it would be a good addition, for symmetry too. Users would grasp it as buffers and large value builders for use with SQL. Some memory optimisation would be possible too.

msobczak:
The way I see it is that BLOB is a file-like object, whereas long_string and xml_type are string-like objects. They have different APIs and cater to different needs. I also expect that they will be used to handle data of different sizes, even if they migh thave the same theoretical limits. That is, I'd expect BLOB to be used to handle opaque data that in practice are high-volume media files, whereas strings will be, well, just strings, of sizes in the kB to few MB range. This also means that the benefits of different memory management for BLOBs will not be relevant for strings.
From this perspective, adding a separate CLOB class does not solve any actual user problem.

mloskot:
> From this perspective, adding a separate CLOB class does not solve any actual user problem.

No, it doesn't. 

The way I looked at it when I contemplated about CLOB is that both, BLOB and CLOB, are facades that simplify managing the file-like and the memory resources by providing unified API. Both can offer exactly the same interface, `get_len`, `read`, `write`, `append`, etc. That fact makes `clob` class a compelling idea.

However, I undesrtand Maciej's point. I remember similar discussion when I was not quite happy to use `std::string` for binary data as in array of raw bytes :-)

vadz:
I still think having clob would be better, but I agree that it doesn't have any practical advantages right now, just consistency. So I'll merge this one as is then and implement these types as synonyms for string for ODBC (because I don't see what else can be done there).

PR #579, opened 2017-09-13
Restore Oracle build job on Travis CI

### LINKED ISSUE #554: Make Oracle work again on Travis CI
Unfortunately the current scripts don't work any longer and this is a pity because Oracle is a backend that is not simple to test locally -- yet important enough to care about not breaking it.

It looks like the simplest way to use it on Travis is https://github.com/cbandy/travis-oracle, so perhaps we should try doing this instead.

mloskot:
Clones and uses scripts to install Oracle Database XE (64-bit) on Travis CI from https://github.com/cbandy/travis-oracle.
Switch Oracle backend build from 32-bit to 64-bit.
Closes #554 

------

The Oracle downloader (@cbandy/travis-oracle) requires two secret environment variables.

The CI builds for PRs do not work anymore:
> Pull requests sent from forked repositories do not have access to encrypted variables or data.
https://docs.travis-ci.com/user/pull-requests/#Pull-Requests-and-Security-Restrictions

We will have to apply some workaround (eg. run Oracle build job only for regular pushes).
Meanwhile, I'm merging this PR, so we see if the Oracle build is properly triggered.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -3,7 +3,7 @@
 #
 # Copyright (c) 2013 Mateusz Loskot <mateusz@loskot.net>
 #
-# TODO: Switc to Trusty https://github.com/SOCI/soci/issues/574
+# TODO: Switch to Trusty https://github.com/SOCI/soci/issues/574
 dist: precise
 language: cpp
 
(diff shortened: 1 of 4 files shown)


vadz:
No real comments, just wanted to say "thanks" for working on this!

PR #584, opened 2017-09-14
Skip running Oracle tests for PR builds on Travis CI

vadz:
This doesn't work currently because Oracle installation requires access
to secret environment variables which are unavailable for the PR builds
for the security purposes, so skip trying to do it to at least avoid
failing all PR builds.

### DIFF
--- a/scripts/travis/before_install_oracle.sh
+++ b/scripts/travis/before_install_oracle.sh
@@ -4,6 +4,11 @@
 # Uses Oracle downloader and installer from https://github.com/cbandy/travis-oracle
 #
 # set -ex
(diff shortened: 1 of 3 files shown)


vadz:
Oops, I didn't realize that we couldn't even compile Oracle code without installing Oracle as we don't have the headers neither. Looks like we just need to skip this build entirely for the PRs :-(

vadz:
@mloskot I'm committing this in order to be able to test the other PRs, but feel free to change it if you see some better way to deal with this issue.

mloskot:
@vadz LGTM. I will change it so Oracle works for PRs properly.

PR #586, opened 2017-09-14
More XML/CLOB-related changes, including ODBC "support"

vadz:
In practice, using "xml" columns in SQL Server database seems to work just fine, even without using the `SQL_SS_XML` column type specific to Microsoft ODBC driver (and not supported by e.g. FreeTDS driver which is commonly used on Unix systems).

### DIFF
--- a/docs/lobs.md
+++ b/docs/lobs.md
@@ -54,3 +54,10 @@ For PostgreSQL, these types are:
 
 For Firebird, there is no special XML support, but `BLOB SUB_TYPE TEXT` can be
 used for storing it, as well as long strings.
+
+For ODBC backend, these types depend on the type of the database connected to.
+In particularly important special case of Microsoft SQL Server, these types
+are:
+
+* `xml`
+* `text`
--- a/docs/lobs.md
+++ b/docs/lobs.md
@@ -61,3 +61,11 @@ are:
 
 * `xml`
 * `text`
+
+When using ODBC backend to connect to a PostgreSQL database, please be aware
+that by default PostgreSQL ODBC driver truncates all "unknown" types, such as
+XML, to maximal varchar type size which is just 256 bytes and so is often
+insufficient for XML values in practice. It is advised to set the
+`UnknownsAsLongVarchar` connection option to 1 to avoid truncating XML strings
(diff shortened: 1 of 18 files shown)


docs/lobs.md
```diff
@@ -51,3 +51,21 @@ For PostgreSQL, these types are:
...
+are:
+
+* `xml`
+* `text`
+
+When using ODBC backend to connect to a PostgreSQL database, please be aware
+that by default PostgreSQL ODBC driver truncates all "unknown" types, such as
+XML, to maximal varchar type size which is just 256 bytes and so is often
+insufficient for XML values in practice. It is advised to set the
+`UnknownsAsLongVarchar` connection option to 1 to avoid truncating XML strings
```
mloskot: BTW, there was discussion about those variables recently on pgsql-odbc, [Bug? Data Type Option 'Text as LongVarChar' not working anymore?](https://www.postgresql.org/message-id/6d56a09b-a42f-1d8e-8e88-28c57376db44%40dream.email.ne.jp)
quoting the driver author:

> I want to eliminate *Text as LongVarChar*, *Unknowns as LongVarChar* or *Unknown Sizes* option if possible.
> BTW *UNKNOWNS as LongVarChar* option(default off) would be INVALID for PG10 servers.
> Are there any problems?
> Because 'unknown' type would be 'text' type in PG10, 'unknown' items before would be controlled by *Text as LongVarChar* option(default on).
>

mloskot:
LGTM

docs/lobs.md
vadz: Hmm, interesting, thanks. So maybe this `SQLSetConnectAttr(SQL_ATTR_PGOPT_UNKNOWNSASLONGVARCHAR)` call is not worth doing, finally, if it's only going to end up being supported between 9.6.300 and 10.0 driver versions... OTOH it probably doesn't really hurt and doesn't take long (as it doesn't involve a trip to the database, but just sets a flag in some driver data structure locally).

mloskot: Yes, it doesn't hurt. It's better to stick to the current released version of the driver and not take any for-future actions, yet.

PR #591, opened 2017-09-21
Revert 1 master

hrabe:


mloskot:
Hmm, a mistake?

hrabe:
Would say yes. Was overslept. Was patching tests core cause did not compile with MSVC compilers (8.0) as example. Know that this is a way old but needs to be supported for a short timeperiod too

mloskot:
OK, no problem. Good idea. Patches welcome :)

PR #593, opened 2017-09-22
Repair VC8 compile of test applications

hrabe:
Due to unsupported default param usage in macros and no qualified given _WIN32_WINNT version the test apps will not compile with some of Microsoft compilers.
Exception Features used are only available if a minimum version is set due to conditional enabling in _windows.h_ file (and sub required files)

### DIFF
--- a/include/soci/soci-platform.h
+++ b/include/soci/soci-platform.h
@@ -75,7 +75,7 @@ namespace std {
 //define DLL import/export on WIN32
 #ifdef _WIN32
 # ifndef _WIN32_WINNT
-#   define _WIN32_WINNT 0x0500
+#   define _WIN32_WINNT 0x0502 //_WIN32_WINNT_WS03, VS2015 support: https://msdn.microsoft.com/de-de/library/6sehtctf.aspx
 # endif // _WIN32_WINNT
 # ifdef SOCI_DLL
 #  ifdef SOCI_SOURCE
--- a/include/soci/soci-platform.h
+++ b/include/soci/soci-platform.h
@@ -74,6 +74,7 @@ namespace std {
 
 //define DLL import/export on WIN32
 #ifdef _WIN32
+# define _WIN32_WINNT 0x0500
 # ifdef SOCI_DLL
 #  ifdef SOCI_SOURCE
 #   define SOCI_DECL __declspec(dllexport)
--- a/include/soci/soci-platform.h
+++ b/include/soci/soci-platform.h
@@ -74,7 +74,9 @@ namespace std {
 
 //define DLL import/export on WIN32
 #ifdef _WIN32
-# define _WIN32_WINNT 0x0500
+# ifndef _WIN32_WINNT
+#   define _WIN32_WINNT 0x0500
+# endif // _WIN32_WINNT
 # ifdef SOCI_DLL
 #  ifdef SOCI_SOURCE
 #   define SOCI_DECL __declspec(dllexport)
--- a/tests/common-tests.h
+++ b/tests/common-tests.h
@@ -537,7 +537,7 @@ TEST_CASE_METHOD(common_tests, "Exception on not connected", "[core][exception]"
     CHECK_THROWS_AS(sql.get_last_insert_id(s, l), soci_error);
 }
 
-TEST_CASE_METHOD(common_tests, "Basic functionality")
+TEST_CASE_METHOD(common_tests, "Basic functionality", "basics")
 {
     soci::session sql(backEndFactory_, connectString_);
 
@@ -570,7 +570,7 @@ TEST_CASE_METHOD(common_tests, "Use and into", "[core][into]")
 
     auto_table_creator tableCreator(tc_.table_creator_1(sql));
 
-    SECTION("Round trip works for char")
+    SECTION("Round trip works for char", "section")
     {
         char c('a');
         sql << "insert into soci_test(c) values(:c)", use(c);
@@ -578,7 +578,7 @@ TEST_CASE_METHOD(common_tests, "Use and into", "[core][into]")
         CHECK(c == 'a');
     }
 
-    SECTION("Round trip works for string")
+    SECTION("Round trip works for string", "section")
     {
         std::string helloSOCI("Hello, SOCI!");
         sql << "insert into soci_test(str) values(:s)", use(helloSOCI);
@@ -587,7 +587,7 @@ TEST_CASE_METHOD(common_tests, "Use and into", "[core][into]")
         CHECK(str == "Hello, SOCI!");
     }
 
-    SECTION("Round trip works for short")
+    SECTION("Round trip works for short", "section")
     {
         short three(3);
         sql << "insert into soci_test(sh) values(:id)", use(three);
@@ -596,7 +596,7 @@ TEST_CASE_METHOD(common_tests, "Use and into", "[core][into]")
         CHECK(sh == 3);
     }
 
-    SECTION("Round trip works for int")
+    SECTION("Round trip works for int", "section")
     {
         int five(5);
         sql << "insert into soci_test(id) values(:id)", use(five);
@@ -605,7 +605,7 @@ TEST_CASE_METHOD(common_tests, "Use and into", "[core][into]")
         CHECK(i == 5);
     }
 
-    SECTION("Round trip works for unsigned long")
+    SECTION("Round trip works for unsigned long", "section")
     {
         unsigned long seven(7);
         sql << "insert into soci_test(ul) values(:ul)", use(seven);
@@ -614,7 +614,7 @@ TEST_CASE_METHOD(common_tests, "Use and into", "[core][into]")
         CHECK(ul == 7);
     }
 
-    SECTION("Round trip works for double")
+    SECTION("Round trip works for double", "section")
     {
         double pi(3.14159265);
         sql << "insert into soci_test(d) values(:d)", use(pi);
@@ -623,7 +623,7 @@ TEST_CASE_METHOD(common_tests, "Use and into", "[core][into]")
         ASSERT_EQUAL(d, pi);
     }
 
-    SECTION("Round trip works for date without time")
+    SECTION("Round trip works for date without time", "section")
     {
         std::tm nov15 = std::tm();
         nov15.tm_year = 105;
@@ -645,7 +645,7 @@ TEST_CASE_METHOD(common_tests, "Use and into", "[core][into]")
         CHECK(t.tm_sec  == 0);
     }
 
-    SECTION("Round trip works for date with time")
+    SECTION("Round trip works for date with time", "section")
     {
         std::tm nov15 = std::tm();
         nov15.tm_year = 105;
@@ -667,7 +667,7 @@ TEST_CASE_METHOD(common_tests, "Use and into", "[core][into]")
         CHECK(t.tm_sec  == 17);
     }
 
-    SECTION("Indicator is filled correctly in the simplest case")
+    SECTION("Indicator is filled correctly in the simplest case", "section")
     {
         int id(1);
         std::string str("Hello");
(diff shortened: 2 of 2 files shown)


mloskot:
> with some of Microsoft compilers.

What are these?

The Catch does not seem to like the change, I'm afraid.
The AppVeyor build is failing.

hrabe:
The AppVeyor show similar things like my local build for ODBC.
```C:\projects\soci\tests\catch.hpp(6423): error C3861: 'AddVectoredExceptionHandler': identifier not found [C:\projects\soci\build.msvc2012\tests\sqlite3\soci_sqlite3_test_static.vcxproj]```
This should be fixed with the define inside _soci/soci-platform.h_ related to this Remarks: https://msdn.microsoft.com/de-de/library/windows/desktop/ms679274(v=vs.85).aspx 
Seems that something will not be compiled using platform?

File: include/soci/soci-platform.h [hunk 1]
```diff
@@ -74,6 +74,7 @@ namespace std {
 
 //define DLL import/export on WIN32
 #ifdef _WIN32
+# define _WIN32_WINNT 0x0500
```
mloskot: I'd do:
```
# ifndef _WIN32_WINNT
#  define _WIN32_WINNT 0x0500
# endif
```

tests/common-tests.h
```diff
@@ -537,7 +537,7 @@ TEST_CASE_METHOD(common_tests, "Exception on not connected", "[core][exception]"
     CHECK_THROWS_AS(sql.get_last_insert_id(s, l), soci_error);
 }
 
-TEST_CASE_METHOD(common_tests, "Basic functionality")
+TEST_CASE_METHOD(common_tests, "Basic functionality", "basics")
```
mloskot: Hmm, isn't the third argument of [TEST_CASE_METHOD](https://github.com/philsquared/Catch/blob/master/docs/test-fixtures.md) supposed to be a tag?

Aren't [Catch tags](https://github.com/philsquared/Catch/blob/master/docs/test-cases-and-sections.md) supposed to be in form of `["tag1", "tag2"]`?

tests/common-tests.h
```diff
@@ -570,15 +570,15 @@ TEST_CASE_METHOD(common_tests, "Use and into", "[core][into]")
 
     auto_table_creator tableCreator(tc_.table_creator_1(sql));
 
-    SECTION("Round trip works for char")
+    SECTION("Round trip works for char", "section")
```
mloskot: Same here, I'd make it a proper tag like `["char"]`

tests/common-tests.h
```diff
@@ -570,15 +570,15 @@ TEST_CASE_METHOD(common_tests, "Use and into", "[core][into]")
...
     {
         char c('a');
         sql << "insert into soci_test(c) values(:c)", use(c);
         sql << "select c from soci_test", into(c);
         CHECK(c == 'a');
     }
 
-    SECTION("Round trip works for string")
+    SECTION("Round trip works for string", "section")
```
mloskot: Same here, I'd make it a proper tag like `["string"]`

tests/common-tests.h
```diff
@@ -587,7 +587,7 @@ TEST_CASE_METHOD(common_tests, "Use and into", "[core][into]")
         CHECK(str == "Hello, SOCI!");
     }
 
-    SECTION("Round trip works for short")
+    SECTION("Round trip works for short", "section")
```
mloskot: Same here, I'd make it a proper tag like `["short"]`

etc.

tests/common-tests.h
```diff
@@ -596,7 +596,7 @@ TEST_CASE_METHOD(common_tests, "Use and into", "[core][into]")
         CHECK(sh == 3);
     }
 
-    SECTION("Round trip works for int")
+    SECTION("Round trip works for int", "section")
```
mloskot: Same here, I'd make it a proper tag like `["int"]`

**Unless**, we are going to get rid of it once SOCI switches to C++11 anyway.

If so, we may agree to replace all those with `["dummy"]` (or even empty string if kosher for Catch) as we don't seem to use the tags currently anyway. That would be simpler to re-work the PR with search & replace.

Either would work for me.

hrabe:
After updating the PR AppVeyor passes the MinGW build and tests.
Only the VS based builds have probblems. It seems to me that files are not trashed correctly?
Something other than this PR may also influence the 3 Studio builds.

mloskot:
> Something other than this PR may also influence the 3 Studio builds.

Let's see if the master builds https://ci.appveyor.com/project/mloskot/soci/build/4.0.0.287

vadz:
The additions of `"[dummy]"` are really ugly. I do still use VC9 for a project also using CATCH and didn't have any problems like this, is this for VC8 only? It would be nice to find another way to work around this...

And if not, I think there ought to be a macro like `DUMMY_TAGS_FOR_MSVC8` or something like this to allow getting rid of it easily when support for this compiler is dropped.

hrabe:
Ok, I will try to find another way but the issue at VC8 by one of many examples is:
```
2>test-odbc-access.cpp
2>C:\Github\hrabe\soci\tests\common-tests.h(573) : warning C4003: not enough actual parameters for macro 'SECTION'
2>C:\Github\hrabe\soci\tests\common-tests.h(573) : error C2059: syntax error : ')'
``` 
on _C:\Github\hrabe\soci\tests\common-tests.h(573)_ as:
```
    SECTION("Round trip works for char")
```

mloskot:
We're at Catch v1.9.6. Could you try to pull the latest (single header) and check if it works?

Meanwhile, I asked to confirm VC++ 8.0 support https://github.com/philsquared/Catch/issues/1029

hrabe:
Will try that, but there seems to be a problem on MS side: https://connect.microsoft.com/VisualStudio/feedback/details/380090/variadic-macro-replacement

hrabe:
Used now 
```
 *  Catch v1.10.0
 *  Generated: 2017-08-26 15:16:46.676990
```
for compile but *no luck at all*, the problem of VC 8 variadic macros with too less argument persist!

mloskot:
@hrabe Thanks for checking Catch 1.10.0. I think we can safely assume VC8 is not (fully) supported any longer.

hrabe:
So how to solve? Drop support of VC 8.0 is no option to me.
Any thoughts to proceed if this won't be addresses by Catch?
BTW: Tested 2.0 branch of Catch too, I'm afraid this will bring much more dropped support!

mloskot:
@hrabe 
> So how to solve? 

I, personally, don't care about anything older than VS2015 (and GCC/clang with equivalent C++ support).

Nevertheless, I understand
* @vadz needs support for older compilers (not sure which versions Vadim needs though)
* there might be users who need suport for older compilers (you, @hrabe)

> Any thoughts how to proceed if this won't be addressed by Catch

Any chance to address @vadz suggestion https://github.com/SOCI/soci/pull/593#issuecomment-331413699 ?

Then, either Vadim or myself, we'll be happy to merge this PR in.

---

FYI, until SOCI team decides to switch to C++11 or later, I'm fine accepting PRs with any (reasonable) plumbing required by the older compilers. I, however, am not able to provide any assistance myself or promise I won't break anything with some commits in future - simply, I have no means to try such old compilers. Unless AppVeyor can help us.

hrabe:
Would try the macro for dummy but what would you say @vadz and @mloskot about this:
```TEST_CASE_METHOD(common_tests, "Use and into", "[core][into]")
{
    soci::session sql(backEndFactory_, connectString_);

    auto_table_creator tableCreator(tc_.table_creator_1(sql));

    SECTION("Round trip works for char", "[core][into][char]")
    {
        char c('a');
        sql << "insert into soci_test(c) values(:c)", use(c);
        sql << "select c from soci_test", into(c);
        CHECK(c == 'a');
    }

    SECTION("Round trip works for string", "[core][into][string]")
    {
        std::string helloSOCI("Hello, SOCI!");
        sql << "insert into soci_test(str) values(:s)", use(helloSOCI);
        std::string str;
        sql << "select str from soci_test", into(str);
        CHECK(str == "Hello, SOCI!");
    }

    SECTION("Round trip works for short" , "[core][into][short]")
...
```
Why not using the tags really without _dummy_ but with prefixed TestMethod Tag and Tag for current section like _char_, _string_, _short_ etc?

vadz:
I'm not sure, can tags be actually used with sections? I've never done this... If they can, then using real tags is indeed better, but if they can't, then this would be surprising/confusing, so I'd still prefer the dummy macro idea.

In fact... what about something like this:

    #if defined(_MSC_VER) && (_MSC_VER < 1500)
    #undef SECTION
    #define SECTION(name) INTERNAL_CATCH_SECTION(name, "dummy-for-vc8")
    #endif

Shouldn't this solve the problem for it?

mloskot:
AFAIK, tags work for `TEST_CASE[_METHOD]` and `SCENARIO` macros only.


> In fact... what about something like this:
> 
>     #if defined(_MSC_VER) && (_MSC_VER < 1500)
>     #undef SECTION
>     #define SECTION(name) INTERNAL_CATCH_SECTION(name, "dummy-for-vc8")
>     #endif
> 
> Shouldn't this solve the problem for it?

👍

vadz:
In fact the second parameter of the `SECTION` macro is "description", not "tag" at all.

So we should either add real descriptions to all of them, which seems to be a bit too verbose and redundant with the name, but bearable, or use the hack for VC8 above.

hrabe:
>In fact... what about something like this:
>```
>#if defined(_MSC_VER) && (_MSC_VER < 1500)
>#undef SECTION
>#define SECTION(name) INTERNAL_CATCH_SECTION(name, "dummy-for-vc8")
>#endif
>```
>Shouldn't this solve the problem for it?

This would solve it but only as long as nobody uses descriptions, if so, this workaround will break.

hrabe:
Now Travis complains about wrong CMake version.
```
CMake Error at CMakeLists.txt:13 (cmake_minimum_required):
  CMake 2.8.10 or higher is required.  You are running version 2.8.7
```

mloskot:
> Now Travis complains about wrong CMake version.

This is due to Travis timeout'ed during apt steps and did not pull CMake update from PPA repository.
The failure is also accompanied with `Could not connect to ppa.launchpad.net:80 (91.189.95.83), connection timed out` or similar (like here https://travis-ci.org/SOCI/soci/jobs/276502301).
Nudging Travis build usually helps.

I have manually restarted the three failing build jobs.

mloskot:
Travis is green.

AppVeyor is also green https://ci.appveyor.com/project/mloskot/soci/build/4.0.0.288 despite the "Waiting for status to be reported". I'll hit "Update Branch" button to rebase from master and nudge the rebuild.

Once builds complete, it's ready to merge. @vadz you agree?

mloskot:
Holy s***t! I had no idea the "Update branch" button messes the commit history a bit:
with "Merge branch 'master' into master" commit 4a4b9cbf3933a2e76c1f38e1e9d3181fc0745187
https://ci.appveyor.com/project/mloskot/soci/build/4.0.0.289

I'm not gonna use this button any more:)

@hrabe I'm sorry, feel free to `git push -f` to update your PR as you wish.

hrabe:
PR updated. Hope will now be valid for merge after travis and appveyor.

Small question beside: If I work on a branch at my fork, could I also use travis and appveyor at forked branch?

mloskot:
@hrabe 
> Small question beside: If I work on a branch at my fork, could I also use travis and appveyor at forked branch?

Yes, of course. I do it myself. Just enable the the services for your fork.

CI builds don't send any notifications to mailing list, so there won't be any unwanted noise. 
CI builds send notifications to https://gitter.im/SOCI room though, but that is fine.

Good practices is also to develop contributions in topic branch, not in master branch of your fork. Then, you submit PR from on that topic branch.

hrabe:
Thanks for confirmation and suggestion. Will do next changes in feature branches beside master and will use a _forked_master_ to sum up all in fork too.

mloskot:
@hrabe I wouldn't worry about using `forked_master`, just keep `master` on your github.com/hrabe/soci plus topic branches. No need to shadow the real master, unless you want/need yourself.

mloskot:
FYI, I've decide to try out the GitHub's [Squash and merge](https://help.github.com/articles/merging-a-pull-request/) button (second of the three buttons). I think it worked fine, no damage noticed.

mloskot:
@hrabe Thanks for your efforts!

PR #594, opened 2017-09-23
[CMake] Add SOCI_POSTGRESQL_NOSINLGEROWMODE option (default OFF)

mloskot:
By default, retrieving query results row-by-row is enabled.

CMake tries to detect version of PostgreSQL (libpq) used to build SOCI
and if detected version is less than 9.x.x, then the option is
flipped to ON for backward compatibility.

Complements #571

### DIFF
--- a/docs/backends/postgresql.md
+++ b/docs/backends/postgresql.md
@@ -168,3 +168,4 @@ To support older PostgreSQL versions, the following configuration macros are rec
 * `SOCI_POSTGRESQL_NOBINDBYNAME` - switches off the query rewriting.
 * `SOCI_POSTGRESQL_NOPARAMS` - disables support for parameterized queries (binding of use elements),automatically imposes also the `SOCI_POSTGRESQL_NOBINDBYNAME` macro. It is necessary for PostgreSQL 7.3.
 * `SOCI_POSTGRESQL_NOPREPARE` - disables support for separate query preparation, which in this backend is significant only in terms of optimization. It is necessary for PostgreSQL 7.3 and 7.4.
+* `SOCI_POSTGRESQL_NOSINLGEROWMODE` - disable single mode retrieving query results row-by-row. It is necessary for PostgreSQL prior to version 9.
--- a/src/backends/postgresql/CMakeLists.txt
+++ b/src/backends/postgresql/CMakeLists.txt
@@ -11,6 +11,10 @@
 
 include(CMakeDependentOption)
 
+option(SOCI_POSTGRESQL_NOSINLGEROWMODE
+  "Do not use single row mode. PostgreSQL <9 portability."
+  OFF)
+
 option(SOCI_POSTGRESQL_NOPARAMS
   "Do not use input parameters. PostgreSQL 7.x portability."
   OFF)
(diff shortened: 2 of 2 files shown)


docs/backends/postgresql.md
```diff
@@ -166,5 +166,7 @@ format of UUID on output. See the test `test_uuid_column_type_support` for usage
 To support older PostgreSQL versions, the following configuration macros are recognized:
 
 * `SOCI_POSTGRESQL_NOBINDBYNAME` - switches off the query rewriting.
-* `SOCI_POSTGRESQL_NOPARAMS` - disables support for parameterized queries (binding of use elements),automatically imposes also the `SOCI_POSTGRESQL_NOBINDBYNAME` macro. It is necessary for PostgreSQL 7.3.
+* `SOCI_POSTGRESQL_NOPARAMS` - disables support for parameterized queries (binding of use elements),automatically imposes also the
+* `SOCI_POSTGRESQL_NOBINDBYNAME` macro. It is necessary for PostgreSQL 7.3.
```
alexott: The description of macro `SOCI_POSTGRESQL_NOBINDBYNAME` is duplicated

docs/backends/postgresql.md
```diff
@@ -166,5 +166,7 @@ format of UUID on output. See the test `test_uuid_column_type_support` for usage
...
 * `SOCI_POSTGRESQL_NOPREPARE` - disables support for separate query preparation, which in this backend is significant only in terms of optimization. It is necessary for PostgreSQL 7.3 and 7.4.
+* `SOCI_POSTGRESQL_NOSINLGEROWMODE` - disable single mode retrieving query results row-by-row. It is necessary for PostgreSQL prior version 9.
```
vadz: Tiny remark: "prior _to_ version 9" would seem to be better wording.

docs/backends/postgresql.md
mloskot: Good catch. I created new bullet by accident.

docs/backends/postgresql.md
mloskot: Thanks. Fixed.

mloskot:
@vadz
> I hoped that CMake would be able to set this automatically depending on the version of PostgreSQL detected. 

The CMake workflow for this new option boils down to these two steps:

```
option(SOCI_POSTGRESQL_NOSINLGEROWMODE
  "Do not use single row mode. PostgreSQL <9 portability."
  OFF)

if (POSTGRESQL_VERSION VERSION_LESS "9.0.0")
  set(SOCI_POSTGRESQL_NOSINLGEROWMODE ON CACHE BOOL "Use single row mode for PostgreSQL 9+" FORCE)
endif()
```

1. `NOSINLGEROWMODE` is disabled by default, we assume use of fairly modern PostgreSQL.

2. PostgreSQL version is checked, just in case, and `NOSINLGEROWMODE` is disabled if old PostgreSQL is detected.

Now, there is slight chance CMake will fail to detect the version then the default set in 1. is not affected.
In such case user will have to help CMake, if she needs, by going explicit with `-DSOCI_POSTGRESQL_NOSINLGEROWMODE=ON`.

Does that answer your concern about the 'automatically'?

vadz:
Oops, sorry, I've somehow missed the comparison of the version with 9.0 (I blame horrible CMake syntax but, to be totally candid, it is, of course, entirely my own fault). Sorry for the stupid question and thanks for the explanation!

mloskot:
@vadz No problem at all.

(I still wait for a better build configuration for SOCI ;-))

PR #596, opened 2017-09-24
Add SOCI_OVERRIDE macro as conditional C++11 override specifier

mloskot:
MSVC++ 1900+ always compile with C++11 mode enabled, so it should be safe to selectively enable 'override'specifier for internal use.

It does not enable all C++11 features for SOCI and we still compile  with C++11 compilation mode SOCI_CXX_C11=OFF by default.

---

# Refactoring Procedure with clang-tidy

* Install clang-tidy as new as you can easily get

```
$ clang-tidy-4.0 --version
LLVM (http://llvm.org/):
  LLVM version 4.0.0

  Optimized build.
  Default target: x86_64-pc-linux-gnu
  Host CPU: haswell
```

* Generate compile command database (`compile_commands.json`) with CMake (sweet, isn't it? :-))

```
$ cd soci
$ mkdir build
$ cmake -DCMAKE_EXPORT_COMPILE_COMMANDS=ON -DSOCI_CXX_C11=ON .. 

# All backends configured!
```

* Run `clang-tidy` with `modernize-use-override` check on all translations units and actually refactor all the code (`-fix`)

```
$ /usr/lib/llvm-4.0/share/clang/run-clang-tidy.py -clang-tidy-binary \
       -header-filter='.*' -checks='-*,modernize-use-modernize' -fix
```

Here we use [run-clang-tidy.py](https://github.com/llvm-mirror/clang-tools-extra/blob/master/clang-tidy/tool/run-clang-tidy.py)
utility script and run  inside the directory with `compile_commands.json` database.
The script is recommended way to run `clang-tidy`, it runs concurrent executions, etc,

* Finally, use your favourite tool to run case-sensitive whole-word search for `override` and replace with `SOCI_OVERRIDE`.

Voilà!

### DIFF
--- a/include/soci/empty/soci-empty.h
+++ b/include/soci/empty/soci-empty.h
@@ -39,12 +39,12 @@ struct SOCI_EMPTY_DECL empty_standard_into_type_backend : details::standard_into
         : statement_(st)
     {}
 
-    void define_by_pos(int& position, void* data, details::exchange_type type);
+    void define_by_pos(int& position, void* data, details::exchange_type type) SOCI_OVERRIDE;
 
-    void pre_fetch();
-    void post_fetch(bool gotData, bool calledFromFetch, indicator* ind);
+    void pre_fetch() SOCI_OVERRIDE;
+    void post_fetch(bool gotData, bool calledFromFetch, indicator* ind) SOCI_OVERRIDE;
 
-    void clean_up();
+    void clean_up() SOCI_OVERRIDE;
 
     empty_statement_backend& statement_;
 };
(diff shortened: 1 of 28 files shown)


include/soci/soci-platform.h
```diff
@@ -95,6 +95,12 @@ namespace std {
 // mode, we just need to check for the minimal compiler version supporting them
 // (see https://msdn.microsoft.com/en-us/library/hh567368.aspx).
 
+#if defined(SOCI_HAVE_CXX_C11) || (defined(_MSC_VER) && _MSC_VER >= 1800)
```
vadz: Again, hardly important, but it is actually supposed to work with 1700 (VC11) too.

include/soci/empty/soci-empty.h
```diff
@@ -135,20 +135,20 @@ struct empty_rowid_backend : details::rowid_backend
 {
     empty_rowid_backend(empty_session_backend &session);
 
-    ~empty_rowid_backend();
+    ~empty_rowid_backend() SOCI_OVERRIDE;
```
vadz: FWIW I'm personally for using `override` on dtors too but this is recommended _against_ by C++ core guidelines, see [this closed issue](https://github.com/isocpp/CppCoreGuidelines/issues/721).

mloskot: Interesting reading. I've started using the CppCoreGuidelines, but not relying on it strictly due to the performance issues (eg. with `gsl::span`).

Meanwhile, I agree with your comment in that thread.

For now, I'll stick `SOCI_OVERRIDE` in destructors. After we move to C++11 only and we ever start using the CppCoreGuidelines or discuss/agree to do otherwise. we can always remove it.

mloskot:
I've run `override` refactoring across the whole codebase and updated the PR.

include/soci/soci-platform.h
mloskot: Yes, it should even, partially or perhaps work with MSVC 1600 as per [Support For C++11/14/17 Features](https://msdn.microsoft.com/en-us/library/hh567368.aspx).

PR #602, opened 2017-09-30
Fix ODBC backend postgres tester

hrabe:
Running the ODBC tester for postgres either with no given connect string (command line or file DSN) or valid connect string but not existing database will crash the tester with **abnormal program termination** instead of failed test protocol and correct reason for fail.

This is due to using the soci session already in driver version detection in constructor of `test_backend` which will throw an `soci_error`. Unfortunately it is already under control of _Catch_ framework and handles this (at least at Windows Environment) by continue by the next code line after the place the exception was thrown. This now accesses the uninitialized return value of driver connect call for the real connect string. In final consequence this now throws a runtime exception (bad memory) which terminates the program abnormal.

BTW: It is **not** a good idea to call methods potentially raising exceptions in constructors (especially initializers).

This has been fixed using lazy singleton initialization of driver version if required.


vadz:
Something strange has happened with your branch history but, more importantly, I think the fix in 61d3127 is too complicated and I'd rather apply a much simpler #603 instead.

Do you have any objections to it?

BTW, I completely disagree that it's not a good idea to throw from ctors. It's, in fact, almost the only place where it's all but required. The problem here is not throwing from ctor, but rather not catching unhandled exceptions in `main()`. Personally I think there should be a `try/catch` around the entire main body in all the tests.

hrabe:
@vadz Don't see it too complicated, it's just a matter of lazy init when needed the first time.
Related to ctor initializers and thrown exceptions: those classes can never be instantiated using 
`test_context* p_context = new test_context(...);` cause this would leak the memory even if try/catched. 

I'm fine, if this fix is not expected to solve but try/catch within driver version #603 does. Feel free to close this.

vadz:
Well, I think lazy init is not very useful for something that we know will be used in all normal circumstances anyhow (because we show it in the ctor).

As for `foo* f = new foo()`, this is exactly the reason why raw pointers should never be used in C++ code and `std::unique_ptr<>` (or `auto_ptr<>` in our pre-C++11 version) must be used instead.

Anyhow, I'll close this one and apply #603, thanks.

Issue #605, opened 2017-10-18
build fails on macOS 10.12 sierra clang 9.0.0


rocksolidwebdesign:
errors preventing successful build on macOS sierra occur when building from latest master, which as of this writing is: d0552e64abc2be698aaf3007a5fd5e5138f0c830

errors occur in postgresql tests and mysql tests. errors block successful library compilation and install.

errors occur both with and without `-stdlib=libc++`

build succeeds on Ubuntu 16.04 with clang++ 3.8.0 (and also g++ 5.4.0)

build succeeds on Windows 8.1 with VS2015

though I'm not sure the primary cause of the failure, I noticed while investigating some of the messages that:

comparing https://github.com/SOCI/soci/blob/master/tests/mysql/test-mysql.h#L106

to https://github.com/SOCI/soci/blob/master/tests/mysql/test-mysql.h#L116

one uses a namespace qualifier, the other does not i.e. `soci::session` vs `session`

steps to reproduce and resulting output:

```
$ cd ~/Downloads
$ git clone https://github.com/SOCI/soci.git
$ mkdir soci_build
$ cd soci_build
$ cmake -G "Unix Makefiles" -DSOCI_SHARED=ON -DSOCI_LIBDIR=lib -DSOCI_STATIC=OFF -DCMAKE_PREFIX_PATH=~/local -DCMAKE_BUILD_TYPE=Debug -DCMAKE_INSTALL_PREFIX=/Users/me/local -DCMAKE_C_COMPILER=clang -DCMAKE_CXX_COMPILER=clang++ -DCMAKE_CXX_FLAGS=-stdlib=libc++ -DCMAKE_EXE_LINKER_FLAGS=-stdlib=libc++ /Users/me/Downloads/soci
-- The C compiler identification is AppleClang 9.0.0.9000037
-- The CXX compiler identification is AppleClang 9.0.0.9000037
-- Check for working C compiler: /usr/bin/clang
-- Check for working C compiler: /usr/bin/clang -- works
-- Detecting C compiler ABI info
-- Detecting C compiler ABI info - done
-- Detecting C compile features
-- Detecting C compile features - done
-- Check for working CXX compiler: /usr/bin/clang++
-- Check for working CXX compiler: /usr/bin/clang++ -- works
-- Detecting CXX compiler ABI info
-- Detecting CXX compiler ABI info - done
-- Detecting CXX compile features
-- Detecting CXX compile features - done
-- Looking for __i386__
-- Looking for __i386__ - not found
[...]

PR #607, opened 2017-10-21
fix build on macosx sierra

rocksolidwebdesign:
use fully qualified `soci::` namespace prefix for occurrences of `soci::session` to avoid ambiguity with `proc.h` on darwin

fix for #605


mloskot:
Travis CI build job for Oracle is failing for unrelated reasons, so I accept this PR as sound.

Issue #609, opened 2017-10-29
Uninitialized gotData_ in soci::statement


jonathanmcdougall:
Similar to [this one for `procedure`](https://github.com/SOCI/soci/pull/535), `statement::gotData_` is not initialized in constructors nor is it copied in copy constructors or by `operator=`.

    soci::statement f()
    {
        soci::statement s = ...
        assert(!s.got_data());    // passes
        return s;
    }

    auto s = f();
    assert(!s.got_data());        // can fail

mloskot:
PR is welcome

mloskot:
Fixed in #615

PR #615, opened 2018-01-04
Fixing uninitialized gotData_ in soci::statement #609

AngelinNjakasoa:
This fixes the uninitialized gotData_ in constructors and copy assignment operator of the class
soci::statement.

### DIFF
--- a/include/soci/statement.h
+++ b/include/soci/statement.h
@@ -172,9 +172,9 @@ class SOCI_DECL statement
 {
 public:
     statement(session & s)
-        : impl_(new details::statement_impl(s)) {}
+        : impl_(new details::statement_impl(s)), gotData_(false) {}
     statement(details::prepare_temp_type const & prep)
-        : impl_(new details::statement_impl(prep)) {}
+        : impl_(new details::statement_impl(prep)), gotData_(false) {}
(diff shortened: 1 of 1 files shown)


vadz:
Thanks for the changes, but I'm not quite sure what the actual semantics of `gotData_` is, i.e. do we really expect `got_data()` to return true on a new copy of the statement? If so, I wonder why isn't `gotData_` a member of `statement_impl` rather than `statement` itself in the first place?

Anyhow, I don't see any problem with applying this, but `statement` copy ctor should initialize it too then, presumably. And if we can move it to `statement_impl`, this would arguably be even better.

mloskot:
@AngelinNjakasoa would you be willing to refine your PR as per @vadz comment?
Then it will have a chance to make it into 4.0.0 release - I'm not able myself to work on updating your PR now.

mloskot:
Apparently OP has got lost in action and to move forward I'm going to accept this PR as is.
Next, I will add the missing initialisation to the `statement` copy ctor.

Rest of the suggestions of improvements will have to wait.

mloskot:
B***dy hell I should have squashed the commits. I completely forgot PR update via GitHub generates those silly `Merge branch 'master' into master`. Sorry!

Issue #617, opened 2018-01-09
CMAKE_BUILD_TYPE=RelWithDebInfo build fails with "may be used uninitialized" error


Alex-Vol-SV:
When building the library on Ubuntu 14.04 LTS and using RelWithDebInfo the build fails with this error:

```
Scanning dependencies of target soci_postgresql_test
[ 97%] Building CXX object tests/postgresql/CMakeFiles/soci_postgresql_test.dir/test-postgresql.cpp.o
In file included from /home/avolanis/stash/soci-nar/soci/build/src/include/soci/column-info.h:13:0,
                 from /home/avolanis/stash/soci-nar/soci/build/src/include/soci/soci.h:16,
                 from /home/avolanis/stash/soci-nar/soci/build/src/tests/postgresql/test-postgresql.cpp:8:
/home/avolanis/stash/soci-nar/soci/build/src/include/soci/values.h: In member function ‘void soci::values::set(const string&, const T&, soci::indicator) [with T = int, std::string = std::basic_string<char>]’:
/home/avolanis/stash/soci-nar/soci/build/src/include/soci/values.h:37:40: error: ‘baseValue’ may be used uninitialized in this function [-Werror=uninitialized]
/home/avolanis/stash/soci-nar/soci/build/src/include/soci/values.h:176:23: note: ‘baseValue’ was declared here
cc1plus: all warnings being treated as errors
make[2]: *** [tests/postgresql/CMakeFiles/soci_postgresql_test.dir/test-postgresql.cpp.o] Error 1
make[1]: *** [tests/postgresql/CMakeFiles/soci_postgresql_test.dir/all] Error 2
make: *** [all] Error 2
```

Applying this simple change to the code in values.h will correct the problem.

```
$ git diff
diff --git a/include/soci/values.h b/include/soci/values.h
index f286857..72c3db4 100644
--- a/include/soci/values.h
+++ b/include/soci/values.h
@@ -174,10 +174,7 @@ public:
             indicators_.push_back(pind);
 
             base_type baseValue;
-            if (indic == i_ok)
-            {
-                type_conversion<T>::to_base(value, baseValue, *pind);
-            }
+            type_conversion<T>::to_base(value, baseValue, *pind);
 
             details::copy_holder<base_type> * pcopy =
                     new details::copy_holder<base_type>(baseValue);
```

[...]

PR #618, opened 2018-01-09
Issue #617: CMAKE_BUILD_TYPE=RelWithDebInfo build fails with "may be …

Alex-Vol-SV:
…used uninitialized" error

This simple change eliminates the problem and unit tests
continue to pass as before.

### DIFF
--- a/include/soci/values.h
+++ b/include/soci/values.h
@@ -166,7 +166,7 @@ public:
     void set(std::string const & name, T const & value, indicator indic = i_ok)
     {
         typedef typename type_conversion<T>::base_type base_type;
-        if(index_.find(name) == index_.end())
+        if (index_.find(name) == index_.end())
         {
             index_.insert(std::make_pair(name, uses_.size()));
 
@@ -174,10 +174,7 @@ public:
             indicators_.push_back(pind);
 
             base_type baseValue;
-            if (indic == i_ok)
-            {
-                type_conversion<T>::to_base(value, baseValue, *pind);
-            }
+            type_conversion<T>::to_base(value, baseValue, *pind);
 
             details::copy_holder<base_type> * pcopy =
                     new details::copy_holder<base_type>(baseValue);


vadz:
Sorry but I don't think this is the right thing to do, the conversion really shouldn't be performed if there is no value (i.e. indicator is not "ok"). The value almost certainly should be default-constructed in this case.

Alex-Vol:
For the case of set the only values that would make sense for indicator are i_ok and i_null. Depending on the template expansion the "value" may be considered uninitialized by the default construction so in this case the problem is intrinsic to how the default construction works.

IMHO the caller would have had to pass some "sane" value, either default constructed or initialized for the case of i_null so I do not agree that allowing this assumption to hold has no merit.

On the other hand, should this code even be attempting to perform the conversion if i_null is the desired value? And what if anything should happen in this case?

Furthermore, the alternate usage of set a few lines below unconditionally performs the conversion regardless of the indicator. So, which is the correct code? Are both these implementations of set incorrect for indicator != i_ok?

Alex-Vol:
I rebased this against the latest master. Do you have any feedback or alternatives to consider for my implementation? Does the comment I made on Jan 10th make sense?

vadz:
Sorry, I've actually never used `values` in my own code, so I just don't know how is it supposed to work. One thing is clear: the inconsistency between the 2 `set()` overloads, with one checking for `indic == i_ok` and the other one not doing it, must be resolved. If you think that performing the conversion is harmless in the `i_null` case, I don't object to doing it, even though it still _seems_ to me that it doesn't make sense to convert if there is no value (i.e. it's a NOP in the best case). But I could well be missing something...

Alex-Vol-SV:
You have a point regarding the `i_null` conversion. In the extreme case where the type is null-able the reference may be an actual null that cannot be copied into the copy_holder. Perhaps the right thing to do is allow for the `deepCopies_` and `uses_` collections to receive a `null` entry respectively with the understanding that the rest of the code obeys the `i_null` and does not attempt to reference the values in any way.

The else part of the branch logic here will leave the existing `deepCopies_` and `uses_` unmodified if `indic == i_null` which is also weird but makes sense if the deeper layers simply ignore any values there.

Alex-Vol:
After some rather pointless and frustrating attempt to force this to avoid the conversion when `indic == i_null` I came to the conclusion that the behavior in the second form of `set` is correct. The signature of `use` dictates that there must exist an actual instance even one that is default constructed in order to perform a proper call. As proof I submit your own documentation with examples of the uses for indicator `i_null`.

[Bulk insert with NULL](https://github.com/SOCI/soci/blob/master/docs/indicators.md#bulk-operations-with-null-values)

Note that the insert NULL batch operation is still forced to use a vector of dummy values that are obviously never actually inserted in the table since all the indicators provided are `i_null`. The same can be seen in the single row example above.

I am now convinced that the correct implementation is to unconditionally convert the required `T const& value` argument and process it regardless of the value of indic. It is worth noting that the use_type already carries the indicator as a reference and can perform appropriate conditional processing where needed.

I am going to correct this PR to reflect this decision and we can discuss further.

vadz:
OK, let's do it like this, there is no real harm in doing the conversion unnecessarily anyhow, I guess.

Thanks for taking time to investigate this!

Alex-Vol:
The build failed one stage because of a timeout in before_script. Not an issue with my PR. Prior build had passed. Not sure if Travis can just rebuild this without a change.

mloskot:
@Alex-Vol I've just restarted the failing build job.

Alex-Vol-SV:
I rebased against latest master to prepare it for merging. @mloskot would it be possible to get this merged since @vadz and I have agreed it is the right solution?

mloskot:
The Mingw build failure on AppVeyor seems unrelated to your changes. I'll work on those later.

Thank you for your contribution.

mloskot:
I'm travelling now and apparently GitHub does not allow me to merge this on mobile.

@vadz could you merge it, please? If not I'll do it tonight or tomorrow

vadz:
Done, but I don't have time to look into MinGW issue right now. It looks like it uses MSVC headers instead of MinGW ones, but I don't know why.

mloskot:
@vadz Thanks. No worries. I'm not a MinGW user myself, and I'd happily ignore it, but I will try to look at it when I'm back home by the weekend.

PR #622, opened 2018-02-22
Fix typo in SOCI_POSTGRESQL_NOSINLGEROWMODE option/macro name

mloskot:
How this has slipped through a few pair of reviewing eyes ;-)

Corrects changes added in #594


vadz:
I had to stare at the diff for something like 30 seconds to see the change even now :-/ Thanks for fixing it (and I still have no idea how did you manage to notice it unprompted)!

mloskot:
@vadz I just spotted it by a complete flux - I opened `CMakeLists.txt` and was about to remove `SOCI_POSTGRESQL_NOPARAMS` and somehow looked a that crooked option a couple of lines higher in the code :)

PR #624, opened 2018-02-23
Add statement_timeout to PostgreSQL connection parametres.

### LINKED ISSUE #129: Add timeout support for postgresql
this is just not to forget...

we can implement timeouts in postgresql backend by issuing the `SET SESSION statement_timeout TO NNN` command right after opening a connection to database. `NNN` is in milliseconds.

When getting corresponding error back, we can wrap it into separate exception.

P.S. I think, that this is possible to implement timeouts not only for postgesql backend....

mloskot:
Closes #129

### DIFF
--- a/docs/backends/postgresql.md
+++ b/docs/backends/postgresql.md
@@ -51,7 +51,8 @@ The set of parameters used in the connection string for PostgreSQL is the same a
 
 In addition to standard PostgreSQL connection parameters, the following can be set:
 
-* `singlerow` or `singlerows`
+* `singlerow` or `singlerows` - to control single-row mode
+* `statement_timeout` - to set override statement default timeout (milliseconds)
 
 For example:
 
--- a/src/backends/postgresql/session.cpp
+++ b/src/backends/postgresql/session.cpp
@@ -39,11 +39,13 @@ void hard_exec(postgresql_session_backend & session_backend,
 } // namespace unnamed
 
 postgresql_session_backend::postgresql_session_backend(
-    connection_parameters const& parameters, bool single_row_mode)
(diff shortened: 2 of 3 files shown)


src/backends/postgresql/session.cpp
```diff
@@ -74,6 +76,13 @@ void postgresql_session_backend::connect(
                          : "SET extra_float_digits = 2",
         "Cannot set extra_float_digits parameter");
 
+    {
+        std::ostringstream oss;
+        oss << "SET staetment_timeout = " << statement_timeout_;
```
Mingun: Misprint: `staetment_timeout` instead of `statement_timeout` (and also in line 83)

mloskot: @Mingun Thank you for the review. AT, Having no local dev environment for SOCI, I'm completely relying on the CI builds. I'll fix it.

docs/backends/postgresql.md
```diff
@@ -51,7 +51,8 @@ The set of parameters used in the connection string for PostgreSQL is the same a
 
 In addition to standard PostgreSQL connection parameters, the following can be set:
 
-* `singlerow` or `singlerows`
+* `singlerow` or `singlerows` - to control single-row mode
+* `statement_timeout` - to set override statement default timeout (milliseconds)
```
vadz: Either "set" or "override" seems to be superfluous here.

mloskot:
@vadz `sql << "SET statement_timeout = whatever";` makes perfect sense. I agree, it's hard to justify the code clutter when one can achieve the goal via the existing SQL interface. 

I'm dropping the idea.

PR #629, opened 2018-03-21
Better handling of cancelling ODBC connection parameters dialog

vadz:
I've recently noticed that cancelling the dialog shown by ODBC when the username and/or password is not specified resulted in a weird `bad_alloc` being thrown. Turns out we were using an unitialized variable, which can't be a good thing -- clarity of the error message notwithstanding.


mloskot:
LGTM

PR #630, opened 2018-03-29
Deadlock caused by infinite loop in try_lease resolved

reimerdes:
If the mutex is locked until the timeout of pthread_cond_timedwait is triggered EINVAL will be returned.
EINVAL was not checked which resulted in infinite loop.

### DIFF
--- a/src/core/connection-pool.cpp
+++ b/src/core/connection-pool.cpp
@@ -127,6 +127,12 @@ bool connection_pool::try_lease(std::size_t & pos, int timeout)
         {
             break;
         }
+        // pthread_cond_timedwait can return EINVAL or EPERM
+        if (cc == EINVAL || cc == EPERM)
+        {
+            // Without this break the while loop can result in an infinate loop
+            break;
+        }
     }
 
     if (cc == 0)


vadz:
Sorry, how exactly can `EINVAL` happen here? According to [the documentation](http://pubs.opengroup.org/onlinepubs/009695299/functions/pthread_cond_timedwait.html), it should only be the case if `tm` is invalid, but it ought to be valid here.

FWIW I agree that it would make sense to throw on any error different from `ETIMEDOUT` (giving the value of the error in the error message to help with debugging) as a defensive programming measure, but I also think that this must not happen at all currently.

reimerdes:
Took us a while to figure it out. We had issues were connection pool threads were going to infinit loop. We were able to reproduce the issue using 30 clients simultainusly requesting and a connection pool size of 1. `EINVAL` was occuring.

We had to prevent the occuring infinit loops.

Maybe I can dig into the code but that is a different timing issue.

vadz:
I'm finally going to apply this just because it shouldn't make things worse, even though I still find it very suspicious that POSIX function returns something it is not supposed to. I also wonder if we shouldn't throw an exception in this case, as it's definitely not normal for this to happen.

Thanks and sorry for the delay!

PR #639, opened 2018-05-19
Fix DDL creation of SQLite3 table with date/time column.

jmfermun:
When user-defined C++ types are used, an exception of type std::bad_cast in type_conversion::from_base function is obtained when a std::tm type is used.

If the row type is declared as DATETIME, the error disappears.

I have only tested the case with SQLite3 backend. The version of SQLite used is  3.22.0.

I have attached a sample code which ilustrates the problem:
[sample-sqlite-date.cpp.txt](https://github.com/SOCI/soci/files/2019415/sample-sqlite-date.cpp.txt)

### DIFF
--- a/cmake/modules/FindODBC.cmake
+++ b/cmake/modules/FindODBC.cmake
@@ -1,65 +1,242 @@
-#
-# Find the ODBC driver manager includes and library.
-#
-# ODBC is an open standard for connecting to different databases in a
-# semi-vendor-independent fashion.  First you install the ODBC driver
-# manager.  Then you need a driver for each separate database you want
-# to connect to (unless a generic one works).  VTK includes neither
-# the driver manager nor the vendor-specific drivers: you have to find
-# those yourself.
-#
-# This module defines
-# ODBC_INCLUDE_DIR, where to find sql.h
-# ODBC_LIBRARIES, the libraries to link against to use ODBC
-# ODBC_FOUND.  If false, you cannot build anything that requires MySQL.
+# Distributed under the OSI-approved BSD 3-Clause License.  See accompanying
+# file Copyright.txt or https://cmake.org/licensing for details.
 
-# also defined, but not for general use is
-# ODBC_LIBRARY, where to find the ODBC driver manager library.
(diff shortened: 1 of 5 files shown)


vadz:
According to [SQLite docs](https://www.sqlite.org/datatype3.html) there is no `datetime` in it and using it just specifies "numeric" affinity for the column, so perhaps it would be better/more clear to use `numeric` directly here?

It would also be very nice to add a unit test which fails now and is fixed by applying this PR.

mloskot:
Thank you for your contribution indeed. I'm afraid it won't be accepted without a test case targeting the issue you discovered and verifying your fix.

jmfermun:
SQLite3 hasn't got built-in support for dates or times types. As you say, the SQLite docs specify that the types DATE, TIME and DATETIME has a NUMERIC type affinity. My first attempt was to use the NUMERIC type instead of INTEGER where I place the patch (file  include/soci/sqlite3/soci-sqlite3.h line 324), but the case failed too.

The SOCI backend for SQLite3 uses a hack to work with dt_date type. If you inspect the funtion get_data_type_map (file src/backends/sqlite3/statement.cpp line 409), it defines the possible types (the type asigned to a column when the table is created) that will be recognised as dt_date. They are "date", "time" and "datetime", so I suppose that any of the three fixes the problem. If you define the table column as NUMERIC, it will be recognised as dt_double.

I attached a sample code in my first message (file sample-sqlite-date.cpp.txt) that fails without the patch and works with it. Do you need to include it in the tests contained in the file tests/sqlite3/test-sqlite3.cpp?

vadz:
> I attached a sample code in my first message (file sample-sqlite-date.cpp.txt) that fails without the patch and works with it. Do you need to include it in the tests contained in the file tests/sqlite3/test-sqlite3.cpp?

Yes, please. This will not only make sure that the bug is fixed now, but also that it remains fixed. TIA!

vadz:
It looks like this has been superseded by #643, so closing -- please reopen if I misunderstood.

Issue #640, opened 2018-05-20
Build errors with gcc 8.1


amreisa:
System Linux manjaro

```
cmake ../ -G "Unix Makefiles" -DWITH_BOOST=ON -DSOCI_TESTS=ON -DCMAKE_INSTALL_PREFIX="home/amr/Dev/Libs/SOCI/Linux/" -DSOCI_SHARED=OFF -DSOCI_STATIC=ON -DCMAKE_BUILD_TYPE=RELEASE
-- The C compiler identification is GNU 8.1.0
-- The CXX compiler identification is GNU 8.1.0
-- Check for working C compiler: /bin/cc
-- Check for working C compiler: /bin/cc -- works
-- Detecting C compiler ABI info
-- Detecting C compiler ABI info - done
-- Detecting C compile features
-- Detecting C compile features - done
-- Check for working CXX compiler: /bin/c++
-- Check for working CXX compiler: /bin/c++ -- works
-- Detecting CXX compiler ABI info
-- Detecting CXX compiler ABI info - done
-- Detecting CXX compile features
-- Detecting CXX compile features - done
-- Looking for __i386__
-- Looking for __i386__ - not found
-- Looking for __x86_64__
-- Looking for __x86_64__ - found
-- Looking for __arm__
-- Looking for __arm__ - not found
-- Configuring SOCI: 
-- SOCI_VERSION                             = 4.0.0 
-- SOCI_ABI_VERSION                         = 4.0 
-- SOCI_SHARED                              = OFF 
-- SOCI_STATIC                              = ON 
-- SOCI_TESTS                               = ON 
-- SOCI_ASAN                                = OFF 
-- SOCI_CXX_C11                             = OFF 
-- LIB_SUFFIX                               = 64 
-- Looking for SOCI dependencies: 
-- Threads: 
-- CMAKE_THREAD_LIBS_INIT                   = -lpthread 
-- Boost: 
-- Boost_RELEASE_VERSION                    = 1.66.0 
-- Boost_INCLUDE_DIR                        = /usr/include 
-- Boost_LIBRARIES                          = /usr/lib/libboost_date_time.so 
-- MySQL: 
-- Performing Test HAVE_MYSQL_OPT_EMBEDDED_CONNECTION
-- Performing Test HAVE_MYSQL_OPT_EMBEDDED_CONNECTION - Failed
-- Found MySQL: /usr/include/mysql, /usr/lib/libmysqlclient.so
-- MySQL Embedded not found.
[...]

PR #641, opened 2018-05-20
Avoid g++-8 -Wcatch-value warnings in the tests

vadz:
Catch soci_error (and std::bad_cast in one place) by reference rather
than by value to avoid slicing -- even if it doesn't matter in the test
code, it still provokes the new g++-8 warning about it.

See #640.

### DIFF
--- a/tests/common-tests.h
+++ b/tests/common-tests.h
@@ -528,18 +528,18 @@ TEST_CASE_METHOD(common_tests, "Exception on not connected", "[core][exception]"
     soci::session sql; // no connection
 
     // ensure connection is checked, no crash occurs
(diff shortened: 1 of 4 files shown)


comorekf:
The fix in the commit [(fix-catch-polymorphic-by-value)](https://github.com/vadz/soci/commit/46e9763b330883af55c6b99e9aa335736531974e) works. However includes only odbc-mssql, firebird, and postgres. When I applied the same thing to the test-oracle.cpp it works fine.

vadz:
@comorekf Could you please make a PR fixing the remaining occurrences of this problem?

comorekf:
No problem

PR #643, opened 2018-05-31
Fix portable DDL query CREATE TABLE for SQLite3

jmfermun:
New attempt to patch the error described in pull request #639. This patch solves more related errors.

When the portable DDL query CREATE TABLE is used in conjunction with ORM, the use of the following data types cause an exception of type std::bad_cast in type_conversion::from_base method:
- dt_date
- dt_long_long
- dt_unsigned_long_long

The data type dt_xml isn't supported by the SQLite3 backend. The method now throws an exception when this type is used.

The method statement_backend::describe_column previously failed when a column type name had spaces, for example "unsigned big int". It has been fixed.

Added test "DDL data types" to the file common-tests.h to check the above errors.

### DIFF
--- a/include/soci/type-conversion-traits.h
+++ b/include/soci/type-conversion-traits.h
@@ -36,29 +36,6 @@ struct type_conversion
     }
 };
 
-// Most majority of database engines ignore unsigned integer related types.
-// This hack allows to trait these types as signed integers in the database.
-template<>
-struct type_conversion<unsigned long long>
-{
-    typedef long long base_type;
-
-    static void from_base(base_type const & in, indicator ind, unsigned long long & out)
-    {
-        if (ind == i_null)
-        {
-            throw soci_error("Null value not allowed for this type");
-        }
-        out = static_cast<unsigned long long>(in);
-    }
-
-    static void to_base(unsigned long long const & in, base_type & out, indicator & ind)
-    {
-        out = static_cast<long long>(in);
-        ind = i_ok;
-    }
-};
-
 } // namespace soci
 
 #endif // SOCI_TYPE_CONVERSION_TRAITS_H_INCLUDED
--- a/include/soci/type-conversion-traits.h
+++ b/include/soci/type-conversion-traits.h
@@ -36,6 +36,29 @@ struct type_conversion
     }
 };
 
+// Most majority of database engines ignore unsigned integer related types.
+// This hack allows to trait these types as signed integers in the database.
+template<>
+struct type_conversion<unsigned long long>
+{
+    typedef long long base_type;
+
+    static void from_base(base_type const & in, indicator ind, unsigned long long & out)
+    {
+        if (ind == i_null)
+        {
+            throw soci_error("Null value not allowed for this type");
+        }
+        out = static_cast<unsigned long long>(in);
+    }
+
+    static void to_base(unsigned long long const & in, base_type & out, indicator & ind)
+    {
+        out = static_cast<long long>(in);
+        ind = i_ok;
+    }
+};
+
 } // namespace soci
 
 #endif // SOCI_TYPE_CONVERSION_TRAITS_H_INCLUDED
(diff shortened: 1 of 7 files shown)


vadz:
Could you please look into CI failures? They seem like real problems, at least at the first glance. TIA!

mloskot:
@jmfermun Thanks for taking care of covering the issue/fix with the test. This is highly appreciated!

include/soci/type-conversion-traits.h
```diff
@@ -36,6 +36,29 @@ struct type_conversion
     }
 };
 
+// Most majority of database engines ignore unsigned integer related types.
```
vadz: Some typos: "most majority" is redundant, "trait" is probably supposed to be "handle".

include/soci/type-conversion-traits.h
```diff
@@ -36,6 +36,29 @@ struct type_conversion
...
+// This hack allows to trait these types as signed integers in the database.
+template<>
+struct type_conversion<unsigned long long>
```
vadz: More importantly, the comment above speaks about "types", but the specialization here is just for a single type. I'm not sure if the comment should be trusted and we should also have specializations for `unsigned short`, `unsigned [int]` and `unsigned long` too or if the comment should be updated?

include/soci/type-conversion-traits.h
```diff
@@ -36,6 +36,29 @@ struct type_conversion
...
+{
+    typedef long long base_type;
+
+    static void from_base(base_type const & in, indicator ind, unsigned long long & out)
+    {
+        if (ind == i_null)
+        {
+            throw soci_error("Null value not allowed for this type");
```
vadz: Is this really correct? A database column of `bigint` type could store nulls perfectly well...

mloskot: Yes, this changes is not acceptable for core.

mloskot:
@vadz 
> I'm not sure which problem exactly does it fix, 

Not clear to me either.

> AFAICS SQLite supports unsigned big int just as well (or badly) as bigint, i.e. both are just integers for it.

That is also my understanding that those are equivalent aliases for `INTEGER`, https://www.sqlite.org/datatype3.html Why to prefer `unsigned big int` over `bigint`, no idea.

jmfermun:
I will try to explain the changes done.

I have only used postgresql and sqlite3.

- The test case create the database columns with the method create_column_type.
- For postgresql the method is in file include/soci/soci-backend.h. In this method the types dt_long_long and dt_unsigned_long_long are created as bigint columns in the database.
- For sqlite3 the method is in file include/soci/sqlite3/soci-sqlite3.h. I have changed this method to create the types dt_long_long and dt_unsigned_long_long as bigint columns in the database.

- When ORM is used, the column types in SOCI are defined with the method describe_column.
- For postgresql the method is in file src/backends/postgresql/statement.cpp. In this method the columns created as bigint are detected as dt_long_long type in SOCI.
- For sqlite3 the method is in file src/backends/sqlite3/statement.cpp. I have changed it to detect the columns created as bigint, int8 and unsigned big int as dt_long_long, so the behavior is similar to postgresql.

In summanry:
- For the correct detection of types in the case of ORM, it is necesary that the types in the database are different for each SOCI type. For example, if the same column type (bigint) is used for dt_long_long and dt_unsigned_long_long we obtain a bad_cast exception.
- As most databases haven't got different types for signed integer types and unsigned integer types, I decided to treat dt_unsigned_long_long as dt_long_long in the method type_conversion<unsigned long long> to avoid the above problem. And handle it in the database side as a signed integer type.

davidsummers:
I'm running into this exact error myself.  I am trying to use dt_long_long in the ORM and getting the exception std::bad_cast.

Is there a fix for this yet?  How can I work around this?

Thanks.

vadz:
Unfortunately these changes break several other backends (DB/2, ODBC, Firebird and Oracle), so they can't be applied, sorry.

FWIW I also think that the mapping of C++ types to database types needs to be completely redone in SOCI 5, so more fundamental changes are needed here.

PR #645, opened 2018-06-04
Fixed memory leaks

atvise:
- Fixed memory leak due to public inheritance of std::vector. The standard containers do not have virtual destructors! -> https://stackoverflow.com/questions/6806173/subclass-inherit-standard-containers. This is a qick fix. It's better to extract the exchange function and redesign it to be able to handle any kind of STL like std::vector with any kind of data in it.
- Added clean_up call within destructor of standard_into_type and vector_into_type. The clean_up function will not be called every time (exceptions, missing implementation, etc.)


vadz:
Could you please explain where exactly is the memory leak? I don't see any places where we would destroy `use_type_vector` via a pointer to the base class.

atvise:
Hi @vadz,

currently i see no such place where you are destroing the use_type_vector or into_type_vector via a pointer to the base class. Nevertheless someone should consider in the future to refactor that with said external functionality because you never know what the future brings. Therefore i will remove those changes in my PR. 

The other change in the PR however (missing clean_up in destructur) is crucial and I've made memory leak observations with the tool UI for ETW (https://randomascii.wordpress.com/2015/04/14/uiforetw-windows-performance-made-easier/).

Regards,
atvise

mloskot:
I don't mind if we avoid inheriting publicly from `std::vector`.

Issue #648, opened 2018-06-07
build still fails on macOS due to session qualifier


hrabe:
This is related to already fixed https://github.com/SOCI/soci/issues/605
The problem persists on macOS due to session is already defined and soci doesn't use the full qualifier `soci:session`. 

This is a test build on macOS for ODBC only (using unixodbc).

**Proposal**: Do not use `session` without qualifier but use `soci:session` everywhere.

**Hint**: If proposal is valid for you, I would provide a PR too.

```
Heikos-MacBook-Pro-2:soci heikorabe$ cmake -G "Unix Makefiles" -DSOCI_SHARED=ON -DSOCI_STATIC=OFF -DWITH_BOOST=OFF -DSOCI_EMPTY=OFF -DWITH_DB2=OFF -DWITH_FIREBIRD=OFF -DWITH_MYSQL=OFF -DWITH_ORACLE=OFF -DWITH_POSTGRESQL=OFF -DWITH_SQLITE3=OFF -DWITH_ODBC=ON -DSOCI_TESTS=ON /Users/heikorabe/Github-private/soci
-- The C compiler identification is AppleClang 9.1.0.9020039
-- The CXX compiler identification is AppleClang 9.1.0.9020039
-- Check for working C compiler: /Applications/Xcode.app/Contents/Developer/Toolchains/XcodeDefault.xctoolchain/usr/bin/cc
-- Check for working C compiler: /Applications/Xcode.app/Contents/Developer/Toolchains/XcodeDefault.xctoolchain/usr/bin/cc -- works
-- Detecting C compiler ABI info
-- Detecting C compiler ABI info - done
-- Detecting C compile features
-- Detecting C compile features - done
-- Check for working CXX compiler: /Applications/Xcode.app/Contents/Developer/Toolchains/XcodeDefault.xctoolchain/usr/bin/c++
-- Check for working CXX compiler: /Applications/Xcode.app/Contents/Developer/Toolchains/XcodeDefault.xctoolchain/usr/bin/c++ -- works
-- Detecting CXX compiler ABI info
-- Detecting CXX compiler ABI info - done
-- Detecting CXX compile features
-- Detecting CXX compile features - done
-- Looking for __i386__
-- Looking for __i386__ - not found
-- Looking for __x86_64__
-- Looking for __x86_64__ - found
-- Looking for __arm__
-- Looking for __arm__ - not found
-- Configuring SOCI: 
-- SOCI_VERSION                             = 4.0.0 
[...]

Issue #93, opened 2013-03-06
Update CHANGES for 3.2.0


mloskot:
Update the CHANGES file with major fixes and new features in SOCI 3.2.0, extract release notes.

BTW, may switch to Markdown, nice example here [bootstrap/CHANGELOG](https://github.com/twitter/bootstrap/blob/master/CHANGELOG.md)

PR #649, opened 2018-06-07
Add explicit qualifier to mssql odbc testcase

hrabe:
Fixes https://github.com/SOCI/soci/issues/648

### DIFF
--- a/tests/odbc/test-odbc-mssql.cpp
+++ b/tests/odbc/test-odbc-mssql.cpp
@@ -22,11 +22,11 @@ backend_factory const &backEnd = *soci::factory_odbc();
 // MS SQL-specific tests
 TEST_CASE("MS SQL long string", "[odbc][mssql][long]")
 {
-    session sql(backEnd, connectString);
+    soci::session sql(backEnd, connectString);
 
     struct long_text_table_creator : public table_creator_base
     {
-        explicit long_text_table_creator(session& sql)
+        explicit long_text_table_creator(soci::session& sql)
             : table_creator_base(sql)
         {
             // Notice that 4000 is the maximal length of an nvarchar() column,


vadz:
Thanks for fixing this!

mloskot:
@vadz Please, set the *4.0.0* milestone for every PR you merge. At least for those you merge before we finally release 4.0.0. This is a small tick, but will help greatly to collect release notes. Thanks.

vadz:
Sorry, I didn't know/forgot about this. FWIW I don't think we should mention compilation fixes for the tests in the release notes anyhow because it's implicitly understood that tests should compile anyhow, but I'll do it for any other changes/fixes in the future.

mloskot:
@vadz 
> I don't think we should mention compilation fixes for the tests in the release notes anyhow

Indeed, and we won't.

I just simplify the procedure:
- milestone everything, no-brainer
- dump everything for a milestone
- sieving dump keeping the important bits only

Otherwise, I'd have to ask myself if I have perhaps missed milestone-ing something important.

vadz:
@mloskot I usually look at `git log --oneline the_last_release_tag..` when preparing the release notes. Maybe we could use some tags (e.g. `[change]` or something like this) in the commit messages to simplify this?

mloskot:
@vadz I used to do something similar, as outlined here https://github.com/SOCI/soci/issues/93#issuecomment-15434667

I'm not sure about a commit tag, easier to forget than single milestone click (which also works for issues about future fixes/features).

Ideally, if a PR merge is accompanied with entry in `CHANGES` file, then we have it covered and no need to spend hours filtering logs to compile release notes :)

vadz:
@mloskot This would work too, but for the projects which keep a running changelog like this, another problem is common: merge conflicts for any change touching the changelog (which will be most of them). For SOCI this might not be such a big deal though, as we don't get that many PRs... but it's annoying.

mloskot:
@vadz Good point. I see no golden solution. Feel free to suggest one, we can try.

PR #650, opened 2018-06-07
Remove use of std::unary_function

### LINKED ISSUE #646: does not compile with Visual Studio 2017 Toolset v141
query_transformation.h uses std::unary_function which got removed in C++17

mloskot:
Deprecated in C++11 and removed in C++17

Fixes #646


docs/queries.md
```diff
@@ -52,11 +52,11 @@ sql << "DELETE FROM item";
...
 
 ```cpp
-struct order : std::unary_function<std::string, std::string>
+struct order
 {
-    order(std::string const&amp; by) : by_(by) {}
+    order(std::string const& by) : by_(by) {}
 
-    result_type operator()(argument_type query) const
+    std::string operator()(std::string const query) const
```
vadz: Missing `&`?

mloskot: Indeed. Good catch, thx. Fixed.

PR #661, opened 2018-06-16
Avoid memory corruption at ODBC backend in vector_into

hrabe:
BUG: potentially a new memory block gots allocated inside **indHolderVec_**
but still the old (and now freed) memory is bound by **indHolders_** member.
In multithreaded apps this additionally may overwrite other threads
memory.

### DIFF
--- a/src/backends/odbc/vector-into-type.cpp
+++ b/src/backends/odbc/vector-into-type.cpp
@@ -351,6 +351,7 @@ void odbc_vector_into_type_backend::post_fetch(bool gotData, indicator *ind)
 void odbc_vector_into_type_backend::resize(std::size_t sz)
 {
     indHolderVec_.resize(sz);
+    indHolders_ = &indHolderVec_[0];
     switch (type_)
     {
     // simple cases


vadz:
Good find, thanks!

We could merge this as is, or we could add a helper `resize_indicators()` function that would do

    indHolderVec_.resize(size);
    indHolders_ = &indHolderVec_[0];

in order to emphasize that these 2 operations always need to be done together. Do you think it's worth it?

mloskot:
I like the idea of `resize_indicators`. Thanks @hrabe

hrabe:
After digging deeper in the call flow, the proposed fix for **indHolders_** is not sufficient or more precise the problem may be even bigger. 

Lets look at the end of `void odbc_vector_into_type_backend::define_by_pos(...)`

```
...
    SQLRETURN rc
        = SQLBindCol(statement_.hstmt_, static_cast<SQLUSMALLINT>(position++),
                odbcType_, static_cast<SQLPOINTER>(data), size, indHolders_);
...
```

Any char type based vector (string, big int, etc.) get's it's independent allocated memory at `buf_` mapped to `data` and binds this to column. Example:
```
    case x_stdstring:
        {
            odbcType_ = SQL_C_CHAR;
            std::vector<std::string> *v
                = static_cast<std::vector<std::string> *>(data);
            colSize_ = statement_.column_size(position) + 1;
            std::size_t bufSize = colSize_ * v->size();
            buf_ = new char[bufSize];

            prepare_indicators(v->size());

            size = static_cast<SQLINTEGER>(colSize_);
            data = buf_;
        }
```

Any other type based vector maps the internal buffer of given vector to `data` and binds now the vectors internal buffer pointer to column. Example:
```
    case x_double:
        {
            odbcType_ = SQL_C_DOUBLE;
            size = sizeof(double);
            std::vector<double> *vp = static_cast<std::vector<double> *>(data);
            std::vector<double> &v(*vp);
            prepare_indicators(v.size());
            data = &v[0];
        }
```

If vectors now get resized within `void odbc_vector_into_type_backend::resize(std::size_t sz)` and lets assume the resize (shrink) does use new memory block within vectors due to a memory manager can uses small and large block caches, the char type vectors bindings are unaffected but the non char vector bindings are now dangling bound pointers!

If we would assume, that shrinking a `std::vector<>` never changes the underlying memory block, this entire fix is not really necessary. If we can't guarantee that shrink changes the memory block, we have to fix the bound non char values anyway too (probably also use `buf_`)?

Prior to refactor this PR I would poll your thought @vadz @mloskot

mloskot:
I'm afraid I'm of little help here - I can't tell what the author's of the backend intention was.

hrabe:
Would guess, that this behavior should avoid copy of results for int/double/... types in the hope hat the affected tables do not contain so much string based columns where copy have to be made always.
So this looks like a speed optimization _hack_ if I would asked.

BTW: I've  a lot of experience with 3rd party memory managers linked to application like this one as example: http://www.microquill.com/smartheapsmp/index.html and they change any aspect of how and when memory blocks will be allocated/deallocated or rebound to different block caches (including new/delete ops).

vadz:
Sorry, I'm quite lost in this code... AFAICS in `x_stdstring` case, the vector of values is actually not used at all by `SQLBindCol()`, only `indHolders_` is. So I'm not sure why would resizing this vector (even assuming it does move its memory, because I don't think we can guarantee that this doesn't happen) invalidate anything?

I.e. which pointer exactly becomes dangling (other than `indHolders_`)? I think it'd be worth adding a test case exercising this code by resizing the vector (by enough to force its reallocation) and checking that it passes with ASAN/Valgrind. This should give detailed information about the problem in case it really happens.

Thanks again for looking at this!

hrabe:
@vadz Here the short version:
```
SQLBindCol(statement_.hstmt_, static_cast<SQLUSMALLINT>(position++),
                odbcType_, static_cast<SQLPOINTER>(data), size, indHolders_);
```

binding of **data** case (A)
```
    case x_double:
        {
...
            data = &v[0];
```

binding of **data** case (B)
```
    case x_stdstring:
        {
...
            data = buf_;
```

- **data** bound either on vectors buffer pointer (A) or own separate allocation (B)
- **indHolders_** always bound as vectors buffer pointer

Using memory managers like _Smartheap_ may replace the memory block in resize of vectors with another block also in case of shrink (lets say from 10000 elem vector to  20 elem vector, cause it's now better handled by a block from _small block cache_ than _large block cache_ of memory manager.

hrabe:
I think, we can close this PR if we ignore for now special memory managers linked against application using soci.

From http://www.cplusplus.com/reference/vector/vector/resize/

> **Iterator validity**
> In case the container shrinks, all iterators, pointers and references to elements that have not been removed remain valid after the resize and refer to the same elements they were referring to before the call.
> If the container expands, the end iterator is invalidated and, if it has to reallocate storage, all iterators, pointers and references related to this container are also invalidated.

mloskot:
@hrabe I don't think we can require or rely on any custom memory managers.
@vadz I'll leave it to you to decide if this should be closed.

Thanks for looking at this.

vadz:
Oops, yes, there is definitely a problem with the data pointer too, I've looked at `x_stdstring` case, but this is the one which is actually fine, while the others, such as `x_double` definitely are not.

@hrabe As you've already confirmed, reducing the vector size doesn't invalidate its iterators and so can't reallocate it. However I don't see why is this really relevant, we must think of increasing the vector size too here and this definitely can reallocate it, so the problem is definitely real. And this PR solves it partially, by avoiding it for the indicators vector, but not completely.

I think the only real solution is to call `SQLBindCol()` again when reallocating the vector changes the pointer to its first element. AFAICS from ODBC docs, doing this is perfectly fine (they call it "Rebinding columns") and we do need to do it.

If this is too complicated, we could also forbid resizing of the vector used with `into()`. In this case we should remove `resize()` from all the different classes defining it now.

hrabe:
@vadz Soci already prevents resize in terms of **grow** in multiple fetches (bulk read) as shown with the `soci_error(...)` above. The resize in terms of **shrink** is used for multiple fetches where the last roundtrip fetches a lower row count then the loops before. 

Example: If vector with 100 elements given and the result amount is 135 rows for the select
- the first fetch fills all 100 elements 
- the second fetch **shrinks** the vector to 35 elements
- the third fetch delivers **false** and does nothing cause no more rows to fetch.

So combined with the behavior description for shrink the pointers should stay stable.
Soci prevents grow by raising a error if between first and second fetch someone tries to grow the already bound vectors.

Thatswhy this only would create trouble, if shrink behavior of vector container changes (e.g. by an other memory manager or vector implementation). So this should no longer be a problem and the PR could be closed. (Would prefer a notice at code level why this works within _into_ at this way)

vadz:
Oops, yes, sorry, I keep looking over this too quickly and missing things, thanks for your explanations.

I finally understand what you mean and agree that this should be closed. We could indeed add a comment in `prepare_indicators()` explaining that `indHolders_` will remain valid because the vector can only shrink or something like that.

Sorry again for the confusion due to my misunderstanding.

PR #681, opened 2018-09-20
Really fix long strings with ODBC

vadz:
This also works with the old ODBC drivers.

See #383.


vadz:
Merging this because the CI failure is bogus, it's due to failing to download Oracle RPM.

PR #682, opened 2018-09-20
Allow Oracle Travis CI build to fail

vadz:
It does fail all the time currently due to not being able to download
the Oracle package any longer.

---

Just in case you don't have time to look into this in more details (I'm totally out of my depth with this stuff...), I think we should merge this, at least "temporarily" (of course, it risks becoming permanent...).


vadz:
Damn, Oracle web site has a nice sense of humour: it's the first passing build since several days and it happens only for this PR.

PR #692, opened 2018-11-22
Add support for MYSQL_OPT_RECONNECT option

bblackcblackc:
Adds an automatically reconnect for MySQL backend

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -21,6 +21,7 @@ project(SOCI)
 option(SOCI_SHARED "Enable build of shared libraries" ON)
 option(SOCI_STATIC "Enable build of static libraries" ON)
 option(SOCI_TESTS "Enable build of collection of SOCI tests" ON)
+option(SOCI_CXX_C11 "Enable c++11" ON)
 option(SOCI_ASAN "Enable address sanitizer on GCC v4.8+/Clang v 3.1+" OFF)
 
 
--- a/.gitignore
+++ b/.gitignore
@@ -42,3 +42,6 @@ CMakeLists.txt.user
 
 # Vagrant
 /.vagrant/
+
+# IDEA
+/.idea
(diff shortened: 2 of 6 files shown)


CMakeLists.txt
```diff
@@ -21,6 +21,7 @@ project(SOCI)
 option(SOCI_SHARED "Enable build of shared libraries" ON)
 option(SOCI_STATIC "Enable build of static libraries" ON)
 option(SOCI_TESTS "Enable build of collection of SOCI tests" ON)
+option(SOCI_CXX_C11 "Enable c++11" ON)
```
mloskot: I'm afraid, this is not going to make it into SOCI 4.0.0

bblackcblackc:
Ok, Mateusz, i will do this one of this days.

2018-11-22 19:28 GMT+03:00, Mateusz Loskot <notifications@github.com>:
> Please, could you also `git rebase master` your branch?
>
> --
> You are receiving this because you authored the thread.
> Reply to this email directly or view it on GitHub:
> https://github.com/SOCI/soci/pull/692#issuecomment-441078311


-- 
С уважением,
Антон.

vadz:
Thanks for updating this and sorry for the delay!

I think that, similarly to #3, this should be ideally be done using `connection_parameters` to provide a much cleaner interface for the same functionality, but I'm going to apply the current version because it's still better than not being able to set this option at all, however I'll change the code to only accept 1 as the option value.

I'm not applying the other changes however:

1. The one to `session` dtor because I don't see why should it be allowed to throw (i.e. I think it shouldn't be)
2. The also unrelated change to C++11 option in `CMakeLists.txt`.
3. The addition of the call to `reconnect()` to `session` ctor because I don't really see why should it be done.

Please let me know if the last one is really needed for MySQL (it definitely shouldn't be for all the other backends). TIA!

PR #706, opened 2019-01-28
Longlong

NStumpel:
Hi,

We use SOCI in our project and we have some servers in 32 and 64 bits.

32 bits use long long and 64 bits use long type for bigint.

In this fork, we have only replaced {unsigned}{long long, long) by {u}int64_t from stdint.h

/usr/include/stdint.h
...
#if __WORDSIZE == 64
typedef long int                int64_t;
# else
__extension__
typedef long long int           int64_t;
#endif

And so we have just one branch to maintain.


vadz:
I'm sorry but I don't understand at all what is this PR trying to fix/change. `long long` is 64 bit in both 32 and 64 bit environments, what is the advantage of using `[u]int64_t` explicitly?

Also, you really shouldn't define these types yourself but use `<cstdint>` instead.

And the PR has a pretty weird history ("fixTU"? "ocean"? "conf-v3rec"?) and contains some unrelated changes (CMake options with French descriptions?). Could you please remake it with just the relevant changes and an explanation of why exactly are they needed/what do they improve? TIA!

PR #720, opened 2019-03-21
Use 64-bit integer for next sequence and last insert ID values

mohansella:
Review this common api changes. If it is fine,will do a seperate PR for sqlite3 changes.


mloskot:
First, the subject of the PR should be clearer, e.g.  "Use 64-bit integer for next sequence and last insert ID values". Currently, it means as much as empty subject.

Second, this seems to introduce C++11 features. There is no plan to switch SOCI 4 to C++11 as requirement.

mohansella:
@mloskot , it can be replaced with `long long`.

mohansella:
The two apis (`get_next_sequence_value` and `get_last_insert_id`) hash been changed to receive reference of `long long` instead of `long`. When upgrading soci, existing users will get compilation error which can be solved by passing the new reference type.

mloskot:
@vadz apparetnly @mohansella has been nagging about merging this PR. What's the final conclusion?

vadz:
Sorry for the delay! I've finally reviewed this and it seems like there are 2 changes here:

1. Change of `session::get_next_sequence_value()` and `get_last_insert_id()` parameter types from `long` to `long long`.
1. Addition of the `soci_sqlite3_integer_type` connection option for SQLite backend.

The first change is clearly needed because practically all (or maybe even all) RDBMS SOCI supports use 64 bits IDs for auto-increment columns or sequences unconditionally, so `long` is not enough in general.  It bothers me personally because we have quite a lot of code using `long`, which has been always enough for us so far, and we'll need to update it, but it's not a reason to not apply this.

The second one looks really ugly however. E.g. it means (AFAICS, please correct me if I'm wrong) that we can only use either `int` or `long long` for _all_ columns accessed via the current connection, while I can very well imagine that we'd want to treat some/most columns as `int` and just a few as `long long`. Would it be possible to accept binding either `int` or `long long` variables to integer-valued columns in SQLite and just throw an exception when reading a value not fitting into an `int`? I think this would be preferable...

In any case, I think the second change should be separated from this PR, which could then be applied.

mloskot:
I overlooed the SQLite change. I agree with @vadz comments.
If the SQLite change is removed from this PR, i.e. keep 1. and remove 2. from @vadz comment, it will be merged immediately.

mohansella:
@mloskot I have removed the changes `2` mentioned above.

@vadz 
> Would it be possible to accept binding either int or long long variables to integer-valued columns in SQLite and just throw an exception when reading a value not fitting into an int?

I am not sure that can be done. Because, unlike other well known RDBMS, in SQLite3, integer width is a variable type to each rows instead of fixed at column level which makes unable to validate the integer type to a column.

include/soci/sqlite3/soci-sqlite3.h
```diff
@@ -332,6 +332,7 @@ struct sqlite3_session_backend : details::session_backend
         }
 
     }
+    data_type integer_type_;
```
mloskot: The clean-up in https://github.com/SOCI/soci/pull/720/commits/a975157a88254bdd3a9377b6c5395f9839bdaf0f missed this one, no?

mohansella: yes. i have missed that line. now updated with removing of that line.

vadz:
Looks good to me, thanks!

tomay3000:
hmmm,
May be this is what broke my code transitioning from v`3.2.3` to v`4.0.0`.
Using `Firebird` v`3.0`, I don't get the next ID after insertion.
May be I will try replacing all My code `long`s with `long long` ones.

Assuming this is what I thought, what should I consider changing to get my code up to date and working for v`4.0.0`

TIA.

Issue #719, opened 2019-03-21
int64_t support for sqlite3


mohansella:
Looks like there is no full support for int64_t with sqlite3. since, sqlite3 datatype `SQLITE_INTEGER` is mapped to `soci::dt_integer` in soci. I guess it needs to be mapped to `soci::dt_long_long`. But it is a breaking change. 

Also, the common api : `soci::get_last_insert_id` gives int32_t which may also need to be changed.

mloskot:
There is chance for a breaking change to be accepted into SOCI 4.
If you'd be willing to provide a PR, so it can be CI tested. Then, we will be able to review and decide.

/cc @vadz

mohansella:
Getting below error while compiling in travis. 

```
In file included from /usr/include/c++/4.6/cstdint:35:0,
                 from /home/travis/build/SOCI/soci/include/soci/soci-backend.h:15,
                 from /home/travis/build/SOCI/soci/include/soci/backend-loader.h:11,
                 from /home/travis/build/SOCI/soci/src/core/backend-loader.cpp:10:
/usr/include/c++/4.6/bits/c++0x_warning.h:32:2: error: #error This file requires compiler and library support for the upcoming ISO C++ standard, C++0x. This support is currently experimental, and must be enabled with the -std=c++0x or -std=gnu++0x compiler options.```

shall i change the flags with -std=c++0x?

mloskot:
See https://github.com/SOCI/soci/pull/720#issuecomment-475188038

mohansella:
@mloskot , Thanks for merging the `get_last_insert_id` changes to support 64 bit value. 

Getting back to the first isssue, 
>`SQLITE_INTEGER` is mapped to `soci::dt_integer` in soci. I guess it needs to be mapped to `soci::dt_long_long` as an option/default to read 64 bit value.

need some comments. because, reading 64bit value from sqlite3 is a must for me.

mloskot:
The 64-bit integer support is there, but you mean https://github.com/SOCI/soci/pull/721#issuecomment-475324131

If there is no clear way to detect those as BIGINT, I have no any immediate recommendations how to work around this.

By the way, `AUTOINCREMENT` is best avoided:

> The AUTOINCREMENT keyword imposes extra CPU, memory, disk space, and disk I/O overhead and should be avoided if not strictly needed. It is usually not needed. 

https://www.sqlite.org/autoinc.html

vadz:
[...]

PR #721, opened 2019-03-21
optional support for 64bit integer read in sqlite3 backend

mohansella:
Rationale discussed in #719


vadz:
Sorry if I'm missing something, but isn't the compatibility breakage really bad here: i.e. do I understand correctly that this means that any existing code using `into(int_var)` will stop working and that there will be no way to fix it other than changing the variable type to `long long` (whether you need it or not)?

Also, from reading the code, it looks like int64 support should already work with SQLite if the column had been declared as "bigint" or "int8" when creating it, isn't it the case?

mohansella:
Yes I can use bigint except.
1.for auto increment PK, only allowed is `INTEGER` which is my case.
2.reading an INTEGER type column with value greater than 32bit (e.g. epochmillis).

And also, I can realize what i am doing is wrong for the old code. So, I think, I should adjust the backend construct with option to change the default datatype without breaking the compatability.

mohansella:
If the connection param is set as follows, we will get int64_t read in sqlite3.
`params.set_option("soci_sqlite3_integer_type", "dt_long_long");`

mloskot:
@vadz apparetnly @mohansella has been nagging about merging this PR. What's the final conclusion?

vadz:
Sorry, I don't even know what patches should be reviewed: there is #719, #720 and #721 (this one), should they all be applied together/sequentially or are some of them included in the other one(s)?

I'm travelling right now but I could review this next week when I get back.

mohansella:
@vadz , sorry for the confusion. both the changes were merged into #720

PR #724, opened 2019-04-04
For performance reasons, use strtoll instead of sscanf.

theovk:
Saves about 50% instructions on my rather large PostgreSQL query result set as measured using valgrind.
Sscanf showed up as a hotspot. Both versions should be functionally identical.

### DIFF
--- a/src/backends/postgresql/common.h
+++ b/src/backends/postgresql/common.h
@@ -11,6 +11,7 @@
 #include "soci/postgresql/soci-postgresql.h"
 #include <limits>
 #include <cstdio>
+#include <cstdlib>
 #include <cstring>
 #include <ctime>
 #include <vector>
--- a/src/backends/postgresql/common.h
+++ b/src/backends/postgresql/common.h
@@ -30,7 +30,14 @@ template <typename T>
 T string_to_integer(char const * buf)
 {
     char * end;
+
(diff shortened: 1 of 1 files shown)


mloskot:
Thanks for the improvement.
Could you address the CI build failure?

https://ci.appveyor.com/project/SOCI/soci/builds/23593803/job/pqaieqclqxeryp7h

> c:\projects\soci\src\backends\postgresql\common.h(32): error C3861: 'strtoll': identifier not found

theovk:
Working on it. No access to a Windows environment so trying to fix this blind...

vadz:
Looks good to me, thanks!

BTW, to obtain the maximal performance we could use [std::from_chars()](https://en.cppreference.com/w/cpp/utility/from_chars) available in MSVS 2019 and which specifically exists for exactly such usage scenarios as we have here.

mloskot:
> No access to a Windows environment so trying to fix this blind...

No problem. You're welcome to employ the CI builds as much as you need.

vadz:
I've created #803, which should fix the build failure with MSVS 2012, and where I plan to extend this change to non-PostgreSQL code too (because I don't see why should it be Postgres-specific), so this PR is superseded by the newer one now.

PR #728, opened 2019-05-05
Deduce reference in boost::fusion::for_each

lebdron:
### Description of the Change
Since Boost 1.68.0, `boost::fusion::for_each` started using `T` instead of `const T&` for function parameter, making it impossible to pass a non-copyable rvalue.
This change passes the function by reference by explicitly specifying template parameters.
Conditional macro based on Boost version is used to avoid "reference to reference" for versions prior to the introduced change in the interface.

Closes https://github.com/SOCI/soci/issues/679

### Benefits
The project compiles with modern versions of Boost

### Possible Drawbacks 
None?


mloskot:
Excellent! Thank you very much!

PR #731, opened 2019-05-08
Add pg reconnection loop

muratovv:
Signed-off-by: Fedor Muratov <muratovfyodor@yandex.ru>

### Proposed changes
In the current implementation, PG backend tries to reconnect only once. But regarding `soci::failover_callback` documentation, a number of connections should be restricted by client code via `failed` method.
The PR contains improvement for PG backend which allows restoring a connection more than once.

### Drawbacks
I haven't seen the usage of `failover_callback` in tests. So the test of the functionality is missed.

### Discussion points
- The implementation is not elegant, and maintainers could guide me in finding better ways to do such changes.
- The number of attempts is limited only by a client's code, and probably there may happen some problems with the connection (e.g. expiration of connection).

### DIFF
--- a/src/backends/postgresql/error.cpp
+++ b/src/backends/postgresql/error.cpp
@@ -95,20 +95,22 @@ details::postgresql_result::check_for_data(char const* errMsg) const
                         callback->started();
                         
                         bool retry = false;
-                        std::string newTarget;
-                        
-                        callback->failed(retry, newTarget);
-                        
-                        if (retry)
-                        {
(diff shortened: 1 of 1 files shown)


mloskot:
Obvious question: what is potential for infinite loop?

Other than that, it looks good.

muratovv:
I think the infinite reconnection is possible only when a client always requests reconnection while the network is lost. But if it is - the client should manage the connection by him/herself.

mloskot:
Sounds good to me. Thanks

PR #733, opened 2019-05-10
Refine try-catch scopes in failover for PostgreSQL

lebdron:
### Description of the Change
- Remove `sessionBackend_` calls from try-catch, making it possible to catch exceptions from them
- Add missing try-catch for `finished` call
- Remove backend argument from connection parameters to avoid "Failed to find shared library for backend postgresql" error

### Benefits
- Consistent exception handling
- No issues when using static library 

### To discuss
In case of exception from `failed` callback and when `retry` was set to true, the execution will continue. Is this behavior acceptable? Maybe set `retry` to `false` in catch block?

### DIFF
--- a/src/backends/postgresql/error.cpp
+++ b/src/backends/postgresql/error.cpp
@@ -108,7 +108,9 @@ details::postgresql_result::check_for_data(char const* errMsg) const
                         }
                         catch (...)
                         {
-                            // ignore exceptions from user callbacks
+                            // do not continue execution because 
+                            // user callback generated an exception
+                            retry = false;
                         }
 
                         if (retry)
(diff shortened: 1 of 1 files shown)


mloskot:
> Is this behavior acceptable?

It's hard to tell for me what author's intention was. 

> Maybe set retry to false in catch block?

If `git blame` does not answer your previous question, I'd go for setting to `false` indeed.

lebdron:
https://github.com/SOCI/soci/blob/3a1f602b3021b925d38828e3ff95f9e7f8887ff7/src/backends/oracle/session.cpp#L88-L105
It seems OK to set `retry` to `false` to make it consistent with Oracle implementation. I will make another commit.

mloskot:
Good finding. Thank you!

PR #739, opened 2019-06-20
Fix crash in Windows by correctly initializing session member in copy…

### LINKED ISSUE #565: ddl_type copy bug
It's not work if compilation not use NRVO.
`soci::ddl_type ddl = sql.create_table("t2");`
Because ddl_type constructor not copy  field **session * s_;**
```
ddl_type::ddl_type(const ddl_type & d) : rcst_(d.rcst_) {
    rcst_->inc_ref();
}
```

davidsummers:
… constructor.  This happened to work fine in Linux/MacOS but crashed in Windows.

-----

Fixes #565

### DIFF
--- a/src/core/once-temp-type.cpp
+++ b/src/core/once-temp-type.cpp
@@ -60,13 +60,14 @@ ddl_type::ddl_type(session & s)
 }
 
 ddl_type::ddl_type(const ddl_type & d)
-    :rcst_(d.rcst_)
+    : s_( d.s_ ), rcst_(d.rcst_)
 {
     rcst_->inc_ref();
 }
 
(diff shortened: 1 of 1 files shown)


vadz:
Thanks, we definitely need to initialize `s_` here, so the fix looks correct to me (I also wonder why is this a pointer and not a reference in the first place).

mloskot:
AFAICT, `ddl_type` just follows (historical?) convention as used in definition of `once_temp_type`, `once_type`, etc.

I don't see any reason who it shouldn't be now a reference.

mloskot:
Merging this fix. Thanks @davidsummers 

The reference suggestions noted as #740

PR #751, opened 2019-09-13
Disable use of std::ptr_fun when compiling as C++17

### LINKED ISSUE #651: Use of deprecated std::ptr_fun
**Problem:**
I would like to use soci with `-std=c++17` applied. There is [one usage](https://github.com/SOCI/soci/blob/master/src/backends/sqlite3/statement.cpp#L495) `std::ptr_fun` which is deprecated in C++17. 

**Solution:**
Replace the usage with a lambda inline if there is compiler support for lambdas (so, C++11 onward I suppose). For example:

```
-    std::string::iterator siter = std::find_if(dt.begin(), dt.end(), std::not1(std::ptr_fun(isalnum)));
+    std::string::iterator siter = std::find_if(dt.begin(), dt.end(), [](const auto c) { return std::isalnum(c); });
```

mloskot:
Fixes #651

### DIFF
--- a/src/backends/sqlite3/statement.cpp
+++ b/src/backends/sqlite3/statement.cpp
@@ -9,11 +9,11 @@
 #include "soci/sqlite3/soci-sqlite3.h"
 // std
 #include <algorithm>
+#include <cctype>
+#include <cstring>
 #include <functional>
 #include <sstream>
 #include <string>
-
-#include <string.h>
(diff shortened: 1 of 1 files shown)


mloskot:
All builds succeeded but Travis CI's jobs for Oracle and DB2 due some package installation issues

PR #753, opened 2019-09-23
Update FindPostgreSQL.cmake from upstream latest version

mloskot:


vadz:
Sorry, not sure what has happened here, but it broke all builds using PostgreSQL , so it clearly can't be applied.

Issue #754, opened 2019-09-24
Postgresql backend does not compile on modern C++ version with -Werror


5cript:
In the postgresql/statement.cpp are 2 functions that might not be used ever.
This is "just" a warning, but since there is a -Werror somewhere, they become errors.
Since these two are the only two errors I had, even though I compiled all of soci and soci-postgresql in C++17 just fine, I post this issue here.

https://github.com/SOCI/soci/blob/c2f934aa95a283f2a02ab7087d9000330f3fc72b/src/backends/postgresql/statement.cpp#L27
and
https://github.com/SOCI/soci/blob/c2f934aa95a283f2a02ab7087d9000330f3fc72b/src/backends/postgresql/statement.cpp#L44

The above is fixable by checking against 
```C++
// note that it is okay to not check _MSC_VER, because if __cplusplus is not defined in older Visual Studio versions, its also not suporting C++17.
#if (__cplusplus >= 201704L)
[[maybe_unused]] void wait_until_operation_complete(postgresql_session_backend & session)
#else
void wait_until_operation_complete(postgresql_session_backend & session)
#endif
```

The below is fixable like:
```C++
#if !defined(SOCI_POSTGRESQL_NOSINGLEROWMODE) and (__cplusplus >= 201704L)
void throw_soci_error(PGconn * conn, const char * msg)
#else
void throw_soci_error(PGconn * conn, const char * msg)
#endif
```

Please let me know if you have strong objections and why.

mloskot:
> I dont even know why the warning does not appear on C++98?

No idea.

> Please let me know if you have (strong) objections and why.

To me your proposal seems not affecting C++98 users or users of any version that's not 201704L, so I have no objections.

I just have a question: the title of your issue says "modern C++ version" (would be helpful if it was more specific), then your change is for `>=201704L`.
Do you consider "modern C++ version" here as C++2a/20 and later? Or, there is a mistake and `201704L` should read `201703L` (i.e. C++17 or later)?

------

Feel free to submit PR.

5cript:
I meant C++17 onwards.
I will make a PR as soon as I can, when I tested it with all different C++ versions.

mloskot:
> I meant C++17 onwards, so I botched the value?

It seems so, given all the below agree that C++17 is `201703L`.

https://en.cppreference.com/w/cpp/preprocessor/replace
https://devblogs.microsoft.com/cppblog/msvc-now-correctly-reports-__cplusplus/
[...]

Issue #752, opened 2019-09-20
Add support for CMAKE_CXX_STANDARD setting interacting with SOCI_CXX11


mloskot:
This is follow-up to discussion started in https://github.com/SOCI/soci/issues/620#issuecomment-533304847 where @5cript suggested:

> And what about optionally checking against __cplusplus if existant/supported,
> secondarily using the special soci cmake option? In principal this allows for
> CMAKE_CXX_STANDARD where supported.

I'd suggest to consider the following, for example:

- If user sets `SOCI_CXX11=ON`, then `CMAKE_CXX_STANDARD=11` is set implicitly.
- If user sets `CMAKE_CXX_STANDARD=11`, then `SOCI_CXX11=ON` is set implicitly
- If user sets `CMAKE_CXX_STANDARD=17` or `=14` or later, then likely `SOCI_CXX11=ON` should be set implicitly too.

(UPDATE: `SOCI_CXX11` is new name since #757)

From CMake perspective, I don't see any reason to fiddle with `__cplusplus`.

Contributions via PRs are always welcome and appreciated.
If any arrives on time for SOCI 4.0.0, then such improvement will be included in this upcoming release.

5cript:
> If user sets SOCI_CXX_C11=ON, then CMAKE_CXX_STANDARD=11 is set implicitly.

If one wants to do that, Checking if both are supplied is important or we might overwrite user settings unintentionally.
But regardeless, I wouldn't say that SOCI_CXX_C11=ON --implies--> CMAKE_CXX_STANDARD=11.
Only the reverse should be true.

```CMake
if (NOT DEFINED CMAKE_CXX_STANDARD AND SOCI_CXX_C11 EQUAL ON) 
  CMAKE_CXX_STANDARD=11
endif()
```

> If user sets CMAKE_CXX_STANDARD=11, then SOCI_CXX_C11=ON is set implicitly

```CMake
if (DEFINED CMAKE_CXX_STANDARD AND (CMAKE_CXX_STANDARD EQUALS 11) OR (CMAKE_CXX_STANDARD GREATER 11))
  SOCI_CXX_C11=ON
endif()
```

> From CMake perspective, I don't see any reason to fiddle with __cplusplus.

I made the remark to __cplusplus as a way to handle it in the C++ code, outside of CMake, in the form of:
```C++
[...]

PR #755, opened 2019-09-28
Improved support for building soci in newer C++ versions.

5cript:
This pull request is the consequence of these issues:
#754 
#752 

Tested it with different C++ versions and tried to stay minimal-impact. I dont want to accidentally break anything.
This should decrease traffic for the uncatchable exception issue from google ;)

### DIFF
--- a/include/soci/soci-platform.h
+++ b/include/soci/soci-platform.h
@@ -97,7 +97,7 @@ namespace std {
 // mode, we just need to check for the minimal compiler version supporting them
 // (see https://msdn.microsoft.com/en-us/library/hh567368.aspx).
 
-#if defined(SOCI_HAVE_CXX_C11) || (defined(_MSC_VER) && _MSC_VER >= 1800)
+#if defined(SOCI_HAVE_CXX_C11) || (defined(_MSC_VER) && _MSC_VER >= 1800) || (__cplusplus >= 201103L)
 # define SOCI_OVERRIDE override
 #else
 # define SOCI_OVERRIDE
@@ -109,7 +109,7 @@ namespace soci
 namespace cxx_details
 {
 
-#if defined(SOCI_HAVE_CXX_C11) || (defined(_MSC_VER) && _MSC_VER >= 1800)
+#if defined(SOCI_HAVE_CXX_C11) || (defined(_MSC_VER) && _MSC_VER >= 1800) || (__cplusplus >= 201103L)
     template <typename T>
     using auto_ptr = std::unique_ptr<T>;
 #else // std::unique_ptr<> not available
@@ -120,7 +120,7 @@ namespace cxx_details
 
 } // namespace soci
 
-#if defined(SOCI_HAVE_CXX_C11) || (defined(_MSC_VER) && _MSC_VER >= 1800)
+#if defined(SOCI_HAVE_CXX_C11) || (defined(_MSC_VER) && _MSC_VER >= 1800) || (__cplusplus >= 201103L)
     #define SOCI_NOT_ASSIGNABLE(classname) \
         classname& operator=(const classname&) = delete;
     #define SOCI_NOT_COPYABLE(classname) \
@@ -136,7 +136,7 @@ namespace cxx_details
 
 #define SOCI_UNUSED(x) (void)x;
 
-#if defined(SOCI_HAVE_CXX_C11) || (defined(_MSC_VER) && _MSC_VER >= 1900)
+#if defined(SOCI_HAVE_CXX_C11) || (defined(_MSC_VER) && _MSC_VER >= 1900) || (__cplusplus >= 201103L)
     #define SOCI_NOEXCEPT_FALSE noexcept(false)
 #else
     #define SOCI_NOEXCEPT_FALSE
--- a/src/backends/postgresql/statement.cpp
+++ b/src/backends/postgresql/statement.cpp
@@ -23,7 +23,9 @@ namespace // unnamed
 {
 
 // used only with asynchronous operations in single-row mode
-
+#if (__cplusplus >= 201703L)
+[[maybe_unused]]
+#endif
 void wait_until_operation_complete(postgresql_session_backend & session)
 {
     while (true)
@@ -41,6 +43,10 @@ void wait_until_operation_complete(postgresql_session_backend & session)
     }
 }
 
+// may be ununsed if the below macro is not defined. 
+#if !defined(SOCI_POSTGRESQL_NOSINGLEROWMODE) && (__cplusplus >= 201703L)
+[[maybe_unused]]
+#endif
 void throw_soci_error(PGconn * conn, const char * msg)
 {
     std::string description = msg;


include/soci/soci-platform.h
```diff
@@ -97,7 +97,7 @@ namespace std {
 // mode, we just need to check for the minimal compiler version supporting them
 // (see https://msdn.microsoft.com/en-us/library/hh567368.aspx).
 
-#if defined(SOCI_HAVE_CXX_C11) || (defined(_MSC_VER) && _MSC_VER >= 1800)
+#if defined(SOCI_HAVE_CXX_C11) || (defined(_MSC_VER) && _MSC_VER >= 1800) || (__cplusplus >= 201103L)
```
vadz: Sorry, I don't understand the changes in this file. `SOCI_HAVE_CXX_C11` should be already defined if `__cplusplus >= 201103L` in `include/soci/soci-config.h`. It doesn't make sense to test for both conditions here, as they're supposed to be equivalent (with the exception of the special case of MSVC without `/std:latest` or similar).

src/backends/postgresql/statement.cpp
```diff
@@ -23,7 +23,9 @@ namespace // unnamed
 {
 
 // used only with asynchronous operations in single-row mode
-
+#if (__cplusplus >= 201703L)
```
vadz: It would be better to define `SOCI_MAYBE_UNUSED` once and then use it in both places (and maybe more places later).

include/soci/soci-platform.h
5cript: > Sorry, I don't understand the changes in this file. SOCI_HAVE_CXX_C11 should be already defined if __cplusplus >= 201103L in include/soci/soci-config.h

how? by user input? Its not for me. Tell me through which mechanism

src/backends/postgresql/statement.cpp
5cript: > The second one blindly fixes the warning instead of actually looking at what it means, namely that these functions are indeed not used when SOCI_POSTGRESQL_NOSINGLEROWMODE is not defined

I didn't see that the other function was also only then not used if the macro was not defined.
I overread the comment above.

> It would be better to define SOCI_MAYBE_UNUSED once and then use it in both places (and maybe more places later).

I do agree, but I am not familiar in your codebase.

include/soci/soci-platform.h
vadz: > > Sorry, I don't understand the changes in this file. SOCI_HAVE_CXX_C11 should be already defined if __cplusplus >= 201103L in include/soci/soci-config.h
> 
> how? by user input? Its not for me. Tell me through which mechanism.

Sorry, I forgot that I modified `soci/soci-config.h.in` locally to define `SOCI_HAVE_CXX_C11` if `__cplusplus >= 201103L`, but this is not in the upstream sources. IMO it could be, but I don't know what @mloskot thinks about it.

> Its not if I run cmake as:
> 
> `cmake .. -DCMAKE_CXX_STANDARD=17`

This should define `SOCI_HAVE_CXX_C11` somehow. If we don't want to add
```cpp
#if __cplusplus >= 201103L
    #define SOCI_HAVE_CXX_C11
#endif
```
to `soci-config.h.in`, then CMake should add `SOCI_HAVE_CXX_C11` to `CONFIGURED_VARIABLES` in this case.

mloskot: > Sorry, I forgot that I modified `soci/soci-config.h.in` locally to define `SOCI_HAVE_CXX_C11`
> if `__cplusplus >= 201103L`, but this is not in the upstream sources.
> IMO it could be, but I don't know what @mloskot thinks about it.

I think it would be useful to define it.

Also, in https://github.com/SOCI/soci/issues/752#issue-496199995 I outlined some obvious rules about relation between `CMAKE_CXX_STANDARD` and `SOCI_CXX_C11`. I forgot about `SOCI_HAVE_CXX_C11 `, which seems to fall into the same category.

vadz: I don't know anything about CMake (except that it's very confusing), but I don't think `SOCI_HAVE_CXX_C11` is really in the same category: it's supposed to be defined if C++11 is available (this is the usual convention for `HAVE` symbols), i.e. it's not directly under the user control, while `SOCI_CXX_C11` is a user-settable option.

OTOH I do see
```cmake
set(SOCI_HAVE_CXX_C11 ${SOCI_CXX_C11} CACHE INTERNAL "Enables C++11 support")
```
in `cmake/SociConfig.cmake`, so it looks like `SOCI_HAVE_CXX_C11` is already set from `SOCI_CXX_C11`, but for some reason it doesn't end up in `soci-config.h`. Any idea why?

mloskot: All three, `CMAKE_CXX_STANDARD` and `SOCI_CXX_C11 and `SOCI_HAVE_CXX_C11`, related to selecting C++11 and that is what I meant. Source of setting of those, user or inference, does not matter really. Shortly, they all three need to be taken into account when any logic is being changed that is related to setting C++11.

>(...) so it looks like SOCI_HAVE_CXX_C11 is already set from SOCI_CXX_C11, but for some reason it doesn't end up in soci-config.h. Any idea why?

Must be because `SOCI_CXX_C11=OFF`.
If you request C++11 like so then `SOCI_HAVE_CXX_C11` is defined:

```
cmake -DSOCI_CXX_C11=ON ..
grep 11 include/soci/soci-config.h
// Enables C++11 support
#define SOCI_HAVE_CXX_C11
```

vadz: But the OP says that it's not set after `cmake .. -DCMAKE_CXX_STANDARD=17`, so do we just need to set `SOCI_CXX_C11=ON` if `CMAKE_CXX_STANDARD` is set to 17 (or 14, 11, 20, ...)?

As I said, I don't really know what is the CMake way of handling this. But I still stand by my original comment, i.e. that it doesn't make sense to test both `__cplusplus` and `SOCI_HAVE_CXX_C11` as they're supposed to be the same thing.

BTW, why do we have the 2nd `C` in `CXX_C11`? It doesn't seem to stand for anything.

mloskot: > But the OP says that it's not set after cmake .. -DCMAKE_CXX_STANDARD=17,

Yes, it is not set because such setting is not implemented.
As far as I understand, it is being / or will be proposed to implement such enhancement to make the three variables set together as necessary. That is why I documented the need for such enhancement as #752 referring to @5cript 's initial comments and idea.

>  so do we just need to set SOCI_CXX_C11=ON if CMAKE_CXX_STANDARD is set to 17 (or 14, 11, 20, ...)?

Yes, I think so. Again, that is what I described in RFC-like issue #752 which I opened as a guide and discussion thread about implementation of improved handling of `SOCI_CXX_C11`, `CMAKE_CXX_STANDARD ` and related variables.

> BTW, why do we have the 2nd C in CXX_C11? It doesn't seem to stand for anything.

AFAIS, it was proposed like that in #327 and I did not object the name (or I missed the `C`).
I don't mind renaming, for example, to `SOCI_CXX_11`.

vadz: I am not qualified to talk about anything CMake-related, so I won't, but IMO it would indeed be better to rename all symbols with `CXX_C11` in their names to use just `CXX11` (or `CXX_11` if you prefer).

mloskot: @vadz I have renamed `SOCI_CXX_C11` to `SOCI_CXX11` in #757

Next step will be to add support for `CMAKE_CXX_STANDARD=11` and later values. I think I can do it as well.

mloskot:
@5cript As mentioned in earlier comments, I renamed the `SOCI_CXX_C11` to `SOCI_CXX11` in #757.
Are you still going to apply the inter-setting with `CMAKE_CXX_STANDARD` as mentioned here https://github.com/SOCI/soci/issues/752#issuecomment-534549449

mloskot:
> @5cript As mentioned in earlier comments, I renamed the `SOCI_CXX_C11` to `SOCI_CXX11` in #757.
> Are you still going to apply the inter-setting with `CMAKE_CXX_STANDARD` as mentioned here [#752 (comment)](https://github.com/SOCI/soci/issues/752#issuecomment-534549449) ?


FYI, I have just proposed implementation for the  `CMAKE_CXX_STANDARD` support in #758

include/soci/soci-platform.h
mloskot: FYI, `CMAKE_CXX_STANDARD` support has now been added. So, this PR could be re-worked and updated to add support for `__cplusplus`.  If it does not happen within next 1-2 weeks, it will be the idea may miss 4.0.0 release.

5cript:
I'd rather make a new pr for the 2 lines of macros in postgresql that "macro-out" the unused functions.
So this can be closed and not-merged.

mloskot:
Sounds good

PR #759, opened 2019-10-06
Potentially Unused functions in Postgresql Backend

5cript:
Fixes issue 754


mloskot:
@vadz If you agree with this change, shall we include in into 4.0.0?

vadz:
Yes, sure, this is exactly what I think should be done and it it's completely harmless, so please include it in 4.0.0 too. TIA!

mloskot:
@5cript Thanks!

PR #765, opened 2019-11-03
Fix MySQL link

tt4g:
Fix MySQL link in documentation.
Howerver, I don't know the previous page, so the new link URL may be wrong.


mloskot:
Thanks for the update

Issue #694, opened 2018-11-28
SQLite backend accesses freed memory when using std::tm


pmir:
When binding std::tm via soci::use(), sqlite backend creates a temporary string to convert the tm to text. This buffer is passed to sqlite3_bind_text and later deleted in post_use/clean_up.

This seems to work when calling execute(true), however, when reading multiple rows with execute(false) + fetch(), sqlite tries to use this (now deleted) text buffer, causing undpredictable results.

A simple query like this is enough to reproduce this issue. The output printed to stdout is garbage. Everything works if `std::tm datetime` is replaced with `std::string` and the date-to-string conversion is done in main()

```c++
#include <soci/soci.h>
#include <soci/sqlite3/soci-sqlite3.h>
#include <iostream>

int main()
{
	soci::session session;
	session.open(soci::sqlite3, ":memory:");

	std::tm datetime {};
	std::string output;

	soci::statement statement(session);
	statement.exchange(soci::use(datetime, "datetime"));
	statement.exchange(soci::into(output));
	statement.alloc();
	statement.prepare("select :datetime");
	statement.define_and_bind();
	statement.execute(false);

	while (statement.fetch())
	{
		std::cout << output << std::endl;
	}

	return 0;
}
```

This seems to match errors returned when running under Dr.Memory
```
Error #1: UNADDRESSABLE ACCESS of freed memory: reading 0x018048f8-0x018048fc 4 byte(s)
# 0 replace_memcpy                                     
# 1 sqlite3VdbeMemGrow
# 2 vdbeMemAddTerminator
# 3 sqlite3VdbeExec
# 4 sqlite3_step   
# 5 soci::sqlite3_statement_backend::fetch
# 6 soci::details::statement_impl::fetch
# 7 main                                                
Note: @0:00:00.699 in thread 26116
Note: next higher malloc: 0x01804998-0x018049ac
Note: prev lower malloc:  0x018048c8-0x018048d8
Note: 0x018048f8-0x018048fc overlaps memory 0x018048f8-0x0180490c that was freed here:
Note: # 0 replace_operator_delete_array                                 
Note: # 1 soci::sqlite3_standard_use_type_backend::post_use
Note: # 2 main
```

This can be fixed by passing SQLITE_TRANSIENT to sqlite3_bind_text under ```sqlite3_statement_backend::bind_and_execute```. However I'm not sure if that's the optimal solution - I'm not familiar with lifetimes of various soci types. Perhaps this buffer can be kept around until the statement is done?

[...]

PR #770, opened 2019-11-23
Fixed std::tm bind for sqlite backend

dirkvdb:
Fixes the issue as reported in #694
Adds a test that reproduces the issue (will fail without the fix)

### DIFF
--- a/src/backends/sqlite3/statement.cpp
+++ b/src/backends/sqlite3/statement.cpp
@@ -265,8 +265,11 @@ sqlite3_statement_backend::bind_and_execute(int number)
                 switch (col.type_)
                 {
                     case dt_string:
-                    case dt_date:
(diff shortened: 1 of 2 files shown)


mloskot:
@dirkvdb Thanks for the contribution

Travis CI job https://travis-ci.org/SOCI/soci/builds/616009673 is failing:

```
/home/travis/build/SOCI/soci/tests/sqlite3/test-sqlite3.cpp:294:11: error: ‘get_time’ is not a member of ‘std’
     ss >> std::get_time(&datetime, "%Y-%m-%d %H:%M:%S");
           ^
```

dirkvdb:
The test used std::get_time which is not supported by the travis version of gcc.I modified the test to use the time_t epoch value instead of parsing the time string.

mloskot:
Thank you!

Issue #666, opened 2018-07-02
Increase allowable buffer size for Oracle service string


ristillu:
Our Oracle connection string looks something like this:
```C++
auto connectionString = "service=(DESCRIPTION=(ADDRESS_LIST=(ADDRESS=(PROTOCOL=TCP)(HOST=somehostname.com)(PORT=1521)))(CONNECT_DATA=(SERVICE_NAME=some_service_name))) user=some_username password=some_password"s;
soci::session sql("oracle", connectionString);
```
In soci/src/backends/oracle/session.cpp there is:
```C++
// arbitrary length for charset conversion buffer
const size_t nlsBufLen = 100;
```
Using that connection string above our code triggers:
```C++
throw soci_error("Service name is too long.");
```
The simple fix for our code was to increase the size of the buffer guard and recompile. But I don't know what a reasonable size is or the set of potential use cases for what is going into the soci service name variable. I used 256, which covered a FQDN and a reasonable length username and password but possibly 512 would be safer.

Some information about potential service strings might be extrapolated from https://docs.oracle.com/cd/E11882_01/network.112/e41945/concepts.htm#NETAG002

kireevm96:
So far, a buffer of 100 characters

irodushka:
And it's very very sad.

PR #772, opened 2019-12-16
fix nlsBufLen to 4096 (4KB)

kireevm96:
fix https://github.com/SOCI/soci/issues/666


vadz:
It seems a bit weird to use 4KiB for username and password which definitely can't be that long. In fact, it looks like they're limited to just 30 characters, and I think it's more clear if we use different buffer size for them, so I've created #773 doing this and replacing this one. I'll merge it once the CI builds pass.

Thanks for your PR and sorry for not applying it, but I prefer the other version.

PR #778, opened 2020-01-07
Updated FindOracle.cmake for Oracle 18 XE

enricodetoma:
Here is an updated CMake configuration for Oracle 18 XE.
Tested with Oracle 18 XE on Windows.

### DIFF
--- a/cmake/modules/FindOracle.cmake
+++ b/cmake/modules/FindOracle.cmake
@@ -52,7 +52,6 @@ set(ORACLE_LIB_DIR
   ${ORACLE_HOME}/sdk/lib       # Oracle SDK
   ${ORACLE_HOME}/sdk/lib/msvc
   ${ORACLE_HOME}/OCI/lib/msvc # Oracle XE on Windows
-  ${ORACLE_HOME}/OCI/lib/msvc/vc14 # Oracle XE on Windows
   # Instant client from rpm
   /usr/lib/oracle/*/client${LIB_SUFFIX}/lib)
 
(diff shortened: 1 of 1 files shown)


cmake/modules/FindOracle.cmake
```diff
@@ -43,15 +43,16 @@ find_path(ORACLE_INCLUDE_DIR
...
+set(ORACLE_OCCI_NAMES libocci occi oraocci10 oraocci11 oraocci12 oraocci18)
+set(ORACLE_NNZ_NAMES nnz10 libnnz10 nnz11 libnnz11 nnz12 libnnz12 nnz18 libnnz18 ociw32)
 
 set(ORACLE_LIB_DIR
   ${ORACLE_HOME}
   ${ORACLE_HOME}/lib
   ${ORACLE_HOME}/sdk/lib       # Oracle SDK
   ${ORACLE_HOME}/sdk/lib/msvc
   ${ORACLE_HOME}/OCI/lib/msvc # Oracle XE on Windows
+  ${ORACLE_HOME}/OCI/lib/msvc/vc14 # Oracle XE on Windows
```
vadz: Is this really needed here? We didn't add `vcN` subdirectories for the previous version...

enricodetoma: Unfortunately `oraocci18.lib` is in `${ORACLE_HOME}/OCI/lib/msvc/vc14` directory, and could not be found.
I will re-check it later when I'm back at my PC, to be sure it's really needed.

enricodetoma: You are right, `${ORACLE_HOME}/OCI/lib/msvc/vc14` is not needed.
I pushed a commit to remove it.

vadz: Thanks for testing!

PR #780, opened 2020-01-26
Fixes false positives in case of identifier with bind mark (:)  - postgresql

rommelst:
Code changed for **postgresql_statement_backend::prepare** to not consider : as bind variable mark when inside postgresql identifier.


rommelst:
Sorry.
I did some mistake when merged master -> Hotfix-3.2.4.
I'll do the pull request again.

PR #781, opened 2020-01-26
Fixes false positives in case of identifier with bind mark (:)

rommelst:
Code changed for postgresql_statement_backend::prepare to not consider : as bind variable mark when inside postgresql identifier.


vadz:
Thanks for the fix! It looks fine to me, but would it be possible to please have a simple unit test showing some query that fails now but is fixed by this patch? This would allow us to not only see that it's fixed right now, but also that it remains fixed in the future. TIA!

rommelst:
I do not have enough experience with github to solve the conflict on the test case.
But i believe that is because i fixed in the 3.2.3 release.
Maybe we would abandon this pull request and i can do it again on the branch master.
any ideas?

rommelst:
I will do again the pull request with the correct branch.

PR #782, opened 2020-01-27
Fixes false positives in case of identifier with bind mark (:)

rommelst:
Code changed for postgresql_statement_backend::prepare to not consider : as bind variable mark when inside postgresql identifier.

### DIFF
--- a/tests/postgresql/test-postgresql.cpp
+++ b/tests/postgresql/test-postgresql.cpp
@@ -990,6 +990,71 @@ TEST_CASE("Bulk iterators", "[postgresql][bulkiters]")
     sql << "drop table t";
 }
 
+
+// false_bind_variable_inside_identifier
+struct test_false_bind_variable_inside_identifier_table_creator : table_creator_base
+{
+    test_false_bind_variable_inside_identifier_table_creator(session & sql)
+        : table_creator_base(sql)
+        , msession(sql) 
+    {
+
+        try
+        {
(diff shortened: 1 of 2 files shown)


tests/postgresql/test-postgresql.cpp
```diff
@@ -990,6 +990,71 @@ TEST_CASE("Bulk iterators", "[postgresql][bulkiters]")
...
+        test_false_bind_variable_inside_identifier_table_creator tableCreator(sql);
+
+        try
+        {
+            sql << "insert into soci_test(\"column_with:colon\") values(2020)";
+            sql << "SELECT column_name FROM information_schema.columns WHERE table_schema = current_schema() AND table_name = 'soci_test';", into(col_name);
+            sql << "SELECT \"function_with:colon\"() ;", into(fct_return_value);
+            sql << "SELECT unnest(enum_range(NULL::\"type_with:colon\"))  ORDER BY 1 LIMIT 1;", into(type_value);
+        }
+        catch(...){}
```
vadz: Why do we do this? Is the test code supposed to throw anything? Because otherwise we really should consider an exception a failure in the test instead of just ignoring them.

rommelst: It was not really mandatory.
removed.

vadz:
Thanks again and sorry for the delay with merging this, but it's finally done now!

PR #784, opened 2020-02-05
Can specify complete sqlite3 open flags in connection parameters

Cavaler:
For example can be used to open database read-only

### DIFF
--- a/docs/backends/sqlite3.md
+++ b/docs/backends/sqlite3.md
@@ -49,6 +49,8 @@ The set of parameters used in the connection string for SQLite is:
 
 * `dbname` or `db`
 * `timeout` - set sqlite busy timeout (in seconds) ([link](http://www.sqlite.org/c3ref/busy_timeout.html))
+* `flags` - set sqlite open flags ([link](https://www.sqlite.org/c3ref/open.html))
+* `readonly` - open database in read-only mode ([link](https://www.sqlite.org/c3ref/open.html))
 * `synchronous` - set the pragma synchronous flag ([link](http://www.sqlite.org/pragma.html#pragma_synchronous))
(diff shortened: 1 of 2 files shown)


vadz:
I'm not sure that passing numeric flags value is really a good way to do things. Wouldn't it be better to add `readonly` parameter instead?

Also, whicheever solution is chosen it really ought to be mentioned in `docs/backend/sqlite3.md`, could you please update that file as well?

Cavaler:
For the sole purpose of 'readonly' - yes, a separate keyword could be used. But there are a lot of other flags SQLite3 supports, and adding separate keyword for every one is not practical in my opinion.

Cavaler:
I added separate 'readonly' flag, and updated description

vadz:
Thanks for the update! I'll merge `readonly` addition, but more I think about it, less I like the idea of specifying numeric flags as strings. This is fragile (too easy to make a mistake), inconsistent with the other backends and it's not clear at all how does it work in conjunction with the other connection string parameters. So I'm afraid I won't merge this part, sorry.

Cavaler:
> but more I think about it, less I like the idea of specifying numeric flags as strings. This is fragile (too easy to make a mistake), inconsistent with the other backends and it's not clear at all how does it work in conjunction with the other connection string parameters.

Well actually there is already one similar parameter, 'timeout', and it works quite well. So 'flags' wouldn't be any different.

vadz:
Err, no, it's not similar at all. Timeout is a number, not a combination of masks. Specifying a number as string is perhaps not ideal, but there is no danger of confusion, hardcoding accidentally wrong value nor any doubt about how does it interact with the other flags.

What problem are you trying to solve here, exactly? I.e. why do you need to pass numeric flags?

Issue #783, opened 2020-02-05
Soci fails to detect certain SQLite column types


Cavaler:
In src/backends/sqlite3/statement.cpp there is a static map describing column types, and it contains strings like "unsigned big int" - correct. However, in describe_column() only first alphanumeric part of description is analyzed - resulting in 'unsigned' for aforementioned type.
Proposed solution - retain all alphanumeric or spaces.

Cavaler:
Here is a proposed patch
[type_describe.patch.zip](https://github.com/SOCI/soci/files/4161033/type_describe.patch.zip)

vadz:
Could you please make a PR out of this patch, so that CI builds run for it? TIA!

PR #785, opened 2020-02-06
Ignore spaces in SQLite column decltypes

Cavaler:
This allows for "BIG INT" as well as "BIGINT", and fixed previouisly broken "UNSIGNED BIG INT"

### DIFF
--- a/tests/sqlite3/test-sqlite3.cpp
+++ b/tests/sqlite3/test-sqlite3.cpp
@@ -237,6 +237,44 @@ TEST_CASE("SQLite vector long long", "[sqlite][vector][longlong]")
     CHECK(v2[4] == 1000000000000LL);
 }
 
+struct type_inference_table_creator : table_creator_base
+{
+    type_inference_table_creator(soci::session & sql)
+        : table_creator_base(sql)
+    {
+        sql << "create table soci_test(cvc varchar (10), cdec decimal (20), "
+               "cll bigint, cull unsigned bigint, clls big int, culls unsigned big int)";
+    }
+};
+
+// test for correct type inference form sqlite column type
+TEST_CASE("SQLite type inference", "[sqlite][sequence]")
+{
+    soci::session sql(backEnd, connectString);
+
+    type_inference_table_creator tableCreator(sql);
+
+    std::string cvc = "john";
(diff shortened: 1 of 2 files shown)


Cavaler:
This resolves #783

tests/sqlite3/test-sqlite3.cpp
```diff
@@ -237,6 +237,44 @@ TEST_CASE("SQLite vector long long", "[sqlite][vector][longlong]")
...
+
+    sql << "insert into soci_test(cvc, cdec, cll, cull, clls, culls) values(:cvc, :cdec, :cll, :cull, :clls, :culls)",
+        use(cvc), use(cdec), use(cll), use(cull), use(cll), use(cull);
+
+    {
+        rowset<row> rs = (sql.prepare << "select * from soci_test");
+        rowset<row>::const_iterator it = rs.begin();
+        row const& r1 = (*it);
+        CHECK(r1.get<std::string>(0) == cvc);
+        CHECK(r1.get<double>(1) == cdec);
```
vadz: You can use CATCH `Approx` for comparing doubles to avoid the warning which resulted in CI build failures.

tests/sqlite3/test-sqlite3.cpp
```diff
@@ -237,6 +237,44 @@ TEST_CASE("SQLite vector long long", "[sqlite][vector][longlong]")
...
+
+// test for correct type inference form sqlite column type
+TEST_CASE("SQLite type inference", "[sqlite][sequence]")
+{
+    soci::session sql(backEnd, connectString);
+
+    type_inference_table_creator tableCreator(sql);
+
+    std::string cvc = "john";
+    double cdec = 12345.0;  // integers stores preciesly in IEEE 754
```
vadz:
```suggestion
    double cdec = 12345.0;  // integers can be stored precisely in IEEE 754
```

Cavaler:
Was there anything wrong with test infraatructure? I don't see how these changes could affect Oracle tests...

vadz:
Sorry, the intermittent Oracle failure has indeed nothing to do with these changes and I strongly suspect that it's due to a bug in SOCI Oracle backend which leaks connections somehow. I hope to find time to look at it one of these days, but for now I'll just merge this PR in spite of the failure.

Thanks for the update!

PR #788, opened 2020-02-10
Update FindOracle.cmake for Oracle 10-19 XE and print additional info;

Lokimed:
I had problem with build soci with oracle on Windows. When i found how i can solve it i thought, maybe i can do some improovement.

### DIFF
--- a/cmake/modules/FindOracle.cmake
+++ b/cmake/modules/FindOracle.cmake
@@ -42,9 +42,14 @@ find_path(ORACLE_INCLUDE_DIR
   # instant client from rpm
   /usr/include/oracle/*/client${LIB_SUFFIX})
 
+set(ORACLE_VERSIONS 10 11 12 18 19)
 set(ORACLE_OCI_NAMES clntsh libclntsh oci) # Dirty trick might help on OSX, see issues/89
-set(ORACLE_OCCI_NAMES libocci occi oraocci10 oraocci11 oraocci12 oraocci18)
-set(ORACLE_NNZ_NAMES nnz10 libnnz10 nnz11 libnnz11 nnz12 libnnz12 nnz18 libnnz18 ociw32)
+set(ORACLE_OCCI_NAMES libocci occi)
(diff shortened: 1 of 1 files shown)


cmake/modules/FindOracle.cmake
```diff
@@ -42,9 +42,19 @@ find_path(ORACLE_INCLUDE_DIR
   # instant client from rpm
   /usr/include/oracle/*/client${LIB_SUFFIX})
 
+set(ORACLE_OCI_MIN_VERSION 10)
+set(ORACLE_OCI_MAX_VERSION 19)
+message(STATUS "ORACLE_OCI_MIN_VERSION ${ORACLE_OCI_MIN_VERSION} , ORACLE_OCI_MAX_VERSION ${ORACLE_OCI_MAX_VERSION}")
+message(STATUS "If your oracle client version is not in diapason between ${ORACLE_OCI_MIN_VERSION} and ${ORACLE_OCI_MAX_VERSION} probably
```
vadz: I'd say "range" instead of "diapason", but mostly I wonder if we really want to always give this message. Maybe do it only if Oracle libraries are not found?

Lokimed: I will try to do as you suggested in the next two days

Lokimed:
I did update. Maybe i did some wrong and possible do better? Or maybe this wiil be ok as temporary solution which is better than current situation

Lokimed:
How i can rerun test? How i saw by logs look look like than database was not available during failed test

vadz:
There is a known problem with Oracle tests (probably due to a bug in SOCI code but I'm not totally sure) which results in sporadic failures in them. I've relaunched this build, but it's not related to your changes, so even if it fails again, it still shouldn't prevent merging this PR.

vadz:
Thanks! I've finally merged this with a couple of minor changes (please let me know if you see any problems with them), sorry for the delay.

PR #790, opened 2020-02-15
Debug Oracle system limits

vadz:
This is a temporary commit just to check what the system limits on the
number of processes really are.


vadz:
Well, this just doesn't work. Modifying `processes` parameter doesn't seem to actually do anything, it seems we need to create something called "spfile" and restart the database to really change it.

Of course, the default value of 100 should be sufficient in any case, as we only use pool of size 10 (twice), so it's really not clear why does the problem arise in the first place.

PR #791, opened 2020-02-15
Try using Xenial on Travis

vadz:


vadz:
I'm going to merge this if there are no objections, mostly in order to get rid of intermittent Oracle CI errors that were extremely annoying and happened regularly (and even most of the times) under Trusty but didn't happen even once yet under Xenial.

The main drawback of this PR is that it doesn't run MySQL ODBC tests any longer because there is no MySQL ODBC driver any longer in Xenial repositories. However this test is still run by AppVeyor, so IMO it's not such a huge loss.

PR #797, opened 2020-03-12
Fix bind const std::vector<boost::optional<>> with soci::use;

Lokimed:


Lokimed:
i will fix test. boost need test in common-tests.h

include/soci/type-conversion.h
```diff
@@ -346,6 +346,7 @@ class conversion_use_type<std::vector<T> >
     }
 
 private:
+    // we really need this? Why change input data??
```
Lokimed: Maybe this funcction must be deleted?

vadz:
This isn't really specific to vectors of `boost::optional`, is it?

I'm not sure what's the purpose of `convert_to_base()`, it looks like some new calls to it were added by c166625a28f7c907318134f625ff5acea7d9a1f8 but then removed in 6da1388f950c049d257a8bdd0612c30656bd9dbb, however the function itself had existed before then and I think the calls to it from `pre_use()` are really necessary...

Lokimed:
Without my fix not compile exchange() with any const vector<boost::optional<>>
But with const vector<> compile.
Maybe i still not understand all situation.

Lokimed:
problem is than for const vector<boost::optional<int>> deduced user_type_tag instead basic_type_tag i think.
I try fix it.

Lokimed:
I did It!!! Soon will be commit

include/soci/type-conversion.h
```diff
@@ -346,30 +346,6 @@ class conversion_use_type<std::vector<T> >
     }
 
 private:
-    void convert_from_base()
```
Lokimed: it was copy paste and have no real usage i as investigated. I hope i did correct.

tests/empty/test-empty.cpp
```diff
@@ -94,7 +94,10 @@ TEST_CASE("Dummy test", "[empty]")
     sql << query, use(i, ind);
 
     std::vector<int> numbers(100);
+    const std::vector<int>& cref_numbers = numbers;
```
vadz: I'm not sure if this really belongs here, wouldn't it be better to add this to the real tests, i.e. to `common-tests.h`?

tests/common-tests.h
```diff
@@ -3227,6 +3227,18 @@ TEST_CASE_METHOD(common_tests, "NULL with optional", "[core][boost][null]")
             CHECK(omi1.is_initialized() == false);
             CHECK(omi2.is_initialized());
             CHECK(omi2.get().get() == 125);
+
+
+            // test us bind with const vector<boost::optional>
```
vadz: What does "us" mean here? Is it a typo?

tests/common-tests.h
```diff
@@ -3227,6 +3227,18 @@ TEST_CASE_METHOD(common_tests, "NULL with optional", "[core][boost][null]")
...
+            {
+                const std::string query = "insert";
```
vadz: This is a bit weird, what do the duplicated statements using this string instead of the literal constant supposed to check?

include/soci/exchange-traits.h
```diff
@@ -148,6 +152,15 @@ struct exchange_traits<long_string>
...
 
+#ifdef SOCI_HAVE_BOOST
+template <typename T>
+struct exchange_traits<boost::optional<T> >
+{
+    typedef typename exchange_traits<T>::type_family type_family;
+    enum { x_type = exchange_traits<T>::x_type };
+};
+#endif
+
```
vadz: Could this be done in `include/soci/boost-optional.h`? It seems like a more appropriate place for it.

include/soci/type-conversion.h
vadz: It's difficult to be sure it's not used, as it's a virtual method... it's called from `standard_use_type::post_use()` and it's not obvious to me why is this _not_ needed.

It needs to be tested with some type `T` for which `type_conversion<T>::from_base()` has a non-trivial implementation.

Lokimed: Okay. I can do fix and not touch this place. 
But very strange - sodc::use bind input data. Why we need change input data? I understand than we need change output with soci::into

tests/common-tests.h
Lokimed: Us is mitake. Here must be "test use bind with const vector<boost::optional>"
I will fix

tests/common-tests.h
Lokimed: I took so test from "soci/tests/empty/test-empty.cpp". I mean about duplicated statements.
I think aim is check char* and std::sting with operator << work correct and after possible bind values.
My aim is only check bind with vector<boost<optional>> with supported type and CONST vector<boost<optional>> with supported type.
I you want i can delete part of checks?

tests/empty/test-empty.cpp
Lokimed: Okay. I try

include/soci/exchange-traits.h
Lokimed: I try do this

tests/empty/test-empty.cpp
Lokimed: This test not run every time? Why better move to common-tests.h?
If you insist, i will fix.

Lokimed:
I watched on class

template <typename T>
class use_type<std::vector<T> > : public vector_use_type

and also used const_cast

Lokimed:
Let's wath to class conversion_use_type

`template <typename T>
class conversion_use_type<std::vector<T> >
     : private base_vector_holder<T>,
       public use_type<std::vector<typename type_conversion<T>::base_type> >`

______________________________

class `base_vector_holder` have no method - convert_from_base

______________________________

Let's watch to
class `use_type`

`template <typename T>
class use_type<std::vector<T> > : public vector_use_type`

Have have no method - convert_from_base

______________________________

Let's watch to
class `vector_use_type`

Have have no method - convert_from_base

but have method virtual void convert_to_base() {}

______________________________


So if possible usw modern C++ and use override in all pace wheen possible, i think we must get compilation error for class conversion_use_type<std::vector<T>
if mark his methos convert_from_base override.
But class conversion_use_type<std::vector<T> is template. So ithink compiler not create method convert_from_base and we can delete this

Lokimed:
Is there any problems which i need fix or explain?

vadz:
Sorry, no, I just didn't have time to address this yet,

I don't think we still need changes in `tests/empty/test-empty.cpp`, but this is very minor and I can remove them myself when merging.

Lokimed:
> Sorry, no, I just didn't have time to address this yet,
> 
> I don't think we still need changes in `tests/empty/test-empty.cpp`, but this is very minor and I can remove them myself when merging.

Okay i will delete changes in `tests/empty/test-empty.cpp` soon.

vadz:
Thanks, this really looks like the right thing to do to me, so even though I still don't fully understand what's going on here (and notably why were all these `const`s not there since the beginning), I'm going to merge it.

Thanks again!

Issue #723, opened 2019-04-02
std::tm timestamp problem with DST


Saring:
In timezone with the 'daylight savings time' (e.g. Europe/Prague), it's not possible to read 31.3. 2:00:00 from database (e.g. postgresql: timestamp without timezone), because mktime() will always change it to 3:00:00

```c++
TEST_CASE("Time DST", "[postgresql][dst-time]") {
	// hack: set Europe/Prague time zone
	setenv("TZ", "Europe/Prague", 1);

	soci::session sql(backEnd, connectString);

	sql << "drop table if exists soci_test;";
	sql << "create table soci_test(id integer, ts timestamp)";

	// winter summer time shifts from 2:00 -> 3:00
	// but we use the database for UTC values
	sql << "insert into soci_test values (1,'2019-03-31 02:00:00'),(2,'2019-03-31 03:00:00')";

	std::tm t1 {};
	std::tm t2 {};

	sql << "select ts from soci_test where id=1", into(t1);
	sql << "select ts from soci_test where id=2", into(t2);

	// it is not possible to determine if original DB value was 2:00 or 3:00
	// because t1 == t2
	bool t1_equals_t2 = 
		t1.tm_mday == t2.tm_mday &&
		t1.tm_mon == t2.tm_mon &&
		t1.tm_year == t2.tm_year &&
		t1.tm_hour == t2.tm_hour &&
		t1.tm_min == t2.tm_min &&
		t1.tm_sec == t2.tm_sec &&
		t1.tm_isdst == t2.tm_isdst &&
		t1.tm_gmtoff == t2.tm_gmtoff;

	// fails
	CHECK_FALSE(t1_equals_t2);
}
```

So do I have to change time zone of whole program, to be able to read '2019-03-31 02:00:00' from the database?

mloskot:
The string with timestamp is parsed here

https://github.com/SOCI/soci/blob/95e4633a5f00b2f75cd7d79a5266476181bc316d/src/backends/postgresql/standard-into-type.cpp#L108-L111

Possibly, the parser needs to be improved

https://github.com/SOCI/soci/blob/56bc9f9c5e1f12c43217f4949cf70489d91ed4c9/src/core/common.cpp#L45-L93

PRs welcome!

Saring:
I dont think it Postgresql only problem.
`soci::details::parse_std_tm` parses 31.3.2019 2:00:00 with `is_dst=-1` and `tm_gmtoff=0`

[...]

PR #799, opened 2020-03-20
Fix 723 stdtm dst problem

rosscsoft:
This is a fix for Issue #723 which causes date/time inaccuracies reading dates when the date/time is on a Daylight Savings Time threshold.

The main fix is to replace the mktime() call in mktime_from_ymdhms() in soci-mktime.h with a call to timegm() instead, which is time-zone agnostic.

A regression test has been added for Issue #723. This will only work on non-Windows platforms so has been #ifdef'd out on Windows.

The code has been compiled on VC++ 2017 and GCC 7.5.0. Test suite has been run on Oracle/Postgres on Windows 10 and Postgres/SQLite on Ubuntu Linux18.04.

(Edited to reflect change in bug fix to use timegm())

### DIFF
--- a/include/private/soci-mktime.h
+++ b/include/private/soci-mktime.h
@@ -11,7 +11,7 @@
 // Not <ctime> because we also want to get timegm() if available.
 #include <time.h>
 
-#ifdef _MSC_VER
+#ifdef _WIN32
 #define timegm _mkgmtime
 #endif
 
--- a/include/private/soci-mktime.h
+++ b/include/private/soci-mktime.h
@@ -28,15 +28,19 @@ mktime_from_ymdhms(tm& t,
                    int year, int month, int day,
                    int hour, int minute, int second)
 {
-    t.tm_isdst = -1;
+    t.tm_isdst = -1;  // DST is unknown
+    t.tm_wday = -1;   // Not set
+    t.tm_yday = -1;   // Not set
+    
     t.tm_year = year - 1900;
     t.tm_mon  = month - 1;
     t.tm_mday = day;
     t.tm_hour = hour;
     t.tm_min  = minute;
     t.tm_sec  = second;
-
-    mktime(&t);
+    
+    // There is no normalisation via mktime() due to Daylight Saving Time
+    // and time zone complications. See issue 723.
 }
 
 // Helper function for parsing datetime values.
--- a/include/private/soci-mktime.h
+++ b/include/private/soci-mktime.h
@@ -11,6 +11,10 @@
 // Not <ctime> because we also want to get timegm() if available.
 #include <time.h>
 
+#ifdef _MSC_VER
+#define timegm _mkgmtime
+#endif
+
 namespace soci
 {
 
@@ -28,19 +32,15 @@ mktime_from_ymdhms(tm& t,
                    int year, int month, int day,
                    int hour, int minute, int second)
 {
-    t.tm_isdst = -1;  // DST is unknown
-    t.tm_wday = -1;   // Not set
-    t.tm_yday = -1;   // Not set
-    
+    t.tm_isdst = -1;
     t.tm_year = year - 1900;
     t.tm_mon  = month - 1;
     t.tm_mday = day;
     t.tm_hour = hour;
     t.tm_min  = minute;
     t.tm_sec  = second;
-    
-    // There is no normalisation via mktime() due to Daylight Saving Time
-    // and time zone complications. See issue 723.
+
+    timegm(&t);
 }
 
 // Helper function for parsing datetime values.
--- a/tests/common-tests.h
+++ b/tests/common-tests.h
@@ -4023,6 +4023,94 @@ TEST_CASE_METHOD(common_tests, "Bind memory leak", "[core][leak]")
     }
 }
 
+// The issue 723 test does not work under Windows as TZ environment variable
+// does not reliably override the system time zone
+#ifndef _MSC_VER
+
+// Helper functions for issue 723 test
+namespace {
+
+    // Creates a std::tm with UK DST threshold 31st March 2019 01:00:00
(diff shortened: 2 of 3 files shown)


vadz:
The tests currently don't pass due to compilation failures because of the use of `nullptr`, could you please fix this in order to relaunch the CI builds?

Beyond this trivial problem, I wonder if it could be still useful to call `mktime()` on a _copy_ of the date/time object just in order to calculate `tm_wday` and `tm_yday`, which we could then copy into the original struct. I think it could be nice as it would allow any existing code which relied on having valid values in these fields after reading from the database to continue working and it doesn't really cost us much to do it, just an extra `mktime()` call. What do you think?

rosscsoft:
Hi Vadz,

That's frustrating! I'm on my way out now but will try and get to it tomorrow.

There is a discussion about the pros and cons of various options captured in the issue thread, I can't remember off the top of my head but I think timegm() was another alternative we looked at where the bug is fixed and the tm_yday and tm_wday are still calculated. I think the issue there is that like mktime(), timegm() only works properly for dates >= 1970 which is probably fine for most cases but there is a tension between having those fields calculated and ensuring consistent behaviour across the whole domain of values. Someone would need to make a call on that.

vadz:
I don't think it's very important to set `tm_{y,w}day` correctly, so I don't consider it as a huge problem to not do it for the dates before the Unix Epoch. And if it does turn out to be problematic, it's really not that difficult to do it ourselves, I wrote the code to do it 20+ years ago for `wxDateTime` and we could always reuse (the relevant parts of) it.

However the problem is that the current change _silently_ breaks the existing code that might rely on having correct values of these fields for the objects coming from the library. And silent breakage like this is arguably the worst thing that library developers can do, so I'd really like to avoid it, especially because it doesn't seem really difficult to do it.

rosscsoft:
Makes sense to me.

I've updated the fix to use timegm()/_mkgmtime() instead and reverted the Oracle tests as this fix should no longer impact them. Also sorted the nullptr issue.

Looks like all the tests are running green except the Oracle ones - from the logs it looks like an environment issue. Give me a shout if that's not the case.

vadz:
Thanks for the update! This is even simpler than I thought if we can assume `timegm()` availability and it looks like it's not a problem, so I let's do it like this. I've just made one small change to avoid any future problems with the time zone being mysteriously changed by ensuring that it's always restored, please let me know if you have any objections.

Other than that, the only possible improvement I see would be to execute the test under MSW too, using the code based on the [example here](https://docs.microsoft.com/en-us/windows/win32/api/timezoneapi/nf-timezoneapi-settimezoneinformation), but this is not really trivial. Have you already tested the new behaviour under MSW in your application? If not, I'd like to at least do it interactively/manually before merging.

Oh, and Oracle problem is indeed unrelated, this is something wrong with server configuration, I think, as it doesn't like having ~10 connections created in the pool test. I'd love to fix this, but my attempts to do it so far have been fully unsuccessful :-(

Thanks again!

rosscsoft:
That's perfect, I did consider an RAII approach to the time zone but was up against it timewise and judged it would only affect the test process so I didn't implement. 

I would say that the majority of my time on this issue has been spent trying to get the test to run on Windows! Fundamentally it doesn't seem possible to set the timezone reliably on a per-process basis as with *NIX. Theoretically it should be possible with TZ but system settings still interfere. I have written a Windows test for our application but it requires the system to be set to UK timezone and will skip otherwise. This is just about acceptable for UK development but seemed unacceptable internationally.

I believe that SetTimeZoneInformation() changes the system-wide settings which I am reluctant to do - it feels like bad karma to change the system-wide time zone (albeit briefly) without knowing what other processes are running on a user's machine.

I chose the timegm() solution to fix our problem initially so it has been tested on our Windows application with both PostgreSQL and Oracle.

vadz:
> I would say that the majority of my time on this issue has been spent trying to get the test to run on Windows! Fundamentally it doesn't seem possible to set the timezone reliably on a per-process basis as with *NIX.

No, I don't think it's possible without some major trickery.

> Theoretically it should be possible with TZ but system settings still interfere. I have written a Windows test for our application but it requires the system to be set to UK timezone and will skip otherwise. This is just about acceptable for UK development but seemed unacceptable internationally.

We probably could just skip the test when not running in a TZ where the problem used to happen, i.e. when `does_mktime_modify_input_hour()` returns false. In fact I really wonder why wouldn't we do this, what do you think?

> I believe that SetTimeZoneInformation() changes the system-wide settings which I am reluctant to do - it feels like bad karma to change the system-wide time zone (albeit briefly) without knowing what other processes are running on a user's machine.

Yes, these are wise words. I was thinking about the tests running on CI machines where it doesn't matter, so I didn't consider this, but perhaps we could test for this and only do this there (it's simple to test for AppVeyor as it defines "APPVEYOR" environment variable)?
 
> I chose the timegm() solution to fix our problem initially so it has been tested on our Windows application with both PostgreSQL and Oracle.

OK, great, thanks for confirming this. I'm ready to merge this, but just hesitate to enable the new test everywhere, provided we simply skip it when using a TZ where the problem doesn't happen. Please let me know if you see any problems with this. Thanks again!

rosscsoft:
Hi Vadz, sorry for the delayed reply - rather a lot going on this end.

Yep can't see a big problem running the test under MSW and skipping where timezone not appropriate. My only concern was creating noise in the test logs by skipping the test for most people on MSW but I suspect it doesn't affect *that* many people.

It may be best (as you have done) to avoid SetTimezoneInformation() as I have noticed it can mess up builds under Visual Studio so that may be an issue on the CI server if there are builds running in parallel with the tests.

rosscsoft:
The changes look good to me

vadz:
Merged now, thanks again for fixing this!

PR #800, opened 2020-03-26
data_type for odbc backend for numeric and decimal types

torbatamas:
The data type for oracle type NUMBER(2) connecting to the oracle DB via ODBC was mapped to double. This is not as needed, also not as implemented in include/soci/column-info.h.
With this commit it is similarly handled.

### DIFF
--- a/scripts/build.bat
+++ b/scripts/build.bat
@@ -3,7 +3,7 @@ rem Runs CMake to configure SOCI for Visual Studio 2017.
 rem Runs MSBuild to build the generated solution.
 rem
 rem Usage:
-rem 1. Copy build.bat to build.locale.bat (git ignored file)
+rem 1. Copy build.bat to build.local.bat (git ignored file)
 rem 2. Make your adjustments in the CONFIGURATION section below
 rem 3. Run build.local.bat 32|64
 rem 4. Optionally, run devenv.exe SOCI{32|64}.sln from command line


vadz:
Thank you for the patch! This looks correct to me, but ideal would be to have a test showing that binding a `NUMERIC(n)` column to an integer works now. We already have `ul numeric(20)` column in the table created by `table_creator_1()` (see `tests/common-tests.h`), so it shouldn't be difficult to add a test checking for this. Could you please do it? This would allow us to ensure that this also works with the other backends correctly and, perhaps even more importantly, that it keeps working in the future.

Thanks in advance!

torbatamas:
I am testing this with odbc, when the underlying server is oracle. But on the server I test it I cannot install all other servers. How can I test them?

tests/odbc/test-odbc-oracle.cpp
```diff
@@ -0,0 +1,301 @@
...
+#include <ctime>
+#include <cmath>
+
+using namespace soci;
+using namespace soci::tests;
+
+// A generic version class: we might want to factor it out later if it is
+// needed elsewhere (it would probably also need to be renamed to something
+// less generic then).
+class odbc_version
```
vadz, line 23: This doesn't seem needed at all here, PostgreSQL test actually uses the version, but this one doesn't.

tests/odbc/test-odbc-oracle.cpp
```diff
@@ -0,0 +1,301 @@
...
+
+    bool has_fp_bug() const SOCI_OVERRIDE
+    {
+        // The bug with using insufficiently many digits for double values was
+        // only fixed in 9.03.0400 version of the ODBC driver (see commit
+        // a5fed2338b59ae16a2d3a8d2744b084949684775 in its repository), so we
+        // need to check for its version here.
+        //
+        // Be pessimistic if we failed to retrieve the version at all.
+        return !m_verDriver.is_initialized() || m_verDriver < odbc_version(9, 3, 400);
```
vadz, line 214: It doesn't make sense to compare ODBC driver version with 9.03.0400 used for PstgresSQL driver.

vadz (commented):
Thanks for the test and sorry for the delay with the response!

I still think that the main change here is correct, but it really can't be applied in the current state because of the issues mentioned below.

Using Oracle ODBC driver on AppVeyor risks to be tricky, but maybe we can use it on GitHub CI if the Oracle docker container used there includes them, but this would unfortunately require more work.

vadz:
Could you please explain why did you add these YAML files? There is, of course, already `ci.yml` in SOCI...

torbatamas:
> Could you please explain why did you add these YAML files? There is, of course, already `ci.yml` in SOCI...

You mentioned GitHub CI, and I am trying to figure out how it works. Sorry, I didn't know it will immediately affect this pull request too.

Could you please give some advice how to continue with testing ODBC with Oracle backend? In this build: https://github.com/torbatamas/soci/runs/2388755328?check_suite_focus=true it did not detected ORACLE, so I assume GitHub CI needs some further configuration to be able to test with oracle too, but I didn't find what to do.

vadz:
> You mentioned GitHub CI, and I am trying to figure out how it works.

Just to be clear, SOCI already uses GitHub Actions for its CI, please the existing `ci.yml` file in master. I.e. there is no need to add new workflows, although it may be useful to modify the existing ones.

> Could you please give some advice how to continue with testing ODBC with Oracle backend? 

Do you have a working ODBC/Oracle configuration locally? If so, could you please confirm that the new test, that you've added, fails there and it passes with your changes?

> In this build: https://github.com/torbatamas/soci/runs/2388755328?check_suite_focus=true it did not detected ORACLE, so I assume GitHub CI needs some further configuration to be able to test with oracle too, but I didn't find what to do.

There is already a job using Oracle and a job using ODBC, so it would be possible to combine them and run ODBC tests with Oracle too, but I'm not asking you to do this (which doesn't mean that it wouldn't be welcome if you did it, of course!) because it's bit complicated.

PR #801, opened 2020-03-29
Switch to using Oracle XE 18

vadz:


vadz:
Initial database creation takes more than an hour on Travis CI, so the current solution doesn't work. We probably could make this work by using `docker -v` to use database files outside of the container and store them somewhere (but even this is not obvious as they take several gigabytes, even for a completely empty database) and reuse them every time. Or maybe someone (@mloskot? @msobczak?) knows how to configure Oracle to use less space and, hopefully, create the database much faster than what it does now. But for now I'm going to leave this aside as #791 seems to provide a more practical solution to the urgent problem of intermittently failing Oracle CI builds: so far this hasn't happened even once under Xenial.

Note: when/if we finally switch to using Oracle 12+ for CI testing, we should also stop using the hack with `_ORACLE_SCRIPT=true` that everybody does for some reason and just connect to the PDB (pluggable database, of which there can be several in a single CDB) using `sys@localhost/XEPDB1` connection string in order to create local users there instead of creating a common user with non-common (not starting with `c##` name) with this hack.

vadz:
Unfortunately Oracle CI builds still fail under Xenial with the same error, although maybe slightly less often. So upgrading to Oracle 18 would still be very much worth it and if anybody has any idea about how to decrease size/speed up creation of the database, it would be great.

vadz:
This is obsolete since the switch/migration to GitHub CI.

PR #804, opened 2020-04-06
Fix error message wording: s/catched/caught/

bblough:


vadz:
Thanks for catching this!

Issue #802, opened 2020-04-03
type_conversion::to_base() is called even if i_null is explicitly passed to use()


ddurham2:
Given the following statement from the documentation (http://soci.sourceforge.net/doc/release/4.0/indicators/#insert-with-null-values)

> Indicator variables can be also used when binding input data, to control whether the data is to be used as provided, **or explicitly overrided to be null**

e.g. 
```
    auto ind = soci::i_null;
    use(x, ind);
```

If you are using a custom type_conversion<>, to_base() is called even if the explicitly given indicator is i_null.  This seems to contradict the docs.

We have a type_conversion<> for boost::any. In to_base(), we make an empty boost::any return i_null for the indicator.  This is the same as your implementation for boost::optional<>.

But since to_base() is called regardless of the value of an indicator given to use(), the boost::any's value gets put into the database, overriding that i_null was passed to use().   I verified that boost::optional<> behaves the same way, which is erroneous, IMO.

I would propose that a condition be added to conversion_use_type<>::convert_to_base like so:
```
   void convert_to_base() {
        if (&ind_ != &ownInd_ && ind_ == soci::i_null) {
             // An indicator was explicitly passed to soci::use(...), and it species i_null.
             // Therefore, NULL should be put into the DB, and there is no reason to convert the value.
             return;
        }

        type_conversion<T>::to_base(value_,
            base_value_holder<T>::val_, ind_);
    }
```
Does that seem reasonable? Or am I missing something?

NOTE: I'm actually using 3.2 but looking to upgrade to 4.0 to fix such issues, but didn't see that the behavior had changed.

vadz:
I agree that I'd expect things to work like this and I'm also surprised that they don't. And the fix also seems reasonable, although I don't really understand why do we need to check for not using `ownInd_` explicitly here as it seems to me that it would be never set to `i_null` in any case, but, of course, this is harmless  in any case and in the worst case just useless.

If you can take the time to add a new test (see `tests/common-tests.h`) checking that `use(foo, i_null)` works as expected and make a PR with this change, I'd be glad to merge it. TIA!

ddurham2:
What that 'if' is trying to say is that if an indicator was passed in to use (&ind_ != &ownInd_) AND what was passed in is presently set to i_null (ind_ == i_null), then that is when we don't need to call to_base().

[...]

PR #807, opened 2020-05-12
fixing bug with boost::optional and explicit null indicator passed to use()

ddurham2:
Fixes issue #802

The issue explains the need.

The first unit test demonstrates the main problem if the fix is not in place.
The second unit test demonstrates the problem if the fix didn't compare with the address of &ownInd_


ddurham2:
@vadz, I completed my PR for this bug.  Everything passes except the db2, odbc and oracle backends.  

But, they're failing on my second unit test (which I only added to demonstrate the internal need for the check against &ownInd_ which we discussed).  So, I can only imagine that this means that those backends are for some reason modifying the indicator's value on INSERT which is mutating the value of ownInd_ which is supposed to stay the same since a prepared statement may execute multiple times (and that this second test would have failed with or without my actual fix).

I don't have these backends to try and troubleshoot.  Do you?  Otherwise, I can only try to track down the problem by just submitting various commits for travis to run and get it to log things.

vadz:
Yes, I definitely will have to look at this as we use ODBC and Oracle backends (and DB/2 is the same as ODBC anyhow, mostly). I'll try to debug this a.s.a.p. (but this might not be very soon in clock time...), thanks for the tests!

vadz:
I've finally fixed the bug in the ODBC backend, see 3d39f7a88fcb761ab3b4bbdf60f124aed29a546b which is currently getting tested as part of #870. If the CI builds there pass, and if you don't have any objections to my changes there done on top of your commits, I'm going to merge that PR.

Thanks again!

ddurham2:
Sure, not problem.  Just so long as the unit tests I added are incorporated and do pass, then we should be fine on this end.
Thanks!

Issue #808, opened 2020-05-15
Memory leak writing CLOB values in Oracle


rosscsoft:
I have a Windows application writing CLOBs every 100ms or so. I am seeing memory usage increasing at a rate of about 10MB/hour which is a problem for a long running application. The size of CLOB has no bearing on the memory leak.

I think I have located the source of the problem which relates to `oracle_standard_use_type_backend::lazy_temp_lob_init()` and `oracle_standard_use_type_backend::clean_up()`.

`lazy_temp_lob_init()` creates a LOB Locator with `OCIDescriptorAlloc()`, then it creates a temporary LOB with `OCILobCreateTemporary()`.  In `clean_up()`, the temporary LOB is deallocated with `OCILobFreeTemporary()` but I believe there should also be a call to `OCIDescriptorFree()` in order to deallocate the LOB Locator.

I have a provisional fix where `clean_up()` is modified so it calls `OCIDescriptorFree(lobp, OCI_TYPE_LOB)` following the call to `OCILobFreeTemporary()`. This seems to fix the leak.

I am not experienced with the Oracle OCI API but the above is my understanding based on what I have gleaned from the Oracle docs, although I could not find a close example of the way SOCI handles CLOBs.  If anyone who is more familiar with the API can let me know if they agree, that would be great.

I am reasonably confident in the fix since `OCIDescriptorFree()` returns success (indicating that there was indeed something to deallocate) and if I call it a second time it returns failure (nothing to deallocate).

I'll wait a short while for any comments then create a PR.

vadz:
Thanks for finding this! Unfortunately don't know much about Oracle API neither and I don't know if the original author of this code (@msobczak?) is still available to answer questions about it, but your description looks plausible to me, so I'd be glad to merge any PRs implementing this. TIA!

rosscsoft:
I've submitted a PR for this now but just wanted to add a couple of bits of information.

This was found on Windows 10, with SOCI 4.0.0 built against Oracle Instant Client SDK 12.2.

Our production application demonstrated the problem and I was able to create a simple Windows application that also demonstrated the problem. However, oddly I could not replicate the issue with a simple console application.

PR #810, opened 2020-05-22
Add call to OCIDescriptorFree() to free LOB Locator and fix memory leak

rosscsoft:
This is a fix for issue #808 where writing to Oracle CLOBs causes a memory leak in Windows.

`oracle_standard_use_type_backend::clean_up()` is modified so it calls `OCIDescriptorFree(lobp, OCI_TYPE_LOB)` following the call to `OCILobFreeTemporary()`. This will deallocate the LOB Locator for the temporary LOB and seems to fix the leak.


vadz:
Thanks for fixing this and sorry for the delay before merging it!

PR #813, opened 2020-06-04
Suppress -Wcast-align in ODBC backend code under ARM

vadz:
This warning should be harmless, so just suppress it to allow building
SOCI under ARMv7hl architecture (all warnings are fatal due to the use
of -Werror).

Closes #812.

Issue #812, opened 2020-06-03
[armv7hl] error: cast from 'char*' to 'TIMESTAMP_STRUCT*'


denisarnaud:
Hello,

the build of SOCI v4.0.0 on Fedora for the ARMv7HL architecture has been consistently failing. An example of such a build is https://koji.fedoraproject.org/koji/taskinfo?taskID=45383684

The build log is https://kojipkgs.fedoraproject.org//work/tasks/3727/45383727/build.log
and the error is happening around the middle of that page:
```bash
make[2]: Entering directory '/builddir/build/BUILD/soci-4.0.0/tmpbuild'
[ 16%] Building CXX object src/backends/odbc/CMakeFiles/soci_odbc_static.dir/standard-into-type.cpp.o
cd /builddir/build/BUILD/soci-4.0.0/tmpbuild/src/backends/odbc && /usr/bin/g++  -DBOOST_ALL_NO_LIB -DHAVE_DL=1 -DSOCI_ABI_VERSION=\"4.0\" -I/builddir/build/BUILD/soci-4.0.0/tmpbuild -I/builddir/build/BUILD/soci-4.0.0/include -I/builddir/build/BUILD/soci-4.0.0/tmpbuild/include -I/builddir/build/BUILD/soci-4.0.0/include/private  -O2 -fexceptions -g -grecord-gcc-switches -pipe -Wall -Werror=format-security -Wp,-D_FORTIFY_SOURCE=2 -Wp,-D_GLIBCXX_ASSERTIONS -specs=/usr/lib/rpm/redhat/redhat-hardened-cc1 -fstack-protector-strong -specs=/usr/lib/rpm/redhat/redhat-annobin-cc1  -march=armv7-a -mfpu=vfpv3-d16 -mtune=generic-armv7-a -mabi=aapcs-linux -mfloat-abi=hard -pedantic -Werror -Wno-error=parentheses -Wall -Wextra -Wpointer-arith -Wcast-align -Wcast-qual -Wfloat-equal -Woverloaded-virtual -Wredundant-decls -Wno-long-long -std=gnu++98  -Wno-variadic-macros   -o CMakeFiles/soci_odbc_static.dir/standard-into-type.cpp.o -c /builddir/build/BUILD/soci-4.0.0/src/backends/odbc/standard-into-type.cpp
[...]

PR #815, opened 2020-06-04
[Build] The -Werror flag is enabled only for Debug builds - issue #812

denisarnaud:
See also #813 and #812

### DIFF
--- a/cmake/SociConfig.cmake
+++ b/cmake/SociConfig.cmake
@@ -63,8 +63,15 @@ if (MSVC)
 
 else()
 
+  # Set the -Werror compilation flag (where all the compilation warnings are
+  # considered as errors) only when building in Debug mode
+  set(SOCI_WERROR_OPTION "")
+  if (CMAKE_BUILD_TYPE STREQUAL "Debug")
+     set(SOCI_WERROR_OPTION "-Werror")
+  endif (CMAKE_BUILD_TYPE STREQUAL "Debug")
+
   set(SOCI_GCC_CLANG_COMMON_FLAGS
-    "-pedantic -Werror -Wno-error=parentheses -Wall -Wextra -Wpointer-arith -Wcast-align -Wcast-qual -Wfloat-equal -Woverloaded-virtual -Wredundant-decls -Wno-long-long")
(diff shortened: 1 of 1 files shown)


denisarnaud:
That change allows to cleanly build on ARMv7hl: https://koji.fedoraproject.org/koji/taskinfo?taskID=45426720
So, awaiting for your accepting to merge that pull request :)

vadz:
I'd really like to modify Travis/Appveyor configs to enable using `-Werror`, we still want to do it there. If you could do it, it would be great, otherwise I'll try to do it later but, as I said, not immediately.

denisarnaud:
> I'd really like to modify Travis/Appveyor configs to enable using `-Werror`, we still want to do it there. If you could do it, it would be great, otherwise I'll try to do it later but, as I said, not immediately.

If you build in Debug mode (`-DCMAKE_BUILD_TYPE=Debug`)), `-Werror` will be activated by default. And Debug mode makes sense for CI (at least to me). I know how to configure Travis CI, but not (yet) AppVeyor.

vadz:
No, sorry, I disagree. We really want to run the tests using the production version, because this is what we ultimately care about. If some optimization breaks something in SOCI (which is not impossible), we want to know about it.

denisarnaud:
> No, sorry, I disagree. We really want to run the tests using the production version, because this is what we ultimately care about. If some optimization breaks something in SOCI (which is not impossible), we want to know about it.

You are absolutely right in principle (and I actually fully subscribe to such principle myself: https://www.12factor.net/dev-prod-parity). Normally, with CI tools, it's pretty straightforward to test several configurations in parallel. In Travis CI, for instance, one test per line in the `env:` section is made. So, we can have `-DCMAKE_BUILD_TYPE=Debug` and `-DCMAKE_BUILD_TYPE=Release` kind of rows (out of my memory; there are surely other ways to that properly).

vadz:
I'd really prefer to keep things simple and just pass extra `CXXFLAGS` to the build. I'm completely CMake ignorant so I always forget how to do it, but surely this must be possible.

denisarnaud:
Ok, no problem, I'm with you: 5afc2f54a3454f06c1001b13603054f74d4d16ab
If, in the `cmake` command-line, you pass the `-DSOCI_ENABLE_WERROR=ON` option, `-Werror` will appear where it used to appear. By default, `-Werror` does not appear.

vadz:
Sorry for the delay, I've finally returned to this and made the changes to keep using `-Werror` in the CI builds in #825, so I'm closing this one in favour of the other PR which I'll merge as soon as the CI builds there finish successfully.

PR #817, opened 2020-07-07
Feature/modern cmake

Leon0402:
Hi everyone,

I want to share my first version of a cmake rewrite. The main purpose of this was to make using soci a lot easier e.g. by providing a SociConfig file and support components and stuff like that. But I thought I could use that to clean the cmake code up as well.

I already tested the code on Linux and Windows with the sqlite3 backend (instructions in the docs in install.md). Theoretically the other backends could work as well, but it's likely that there are some things missing (e.g. the stuff with connection strings for example). 

Unfortunately git also says that I changed source code, but except from a very few lines of code I didn't. That's just my formatting tool, which formatted the code, but I probably can't revert that now.

Feedback, help (testing it, telling me what is missing for the other backends), improving the docs is highly appreciated. I don't have the ressources to test  all of the backends myself. 

I would also be interested in the current roadmap of the project. Most important: Is it possible to switch to C++11 ... as this pull request is also requesting a new cmake version (and therefore is probably something for a soci 5.0.0 release), it would make sense to also bump the C++ version. C++11 is out for quite some time now.

Edit: There are actually a few issues with sqlite3 at the moment I introduced in recent commits, need to fix them.


vadz:
Thanks for your contribution, but unfortunately this is unreviewable because of all the formatting changes, could you please undo/revert them?

It's also going to be really difficult to review so many changes to CMake, I honestly don't know how would be capable of doing it. If you could split this in more manageable chunks, it would be great.

Next version of SOCI (maybe even 4.1 and not 5.0) will indeed require C++11, but this is somewhat orthogonal to CMake stuff.

Leon0402:
Hi, I removed the format issues I had. I'm really not sure how they get into it as I never looked in most of these files. 

In fact, I basically did remove anything and started from scratch regarding cmake. It's not really helping chunking it up in different pieces as so much fundamentally has changed with the cmake 3.x area and I really think it's the easiest to see the end result and verify that everything still works fine. The original maintainer actually did the same thing in his modernCmake branch here, where I picked up some ideas from. 
Actually the cmake code is much easier now, I deleted more cmake code than I added, don't be afraid of it. 

This pull request is WIP though, there are still some problems, especially in the test area. I haven't implemented yet everything what was there before, but I will add that.

vadz:
I'm sorry but it seems unlikely that this is ever going to be merged, there are just way, way too many changes and I have no idea why most of them are done, i.e. what does this actually fix/change for SOCI users. Recent changes (see cb5e09a432cef317ff27ce78e4b29925b61a7a5e and 3f1109b711344cbc087b14dbec27f0d928b6ac20) have fixed a couple of problems in the current CMakefiles and I am really not sure why would we want to entirely change/rewrite everything now. Any concrete remaining problems should ideally be fixed by separate, small(er) PRs. TIA!

PR #819, opened 2020-08-04
Fix sql type detection when the string size is exactly ODBC_MAX_COL_SIZE

wang1972:
Since a null character is added, so must use SQL_LONGVARCHAR in this case.


vadz:
Thanks for the PR! It looks like the [test failures](https://travis-ci.org/github/SOCI/soci/jobs/714873182) are unrelated to the changes here, so this can still be merged and MySQL Valgrind failures will probably need to be suppressed...

PR #824, opened 2020-10-02
Ignore alignment warnings/errors in firebird backend

bblough:
When building on various non-x86 architectures (including arm, mips, sparc), SOCI fails to build from source due to the combination of -Wcast-align and -Werror.

This issue is somewhat related to #812, #813, and #815.  However, rather than being an issue with the ODBC backend, it's the Firebird backend in this case.

Since the warning/error cases all cast the first array element and pass/return the result by value, the alignment should be guaranteed. So it seems appropriate to ignore the warnings as was done in #813.

Applying the changes in this commit (as well as the commits from #813) allows for successful builds on all Debian-supported architectures where the dependencies can be satisfied (i.e., everything but kfreebsd and hurd).

### DIFF
--- a/include/private/firebird/common.h
+++ b/include/private/firebird/common.h
@@ -9,6 +9,7 @@
 #define SOCI_FIREBIRD_COMMON_H_INCLUDED
 
 #include "soci/firebird/soci-firebird.h"
+#include "soci-compiler.h"
 #include <cstdlib>
 #include <cstring>
 #include <ctime>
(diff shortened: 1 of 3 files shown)


vadz:
Thanks for the fixes, I'm merging them with just a minor change: I've replaced hard TABs with spaces for consistency with the rest of SOCI code.

bblough:
No problem, and also, sorry about that.  I usually try to match the formatting, but in this case I apparently forgot.

PR #825, opened 2020-10-10
Disable -Werror by default

vadz:
This completes and replaces #815.


vadz:
For future reference: I have really no idea what has happened with Travis Firebird job, but it mysteriously failed twice in a row (once after manually restarting it) due to not setting the expected SYSDBA password. Equally mysterious, it passed after adding `-d` to expect command, even though this shouldn't have changed anything. I'm leaving this option in place for now, to have more information if this fails again, even if it's unrelated to this PR.

PR #829, opened 2020-10-13
search_paths() not defined in dynamic_backends namespace

xgzeng:
dynamic_backends::search_paths() is declared in backend-loader.h,
but in backend-loader.cpp, search_paths() is not defined in namespace properly.

The problem cause link error if try to use dynamic_backends::search_paths().


vadz:
Thanks for noticing and fixing this! I've cherry-picked this commit with a slightly expanded commit message now.

PR #832, opened 2020-10-31
add namespace qualifiers to sessions to allow macOS builds

noajshu:
Hello,

Due to the polluted global namespace of C++ headers on MacOS, SOCI could not be built with postgresql support on MacOS. I have attached a build log, the error is quite simple. The same problem has occurred before:
[https://github.com/SOCI/soci/issues/605](https://github.com/SOCI/soci/issues/605)
The fix is simply to add the namespace qualifier `soci::` to all session declarations.

[build_log.txt](https://github.com/SOCI/soci/files/5468277/build_log.txt)


vadz:
Thanks, I'm going to apply this, of course (except for whitespace-only changes), but it's annoying to have the same problem happen again. We should probably fix it once and for all by removing `using namespace soci;` from the tests code to enforce the need for using fully qualified names in it.

Issue #762, opened 2019-10-20
`soci` not found when adding as subproject


msis:
I can't seem to be able to add  `soci` as a subproject.

I have tried with `ExternalProject_Add`, and  `FetchContent` but I can't figure out why I keep getting:
```
CMake Error at src/test_app/CMakeLists.txt:21 (add_dependencies):
  The dependency target "soci" of target "test_app" does not exist.
```
---
# Project description
My project structure is the following:
```
└── src
    ├── CMakeLists.txt
    ├── cmake
    │   ├── soci.cmake
    │   └── spdlog.cmake
    ├── test_app
    │   └── CMakeLists.txt
    └── (...)
```
###  `src/CMakeLists.txt`
```CMake
cmake_minimum_required(VERSION 3.11)

set(CMAKE_CXX_STANDARD 11)
set(CMAKE_CXX_STANDARD_REQUIRED ON)
set(CMAKE_CXX_EXTENSIONS OFF)
set(CMAKE_POSITION_INDEPENDENT_CODE ON)

include(cmake/spdlog.cmake)
include(cmake/soci.cmake)

add_subdirectory(test_app)
```
### `src/test_app/CMakeLists.txt`
```CMake
cmake_minimum_required(VERSION 3.11)

project(test_app LANGUAGES CXX)

add_executable(test_app main.cpp)

target_link_libraries(test_app PRIVATE
    spdlog
    SOCI::core
)

add_dependencies(test_app spdlog soci)
```
### `spdlog.cmake`
```CMake
cmake_minimum_required(VERSION 3.11)

message(STATUS "Extern: spdlog v1.4.2")

set(CMAKE_CXX_STANDARD 11)
set(CMAKE_CXX_STANDARD_REQUIRED ON)
set(CMAKE_CXX_EXTENSIONS OFF)
set(CMAKE_POSITION_INDEPENDENT_CODE ON)

include(FetchContent)

FetchContent_Declare( spdlog
    GIT_REPOSITORY https://github.com/gabime/spdlog
    GIT_TAG v1.4.2
    GIT_SHALLOW ON
)
FetchContent_GetProperties(spdlog)
if(NOT spdlog_POPULATED)
    set(SPDLOG_BUILD_EXAMPLE OFF)
    set(SPDLOG_BUILD_EXAMPLE_HO OFF)
    set(SPDLOG_BUILD_TESTS OFF)
    set(SPDLOG_BUILD_TESTS_HO OFF)
[...]

PR #836, opened 2020-11-28
change source dir from where cmake will copy the "soci-config.h" file

spjuanjoc:
- When configuring cmake, trying to run install(), the binary dir might not have been created yet
- Fixes issue: `soci` not found when adding as subproject #762

Notes:
- The travis jobs will not pass since I don't have permissions from this forked repo.
- The file "soci-config.h" is required to test the creation of the conan package from the original repo, not from my forked.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -149,7 +149,7 @@ set(INCLUDEDIR "include" CACHE PATH "The directory to install includes into.")
 ###############################################################################
 # Configuration files
 ###############################################################################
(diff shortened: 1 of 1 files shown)


vadz:
Sorry, I don't know much about CMake, but I don't understand how can this be correct. Looking at `CONFIG_FILE_OUT` definition, it looks like the generated `soci-config.h` will now be created in the source directory and surely we don't want that?

spjuanjoc:
Ok, I got it. I will close this.

PR #840, opened 2020-12-03
Fix for colon in double quotes in single ticks

ckroehnert:
* is broken with issue #782

With the changes of #782 it wasn't possible anymore to use colon in double quotes in single ticks.

* test added


vadz:
Thanks for fixing this!

I'm merging with a slightly modified commit message.

ckroehnert:
My pleasure.

Issue #28, opened 2013-01-17
Undefined behaviour when use unitialized indicator


b-s-a:
This is code from [official example](http://soci.sourceforge.net/doc/exchange.html#indicators):

``` cpp
string name;
indicator ind;

sql << "select name from person where id = 7", into(name, ind);
```

Mostly this code runs fine (because possibility of ind == i_ok is very very low)...
But, if you change code (this is possible, when use one indicator for few DB calls) to:

``` cpp
string name;
indicator ind = i_ok;

sql << "select name from person where id = 7", into(name, ind);
```

Then this code always will return i_ok in case of row not found!

To fix it you are need to add initialization of indicator to the constructor (or other place - I do not use prepared statements more than once) of class standard_indicator_type (may be vector_indicator_type also required to be modified, because user may pass non-empty vector of indicators).

PR #845, opened 2020-12-26
Initialize indicator to i_null even when there is no data at all

vadz:
It is unexpected and error-prone to leave the indicator uninitialized if
a query doesn't retrieve any data, as it's too easy to forget to check
for got_data() and have the code working unpredictably depending on the
initial value of the uninitialized variable.

So always initialize the indicator value to i_null, this can't do any
harm, avoids bugs due to not checking got_data() and makes the code
simpler in the common case when absence of data is the same as having
NULL value in the database.

See #28.

---

This was supposed to be fixed 7.5 years ago, but somehow still isn't. I'd like to fix this because I just don't see any justification for the current behaviour, it's unnecessarily error-prone and doesn't seem to provide any advantages.

@mloskot If you're not too busy with the vacation, I'd appreciate your opinion about this, just in case I'm missing something fundamental here. TIA!

### DIFF
--- a/docs/indicators.md
+++ b/docs/indicators.md
@@ -45,7 +45,10 @@ if (sql.got_data())
 }
 else
 {
-    // no such person in the database
+    // no such person in the database: notice that indicator will have the
+    // value i_null in this branch because often enough missing value can
+    // be handled in the same way as a null one, but you may also distinguish
+    // between the two cases if necessary, as done here by using got_data()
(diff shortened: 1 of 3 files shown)


vadz:
Applied as d202e0d846d6ea0f0e4b29d307b15226dc402cbb (which accidentally referenced the wrong PR number).

PR #846, opened 2020-12-27
Enable LTO in cmake

NobodyXu:
Link time optimization a.k.a LTO is a optimization available in gcc and clang that enables optimization to be performed accross translation unit.

LTO can make the shared library and staticlly linked executable much, much smaller and probably more performant.

I have observed that LTO is able to make the shared library of SOCI smaller by 10% on average with `clang 10.0.0-4ubuntu1` 
and `lld-10.0.0` on linux x86-64:

No LTO:
```
408K    libsoci_core.so.4.0.1      
60K     libsoci_empty.so.4.0.1     
124K    libsoci_mysql.so.4.0.1     
128K    libsoci_odbc.so.4.0.1      
136K    libsoci_postgresql.so.4.0.1
120K    libsoci_sqlite3.so.4.0.1   
```

With LTO:
```
332K    libsoci_core.so.4.0.1      
56K     libsoci_empty.so.4.0.1     
112K    libsoci_mysql.so.4.0.1     
116K    libsoci_odbc.so.4.0.1      
124K    libsoci_postgresql.so.4.0.1
108K    libsoci_sqlite3.so.4.0.1   
```

Though the archives of static libraries become much larger due to more information is stored.

It also can make applications that staticlly linked with SOCI much smaller.
Below is a simple application that uses SOCI:

```c++
#include <soci/soci.h>
#include <soci/postgresql/soci-postgresql.h>

#include <cstdio>
#include <vector>

int main(int argc, char* argv[])
{
    try {
        soci::session sql(soci::postgresql, "dbname=postgres");

        {
            int count;
            sql << "SELECT COUNT(*) FROM sales", soci::into(count);
            std::printf("COUNT(*) = %d\n", count);
        }

        {
            std::vector<int> product_ids;
            product_ids.resize(100);

            sql << "SELECT product_id FROM sales", soci::into(product_ids);
            std::printf("Product ids:");
            for (int product_id: product_ids)
                std::printf("%d, ", product_id);
            std::puts("");
        }

        {
            soci::statement stmt = sql.prepare << "SELECT product_id FROM sales";

            std::vector<int> product_ids;
            product_ids.resize(2);

            stmt.exchange(soci::into(product_ids));
            stmt.define_and_bind();

            stmt.execute();

            std::printf("Product ids:");
            for (; stmt.fetch(); product_ids.resize(2)) {
                for (int product_id: product_ids)
                    std::printf("%d, ", product_id);
            }
            std::puts("");

            stmt.bind_clean_up();
        }
    } catch (const std::exception &e) {
        std::fprintf(stderr, "Error: %s\n", e.what());
        return 1;
    }

    return 0;
}
```

It is built using command

```
clang++ -std=c++17
              -Ofast -Wall -flto -fno-asynchronous-unwind-tables -fno-unwind-tables -fmerge-all-constants
              -I /usr/include/postgresql/ -L /path/to/soci/build/lib/
              -s -Wl,-icf=all,--gc-sections,--as-needed -Wl,--plugin-opt=O3 -fuse-ld=lld
              -lpthread -ldl -lpq -l:libsoci_core.a -l:libsoci_postgresql.a
```

The result is significant:

```
120K    static-with-lto.out   
160K    static-without-lto.out
```

LTO is able to shrink down the executable by 25%.


NobodyXu:
However, due to unknown reason, cmake's `CheckIPOSupported` decides that my `gcc-9.3.0-17ubuntu1~20.04` doesn't support 
lto, which is confusing for me.

CMakeLists.txt
```diff
@@ -10,7 +10,7 @@
 ###############################################################################
 # General settings
 ###############################################################################
-cmake_minimum_required(VERSION 2.8 FATAL_ERROR)
+cmake_minimum_required(VERSION 3.9 FATAL_ERROR)
```
vadz, line 13: This is a big jump. Maybe it's time to finally leave CMake 2.8 behind, but I wonder if it would be possible to optionally have the LTO option only when CMake version is high enough while leaving this requirement unchanged for now? This would make the decision about merging this PR very easy to make.

CMakeLists.txt
```diff
@@ -22,7 +22,35 @@ option(SOCI_SHARED "Enable build of shared libraries" ON)
...
+            include(CheckCXXCompilerFlag)
+            check_cxx_compiler_flag("-fuse-ld=lld" HAS_LLD)
+            if (HAS_LLD)
+                SET(CMAKE_EXE_LINKER_FLAGS  "${CMAKE_EXE_LINKER_FLAGS} -fuse-ld=lld")
+                SET(CMAKE_SHARED_LINKER_FLAGS  "${CMAKE_SHARED_LINKER_FLAGS} -fuse-ld=lld")
+                SET(CMAKE_MODULE_LINKER_FLAGS  "${CMAKE_MODULE_LINKER_FLAGS} -fuse-ld=lld")
+            endif()
+        endif()
+    elseif(supported)
+        message(STATUS "IPO / LTO is supported but conflicted with ASAN")
```
vadz, line 51:
```suggestion
        message(STATUS "IPO / LTO is supported but conflicts with ASAN and not enabled")
```

CMakeLists.txt
```diff
@@ -22,7 +22,35 @@ option(SOCI_SHARED "Enable build of shared libraries" ON)
 option(SOCI_STATIC "Enable build of static libraries" ON)
 option(SOCI_TESTS "Enable build of collection of SOCI tests" ON)
 option(SOCI_ASAN "Enable address sanitizer on GCC v4.8+/Clang v 3.1+" OFF)
-
+option(SOCI_LTO "Enable link time optimization" ON)
```
vadz, line 25: I'm not sure about the default value here. Wouldn't it apply to debug builds too, where we definitely don't want to use it by default?

vadz (commented):
Thanks, adding optional support for LTO is welcome, of course. I am less sure about enabling it by default and increasing CMake requirements just for this however.

CMakeLists.txt
NobodyXu: I'm not sure.
I will investigate further.

CMakeLists.txt
NobodyXu: I didn't take that into consideration.

I would set it to off by default.

NobodyXu: Done

CMakeLists.txt
NobodyXu: As far as I know, [CheckIPOSupported]  and [CMAKE_INTERPROCEDURAL_OPTIMIZATION] which are the official, 
portable support for checking and enabling LTO are only available in 3.9 and later.

If we want to do it without requiring 3.9, we would have to manually write this by ourselves.

[CheckIPOSupported]: https://cmake.org/cmake/help/latest/module/CheckIPOSupported.html
[CMAKE_INTERPROCEDURAL_OPTIMIZATION]: https://cmake.org/cmake/help/latest/variable/CMAKE_INTERPROCEDURAL_OPTIMIZATION.html

NobodyXu: However, I did find an existing cmake script for checking lto support [here] that might be helpful.

[here]: https://github.com/wang-bin/cmake-tools/blob/46865006323046a3be0646373652087a91634aea/tools.cmake#L435

NobodyXu: @vadz I've found a solution:

I put a separate `cmake_minimum_required` check for the LTO code which are executed only when `SOCI_LTO` is set to 
`ON`.

The global cmake version is restored to 2.8

NobodyXu: Only when someone wants LTO support, will they have to use cmake 3.9

NobodyXu:
@vadz Hey, can you take a look at the oracle test on linux? I have no idea why test 2 have all failed with message similar to:

```
2: /home/travis/build/SOCI/soci/tests/common-tests.h:3908: FAILED:
2:   {Unknown expression after the reported line}
2: due to unexpected exception with message:
2:   ORA-12520: TNS:listener could not find available handler for requested type
2:   of server
```

Is there something wrong with `mysql`?

vadz:
Thanks a lot for the updates! Works fine for me, and I can confirm reduction in libraries size (with matching increase in build time, but it's definitely worth it for the release/production builds). Merging, thanks again!

vadz:
> @vadz Hey, can you take a look at the oracle test on linux?

It's "just" a Travis problem. I've allowed failures in this build now, see #847, because I just don't know what else to do with it.

PR #847, opened 2020-12-27
Allow oracle failures on Travis CI

vadz:
I give up, I have no idea what to do about this, so just ignore the errors.


vadz:
P.S. It's really weird how it seems getting worse and worse with time, in the beginning these errors didn't happen at all under Xenial, then they started to appear, then they became more frequent and now I can't get a single good build after restarting the job 5 times. So it probably has something to do with the load of Travis servers (the only thing likely to increase with time), but this still doesn't help with actually fixing the problem.

PR #848, opened 2020-12-27
Enable visibility for gcc and clang

NobodyXu:
Gcc and clang has support visibility just like MSVC for a long time and it trim off a lot internal symbols from the shared library and executable.

This PR define `SOCI_DECL` and `SOCI_<database>_DECL` to `__attribute__ ((visibility ("default")))` on gcc & clang and 
adds option `SOCI_VIS` (`OFF` by default) which set default visibility to hidden.

On my x86-64 linux with `clang-10.0.0-4ubuntu1` and `lld-10.0.0`, setting default visibility to hidden makes the shared libraries 
smaller.

Visibility disabled, LTO enabled:
```
332K    libsoci_core.so.4.0.1      
56K     libsoci_empty.so.4.0.1     
112K    libsoci_mysql.so.4.0.1     
116K    libsoci_odbc.so.4.0.1      
124K    libsoci_postgresql.so.4.0.1
108K    libsoci_sqlite3.so.4.0.1   
```

Visibility enabled, LTO enabled:
```
324K    libsoci_core.so.4.0.1      
52K     libsoci_empty.so.4.0.1     
104K    libsoci_mysql.so.4.0.1     
104K    libsoci_odbc.so.4.0.1      
112K    libsoci_postgresql.so.4.0.1
96K     libsoci_sqlite3.so.4.0.1   
```

The size of statically linked applications aren't affected and this PR should have minor influence over the build time.


CMakeLists.txt
```diff
@@ -23,6 +23,7 @@ option(SOCI_STATIC "Enable build of static libraries" ON)
 option(SOCI_TESTS "Enable build of collection of SOCI tests" ON)
 option(SOCI_ASAN "Enable address sanitizer on GCC v4.8+/Clang v 3.1+" OFF)
 option(SOCI_LTO "Enable link time optimization" OFF)
+option(SOCI_VIS, "Enable symbol trimming via C++ visibility: Preset visibility to hidden" off)
```
vadz, line 26: Hmm, I would be _for_ setting this one to `ON` by default, why not? I.e. I don't think there are any drawbacks to doing it.

Also, let's use something more readable, even if longer, for the option name, e.g. `SOCI_VISIBILITY`.

include/soci/db2/soci-db2.h
```diff
@@ -21,7 +21,11 @@
 //
 // If SOCI_DB2_DECL isn't defined yet define it now
 #ifndef SOCI_DB2_DECL
-# define SOCI_DB2_DECL
+# if __GNUC__ >= 4
+#  define SOCI_DB2_DECL __attribute__ ((visibility ("default")))
+# else
+#  define SOCI_DB2_DECL
+# endif
```
vadz, line 28: We really must not repeat this in all backends, let's define `SOCI_DECL_{EXPORT,IMPORT}` once in `include/soci/soci-platform.h` and use them here.

Also, the definition itself is not correct: we must use DLL export/import declarations even when using gcc under MSW, as visibility is for ELF platforms only. In fact, I don't know if the tests for gcc version is sufficient, there definitely _used_ to be platforms where even much newer gcc versions had trouble with visibility, but maybe they don't exist any more. Ideal would be to use some CMake module detecting visibility support, there must be one.

vadz (dismissed):
Sorry, as much as I'd like to have visibility support in SOCI (and I definitely would!), this can't be applied in the current state due to the problems mentioned in the other comments. Could you please change this as suggested?

Also, could you please add the new option to `docs/installation.md`? TIA!

CMakeLists.txt
NobodyXu: Done

include/soci/db2/soci-db2.h
NobodyXu:
```c++
// If SOCI_DECL isn't defined yet define it now             
#ifndef SOCI_DECL                                           
# if __GNUC__ >= 4                                          
#  define SOCI_DECL __attribute__ ((visibility ("default")))
# else                                                      
#  define SOCI_DECL                                         
# endif                                                     
#endif                                                      
```

The `SOCI_DECL` would only be defined as `__attribute__ ((visibility ("default")))` when it is not defined, which means 
that `_WIN32` is not defined.

Thus I think it won't break building on windows.

NobodyXu: I really don't have any programming experience on windows and I know nothing about its export and import.

Should I define `SOCI_DECL_EXPORT` and `SOCI_DECL_IMPORT` to `__declspec(dllexport)`  and 
`__declspec(dllimport)` respectively on windows and define both of them to `__attribute__ ((visibility ("default")))` on
linux?

NobodyXu: @vadz I've pushed a few commits that resolves this.

NobodyXu:
Wow, the CI suddenly works and all tests are passed.

NobodyXu:
> Sorry, as much as I'd like to have visibility support in SOCI (and I definitely would!), this can't be applied in the current state due to the problems mentioned in the other comments. Could you please change this as suggested?
> 
> Also, could you please add the new option to `docs/installation.md`? TIA!

oops, sorry I forgot it.

I will open another PR to document option `SOCI_LTO` and `SOCI_VISIBILITY` all together.

vadz:
> > Also, could you please add the new option to `docs/installation.md`? TIA!
> 
> oops, sorry I forgot it.
> 
> I will open another PR to document option `SOCI_LTO` and `SOCI_VISIBILITY` all together.

Please feel free to just push another commit to this one, I've already documented `SOCI_LTO` in 931b8fc728823ee291da4dd8d07fdc1eaf9636b9.

NobodyXu:
@vadz OK

NobodyXu:
@vadz I've added the doc.

NobodyXu:
@vadz There seems to be a bug in the firebird test in travis:

```
Firebird: cat /etc/firebird/2.5/SYSDBA.password
ISC_USER=sysdba
ISC_PASSWORD="a89dfb97"

Firebird: restarting
Firebird: DONE
0.02s$ ./scripts/travis/before_script.sh
Running /home/travis/build/SOCI/soci/scripts/travis/before_script_firebird.sh
ISQL Version: LI-V2.5.4.26856 Firebird 2.5
Statement failed, SQLSTATE = 28000
Your user name and password are not defined. Ask your database administrator to set up a Firebird login.
After line 0 in file /tmp/create_soci_test.sql
The command "./scripts/travis/before_script.sh" failed and exited with 1 during .

Your build has been stopped.
```

include/soci/soci-platform.h
```diff
@@ -74,23 +74,31 @@ namespace std {
...
-#endif // _WIN32
-//
-// If SOCI_DECL isn't defined yet define it now
-#ifndef SOCI_DECL
-# define SOCI_DECL
+#  define SOCI_DECL_EXPORT __declspec(dllexport)
+#  define SOCI_DECL_IMPORT __declspec(dllimport)
+# endif
+
+#elif defined(unix) || defined(__unix__) || defined(__unix) && ENABLE_VISIBILITY_SUPPORT
```
vadz, line 87: This ought to be
```suggestion
#elif (defined(unix) || defined(__unix__) || defined(__unix)) && ENABLE_VISIBILITY_SUPPORT
```
to work correctly, `&&` has higher precedence than `||`.

But I wonder why do we need to test for Unix here, shouldn't CMake test fail (or maybe not even run) under non-Unix anyhow? I.e. maybe it could be just
```
#elif defined(SOCI_ENABLE_VISIBILITY_SUPPORT)`
```
?

CMakeLists.txt
```diff
@@ -23,6 +23,7 @@ option(SOCI_STATIC "Enable build of static libraries" ON)
 option(SOCI_TESTS "Enable build of collection of SOCI tests" ON)
 option(SOCI_ASAN "Enable address sanitizer on GCC v4.8+/Clang v 3.1+" OFF)
 option(SOCI_LTO "Enable link time optimization" OFF)
+option(SOCI_VISIBILITY "Enable symbol trimming via C++ visibility: Preset visibility to hidden" on)
```
vadz, line 26: Sorry for nitpicking but I think correct terminology is important: visibility is a property of ELF binaries, not C++ language, so I'd describe it like this:
```suggestion
option(SOCI_VISIBILITY "Enable hiding private symbol using ELF visibility if supported by the platform" on)
```

docs/installation.md
```diff
@@ -103,6 +103,7 @@ Some other build options:
 
 * `SOCI_ASAN` - boolean - Build with address sanitizer (ASAN) support. Useful for finding problems when debugging, but shouldn't be used for the production builds due to extra overhead. Default is `OFF`.
 * `SOCI_LTO` - boolean - Build with link-time optimizations, if supported. This produces noticeably smaller libraries. Default is `OFF`, but turning it on is recommended for the production builds.
+* `SOCI_VISIBILITY` - boolean - Build with C++ visibility if supported. This option produces smaller libraries by trimming internal function symbols. Default is `ON`.
```
vadz, line 106: Same as above:
```suggestion
* `SOCI_VISIBILITY` - boolean - Use hidden ELF visibility for private symbols if supported by the platform. This option produces smaller libraries by avoiding exporting internal function symbols. Default is `ON`.
```

CMakeLists.txt
```diff
@@ -54,6 +55,27 @@ if (SOCI_LTO)
...
+    # Test whether visibility is supported
+    include(CheckCSourceCompiles)
+    check_c_source_compiles(
+        "
+            __attribute__ (( visibility(\"default\") )) int f1() { return 0; }
+            __attribute__ (( visibility(\"hidden\") ))  int f2() { return 1; }
+
+            int main(int argc, char* argv[]) { f1(); f2(); return 0; }
+        "
+        ENABLE_VISIBILITY_SUPPORT
```
vadz, line 68:
```suggestion
        SOCI_ENABLE_VISIBILITY_SUPPORT
```
just to avoid any possible (even if unlikely) conflict with another library predefining this symbol when compiling SOCI headers.

vadz (commented):
Thanks a lot for refactoring this and sorry for the Firebird failure, it's another Travis fluke, unfortunately (at least I've never been able to reproduce this locally).

I think this is almost ready to be merged, but could you please look at the comments below?

TIA!

include/soci/soci-platform.h
NobodyXu: Hmmm, yeah.
Why check for unix when there is `ENABLE_VISIBILITY_SUPPORT`?
The code can definitly be simplified to `#elif defined(SOCI_ENABLE_VISIBILITY_SUPPORT)`

CMakeLists.txt
NobodyXu: Sure.

NobodyXu:
OOPs, my PR is broken on the CI.

I will fix it.

NobodyXu:
It's strange that the travis CI suddenly warns me about cmake policy not set when the previous builds actually all succeeded.

NobodyXu:
Fixed by manually adding `#cmakedefine SOCI_ENABLE_VISIBILITY_SUPPORT` in `soci-config.h.in`.

vadz:
Thanks for the updates!

`CONFIGURED_VARIABLES` is handled manually in `CMakeLists.txt` AFAICS. I don't really know why is it done like this, but I'd prefer to stay with this schema for now, if we want to switch to using (more standard?) `#cmakedefine`, we should do it for all the options too and not mix and match both approaches as this is just confusing.

NobodyXu:
Sure, I will take a look at `CMakeLists.txt` when I am free.

NobodyXu:
> Thanks for the updates!
> 
> `CONFIGURED_VARIABLES` is handled manually in `CMakeLists.txt` AFAICS. I don't really know why is it done like this, but I'd prefer to stay with this schema for now, if we want to switch to using (more standard?) `#cmakedefine`, we should do it for all the options too and not mix and match both approaches as this is just confusing.

Just find the solution and pushed the commit.

vadz:
Thanks, looks good to me!

I'm just going to test this with one of my own projects using SOCI (it _should_ work, but I'd like to make sure) and then merge -- although, unfortunately, this probably won't happen before the end of the week.

BTW, I'm going to squash merge this one as I don't think we really need the individual commits here, but please let me know if you object. Thanks again!

vadz:
Sorry, I've finally tested this and I'm getting this from CMake
```
-- Performing Test SOCI_HAVE_VISIBILITY_SUPPORT
-- Performing Test SOCI_HAVE_VISIBILITY_SUPPORT - Success
...
-- Configuring done
CMake Warning (dev) at cmake/SociBackend.cmake:347 (add_executable):
  Policy CMP0063 is not set: Honor visibility properties for all target
  types.  Run "cmake --help-policy CMP0063" for policy details.  Use the
  cmake_policy command to set the policy and suppress this warning.

  Target "soci_empty_test" of type "EXECUTABLE" has the following visibility
  properties set for CXX:

    CXX_VISIBILITY_PRESET

  For compatibility CMake is not honoring them for this target.
```
and several more like this. Do you have any idea what does it try to say? I haven't looked at it myself yet...

NobodyXu:
It is about choosing the old behavior or the new behavior for visibility:

```
CMP0063
-------

Honor visibility properties for all target types.

The ``<LANG>_VISIBILITY_PRESET`` and
``VISIBILITY_INLINES_HIDDEN`` target properties affect visibility
of symbols during dynamic linking.  When first introduced these properties
affected compilation of sources only in shared libraries, module libraries,
and executables with the ``ENABLE_EXPORTS`` property set.  This
was sufficient for the basic use cases of shared libraries and executables
with plugins.  However, some sources may be compiled as part of static
libraries or object libraries and then linked into a shared library later.
CMake 3.3 and above prefer to honor these properties for sources compiled
in all target types.  This policy preserves compatibility for projects
expecting the properties to work only for some target types.

The ``OLD`` behavior for this policy is to ignore the visibility properties
for static libraries, object libraries, and executables without exports.
The ``NEW`` behavior for this policy is to honor the visibility properties
for all target types.

This policy was introduced in CMake version 3.3.  CMake version
3.16.3 warns when the policy is not set and uses ``OLD`` behavior.  Use
the ``cmake_policy()`` command to set it to ``OLD`` or ``NEW``
explicitly.

.. note::
  The ``OLD`` behavior of a policy is
  ``deprecated by definition``
  and may be removed in a future version of CMake.
```

The output above is taken from `cmake --help-policy CMP0063`.

This warning can be fixed by `cmake_policy(SET CMP0063 OLD)`, which uses the old behavior that only set visibility for shared library and executables with ``ENABLE_EXPORTS`` property set.

I personally think setting visibility for shared library is enough for SOCI, as those who static link with SOCI would usually build an executable that have all symbols stripped.
It is the shared library that needs these private symbols stripped.

NobodyXu:
Well, I just tried using the old behavior but cmake then reminded me that it is deprecated and will be removed...
Guess I should have predict that since it is the "old behavior".

I will set the policy to new behavior then.

vadz:
Thanks for fixing this, I'm merging it now.

PR #856, opened 2021-03-11
Fix using vectors of custom types

vadz:
The problem is that we can't really bind the parameters before we have the actual data and when using custom types this actual data only becomes available after calling `convert_to_base()` from `vector_use_type::pre_use()` in the core. So the backends that bound parameters before (ODBC and Oracle) didn't use the correct data and so the simple test added in the last commit here failed.

I'm still not sure what's going to happen when `execute()` (and hence `pre_use()`) is called multiple times and whether rebinding the parameters every time is a good idea. Perhaps we should only do it in the first call?


vadz:
Note: d202e0d846d6ea0f0e4b29d307b15226dc402cbb above was supposed to reference #845, not this one.

PR #871, opened 2021-04-13
Fix Firebird memory leak

thesiv:


vadz (commented):
Thanks, this should fix the leak (I still don't understand why wasn't it detected by the CI build though?), but it looks like we could save on `isc_create_blob()` which might be not "free" if we reused the blob instead of recreating it, i.e. did
```cpp
if (blob_)
    blob->trim(0);
else
    blob = new ...;
```
Wouldn't this work just as well?

thesiv:
```cpp
if (blob_)
    blob->trim(0);
else
    blob = new ...
```
I tested that it works. I.e. fix the leak as well.

vadz:
Thanks for testing!

PR #874, opened 2021-04-18
Oracle rowset when precision=0 ORA-01455 error

530967061:


vadz:
Thanks!

I was initially rather puzzled by this because I didn't understand what did precision of 0 actually mean as [Oracle docs](https://docs.oracle.com/cd/B28359_01/server.111/b28318/datatype.htm#CNCPT313) just says

> you can also specify a precision (total number of digits) and scale (number of digits to the right of the decimal point)

and having 0 total digits doesn't make sense. But in [another place](https://docs.oracle.com/cd/E11882_01/appdev.112/e10646/oci06des.htm#LNOCI16458) they say

> When precision is 0, `NUMBER(precision, scale)` can be represented simply as `NUMBER`.

so actually precision=scale=0 means default `NUMBER`, i.e. `NUMBER(38, 10)`.

Taking all this into account and, also because I think the test can be simplified to avoid dealing with `time_t` stuff, which is not required to reproduce the bug, I've created #875 with my variant of the fix, based on your changes. Please let me know if you see any problems with it, otherwise I'll merge it soon.

For now I'm closing this one already, as it's redone in the other PR.

Thanks again!

PR #884, opened 2021-05-10
Change my_bool > bool

marimeireles:
Hello,

Something changed upstream in the library you're importing for the `mysql` port.
The current release version of SOCI doesn't compile if you enable `mysql`, this PR fixes it.

Maybe we should check this on the CI and build and run all of the backends.

Cheers!

### DIFF
--- a/src/backends/mysql/session.cpp
+++ b/src/backends/mysql/session.cpp
@@ -355,7 +355,11 @@ mysql_session_backend::mysql_session_backend(
     }
     if (reconnect_p)
     {
-        my_bool reconnect = 1;
+        #if MYSQL_VERSION_ID < 8
+            my_bool reconnect = 1;
+        #else
+            bool reconnect = 1;
+        #endif
         if (0 != mysql_options(conn_, MYSQL_OPT_RECONNECT, &reconnect))
         {
             clean_up();


vadz:
CI does check MySQL backend, but clearly `my_bool` still exists in the version it uses. I admit I'm rather surprised that it would be removed, do you have any pointers to the corresponding change in MySQL?

The problem with just merging this is that it might break the build for older versions if they require using `my_bool` rather than `bool`, so it would be nice to check that this is not the case for any versions still supported by SOCI.

TIA!

SylvainCorlay:
From the mysql 8.0.1 release note:

> Incompatible Change: The `my_bool type` is no longer used in MySQL source code. Any third-party code that used this type to represent C boolean variables should use the `bool` or `int` C type instead.
> 
> Note
> The change from `my_bool` to `bool` means that the `mysql.h` header file now requires a C++ or C99 compiler to compile.

SylvainCorlay:
I was looking into it, and it looks like a mess:
https://bugs.mysql.com/bug.php?id=87337

SylvainCorlay:
Maybe a way to go would be to use a `char` value instead of `bool`...

vadz:
Thanks for looking at this!

If I'm reading this correctly, `my_bool` is not supposed to be exported since 8.0.3, but it still looks like it used to be `char` before 8.0 and has to be `bool` thereafter. So I think to be really safe we do need a compile-time check and define `soci_my_bool` as either `char` for < 8 or `bool`.

If anybody can modify this PR (or create another one) doing it like this, I'd be glad to merge it, but I'd rather not apply this as is, as I think it must break compilation with the older MySQL.

marimeireles:
>So I think to be really safe we do need a compile-time check and define soci_my_bool as either char for < 8 or bool.

Hey @vadz I don't know what you mean here. You mean instead of `bool` I should use `unsigned char`?

marimeireles:
Or should I create `ifs` in the cmake depending on which version of `mysql` the user is running...?

vadz:
I thought to just use `#if MYSQL_VERSION_ID > some-value-for-8.0` in the code.

marimeireles:
Ok, so I think this looks good?
Before mysql 8 we're using `mysql bool` and after it `uchar` which must be compatible with everything bigger than 8.

src/backends/mysql/session.cpp
```diff
@@ -6,6 +6,8 @@
 // http://www.boost.org/LICENSE_1_0.txt)
 //
 
+#include <iostream>
+
```
vadz, line 10: This seems accidental.

src/backends/mysql/session.cpp
```diff
@@ -355,7 +357,11 @@ mysql_session_backend::mysql_session_backend(
     }
     if (reconnect_p)
     {
-        my_bool reconnect = 1;
+        #if MYSQL_VERSION_ID < 8
+            my_bool reconnect = 1;
+        #else
+            unsigned char reconnect = 1;
```
vadz, line 363: Sorry, why `unsigned char` and not just `bool`? [The docs](https://dev.mysql.com/doc/c-api/8.0/en/mysql-options.html) seem to say that we should use `bool`.

vadz (commented):
Sorry with the delay with reply, but it indeed looks like it ought to be as simple as that, except that I wonder what's wrong with just using `bool`?

src/backends/mysql/session.cpp
marimeireles: Hum yeah, you're right, it might be a problem with version 8.0.1:

>- since my_bool is gone in MySQL 8.0.0 and 8.0.1, it would be natural to
  "typedef bool my_bool", since bool is used everywhere my_bool was
  previously used, and the typedef allows to avoid massive application
  code changes;
>- now 8.0.2 re-introduces my_bool as an alias to char, which is
  compatible with the pre-8.0 definition of my_bool, but is incompatible
  with bool. Which conflicts with "typedef bool my_bool":
>drv_mysql.c:54:14: error: typedef redefinition with different types ('bool' vs 'char')
typedef bool my_bool;

I'll change it to bool

vadz:
Merged now, thanks!

PR #885, opened 2021-05-15
Corrected option value.

Spixmaster:
The option value was in lower case.


vadz:
Thanks for tidying this up!

PR #886, opened 2021-05-17
Fix compile issue wrt SIGSTKSZ no longer being a constant

denisarnaud:
* On recent Fedora distributions, there is a compilation error in the test part similar to https://github.com/onqtam/doctest/issues/473
* I have applied a patch similar to https://github.com/onqtam/doctest/pull/474, which is admittedly very (too?) simple.
* A more elaborate patch could be derived from https://github.com/onqtam/doctest/commit/8ee91f37ba27397ef5eeedd2e542a4aaaa33ddc1 . If anyone would give it a try for a longer term patch, that would be welcome. In the meantime, this pull request allows to compile without error.

### DIFF
--- a/tests/catch.hpp
+++ b/tests/catch.hpp
@@ -6489,7 +6489,7 @@ namespace Catch {
         static bool isSet;
         static struct sigaction oldSigActions [sizeof(signalDefs)/sizeof(SignalDefs)];
         static stack_t oldSigStack;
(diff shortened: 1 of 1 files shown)


vadz:
Thanks for fixing this, the longer term solution here is to start using C++14 and Catch 2.x, but for now let's indeed just apply this to avoid the error.

PR #888, opened 2021-06-13
Fix simultaneous using of long strings vectors and vectors of trivial type values for ODBC

thesiv:
9b4c5484faa4a753b6bdee19673f41912c5b5835 makes the error (sorry) when fetching a vector of trivial type values by a single row. Because vectors of trivial type values binded directly as a pointer to the first element (long string and other types converted in `odbc_vector_into_type_backend::exchange_rows()`).

### DIFF
--- a/include/soci/odbc/soci-odbc.h
+++ b/include/soci/odbc/soci-odbc.h
@@ -145,8 +145,8 @@ struct odbc_vector_into_type_backend : details::vector_into_type_backend,
     // SQLLEN is still defined 32bit (int) but spec requires 64bit (long)
     inline SQLLEN get_sqllen_from_vector_at(std::size_t idx) const;
 
-    // Redefine the single vector value at the given index to the first row.
-    // Used when vector values fetched by a single row.
+    // Rebind the single vector value at the given index to the first row.
+    // Used when vector values are fetched by single row.
     void redefine_row(std::size_t rowInd);
 
     std::vector<SQLLEN> indHolderVec_;
--- a/include/soci/odbc/soci-odbc.h
+++ b/include/soci/odbc/soci-odbc.h
@@ -145,6 +145,10 @@ struct odbc_vector_into_type_backend : details::vector_into_type_backend,
     // SQLLEN is still defined 32bit (int) but spec requires 64bit (long)
     inline SQLLEN get_sqllen_from_vector_at(std::size_t idx) const;
 
+    // Redefine the single vector value at the given index to the first row.
+    // Used when vector values fetched by a single row.
+    void redefine_row(std::size_t rowInd);
+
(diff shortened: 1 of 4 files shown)


include/soci/odbc/soci-odbc.h
```diff
@@ -145,6 +145,10 @@ struct odbc_vector_into_type_backend : details::vector_into_type_backend,
     // SQLLEN is still defined 32bit (int) but spec requires 64bit (long)
     inline SQLLEN get_sqllen_from_vector_at(std::size_t idx) const;
 
+    // Redefine the single vector value at the given index to the first row.
```
vadz, line 148: I think this should say "rebind" rather than redefine.

include/soci/odbc/soci-odbc.h
```diff
@@ -145,6 +145,10 @@ struct odbc_vector_into_type_backend : details::vector_into_type_backend,
...
+    // Used when vector values fetched by a single row.
```
vadz, line 149:
```suggestion
    // Used when vector values are fetched by single row.
```

src/backends/odbc/statement.cpp
```diff
@@ -255,6 +255,25 @@ odbc_statement_backend::fetch(int number)
 
     for (std::size_t i = 0; i < fetchesCount; i += rowsPerFetch)
     {
+        if (fetchVectorByRows_) {
```
vadz, line 258:
```suggestion
        if (fetchVectorByRows_)
        {
```

src/backends/odbc/statement.cpp
```diff
@@ -255,6 +255,25 @@ odbc_statement_backend::fetch(int number)
...
+            switch (intoType_)
+            {
+            case bt_standard:
+                // Standard intos processed in standard define_by_pos().
+                break;
+            case bt_vector:
+                // Unfortunately we ought to redefine all vector intos which
```
vadz, line 265:
```suggestion
                // Unfortunately we need to redefine all vector intos which
```
Also not sure why does it say "unfortunately"...

src/backends/odbc/statement.cpp
```diff
@@ -255,6 +255,25 @@ odbc_statement_backend::fetch(int number)
...
+                // binded directly as first elements of vectors.
```
vadz, line 266:
```suggestion
                // were bound to the first element of the vector initially.
```

vadz (commented):
It seems to me that we're manually redoing what `SQLFetchScroll()` does here, but it probably doesn't matter as we plan to get rid of this code later anyhow.

thesiv:
Fixed.

vadz:
Merged with another minor renaming, thanks.

PR #891, opened 2021-06-24
Session from connection pool call open need reset backend

### LINKED ISSUE #890: session::open for sessions from pool need to set backEnd_ again
If we write code like：
`
connection_pool pool(10);
session s(pool);
s.open(...);
s.commit();
`

we will encounter an exception: "Session is not connected."
so we need to reset backEnd_ when we do open for session from pool

void session::open(connection_parameters const & parameters)
{
    if (isFromPool_)
    {
        pool_->at(poolPosition_).open(parameters);
        **_backEnd_ = pool_->at(poolPosition_).get_backend();_**
    }
    else
    {
        if (backEnd_ != NULL)
        {
            throw soci_error("Cannot open already connected session.");
        }

        backend_factory const * const factory = parameters.get_factory();
        if (factory == NULL)
        {
            throw soci_error("Cannot connect without a valid backend.");
        }

        backEnd_ = factory->make_session(parameters);
        lastConnectParameters_ = parameters;
    }
}

530967061:
fix #890

### DIFF
--- a/src/core/session.cpp
+++ b/src/core/session.cpp
@@ -153,6 +153,7 @@ void session::open(connection_parameters const & parameters)
     if (isFromPool_)
     {
         pool_->at(poolPosition_).open(parameters);
+        backEnd_ = pool_->at(poolPosition_).get_backend();
     }
     else
     {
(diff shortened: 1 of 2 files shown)


vadz:
Thanks for the fix, will merge it in a moment!

PR #892, opened 2021-06-29
Bug fix and many simplifications to odbc_statement_backend

vadz:
For some reason this vector, added recently in 9b4c5484 (Fetch vectors
by a single row for huge strings in ODBC backend, 2021-03-24), used
"void*" for its elements instead of "odbc_vector_into_type_backend*",
requiring casts in several places.

Do use the actual type now, as there doesn't seem to be any problems
with doing this and it simplifies the code and makes it safer.


vadz:
@thesiv I'm going to merge this soon, but I'd still like you to have a look at this when you're back and tell me if you see anything wrong here. Other improvements are definitely possible, but this already makes things much simpler and, of course, fixes an important bug with out of bounds access to the indicators vector.

PR #896, opened 2021-07-10
Update FindOracle.cmake

gypsy5oul:
For checking oracle 21 version


vadz:
Thanks, should we add 20 too or is there some reason to skip it?

gypsy5oul:
Yes you can add 20 also

Get Outlook for iOS<https://aka.ms/o0ukef>
________________________________

PR #898, opened 2021-07-11
[MySQL] my_bool has been replaced by bool

denisarnaud:
`my_bool` has been replaced by `bool` in newer versions of MySQL. For instance, https://github.com/r-dbi/RMariaDB/pull/107

### DIFF
--- a/src/backends/mysql/session.cpp
+++ b/src/backends/mysql/session.cpp
@@ -356,7 +356,7 @@ mysql_session_backend::mysql_session_backend(
     if (reconnect_p)
     {
         #if MYSQL_VERSION_ID < 8
-            my_bool reconnect = 1;
+            bool reconnect = 1;
         #else
             bool reconnect = 1;
         #endif


vadz:
Sorry, but this change clearly doesn't make sense, what's the point of having this `#if` check if both of its branches are the same?

It was introduced in 7c431d9f073dec786fede7a78d6ff4ed44bbdb92, have you seen the discussion in #884? Which version of MySQL do you use and what is the value of `MYSQL_VERSION_ID` for you?

denisarnaud:
> It was introduced in [7c431d9](https://github.com/SOCI/soci/commit/7c431d9f073dec786fede7a78d6ff4ed44bbdb92), have you seen the discussion in #884? Which version of MySQL do you use and what is the value of `MYSQL_VERSION_ID` for you?

Thanks, you are absolutely right: I overlooked that patch, as I started from version 4.0.2 (which does not contain 7c431d9 yet).
It broke my Travis CI pipeline on some other component. Maybe time for 4.0.3 then?

vadz:
Yes, we'll need to make 4.0.3 as there were some important fixes in ODBC backend since 4.0.2 too.

But then I'd really like to migrate to C++14 in master...

PR #899, opened 2021-07-22
V4.0.2 cgx

cedricdelpas:
Allow the possibility to have access to lastConnectParameters_ variable in session class


vadz:
Sorry, something strange has happened with your branch, could you please rebase it on master to get rid of all the merge commits?

Also, could you please document the new function somewhere, probably in `docs/connections.md` to make it discoverable?

TIA!

vadz:
As there is still no reply ~2 months later, it's probably not useful to keep this open any longer.

PR #902, opened 2021-08-03
Fixed MySQL backend not categorizing correctly connection errors to 'soci_error'

mathausmendel:
During a connection error the MySQL backend was not checking the error category
and not propagating it as well. It could lead the user to improperly handle
such situations as they would be unaware of this bug.

### DIFF
--- a/include/soci/mysql/soci-mysql.h
+++ b/include/soci/mysql/soci-mysql.h
@@ -32,7 +32,15 @@ class SOCI_MYSQL_DECL mysql_soci_error : public soci_error
 {
 public:
     mysql_soci_error(std::string const & msg, int errNum)
-        : soci_error(msg), err_num_(errNum) {}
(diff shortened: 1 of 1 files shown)


vadz:
Thanks, but this has broken the build as these constants don't seem to be defined in the MySQL headers used on AppVeyour. We probably need to use the raw numbers, with the symbolic names in comments, unless there is some better way to do it.

I also think it could be more clear to use a `switch` rather than a long `if`, but this is just a detail.

vadz:
P.S. Actually it's worse than this because `cat_` used in the code doesn't exist at all. Have you tried building your PR before submitting it? Please do build it locally first.

mathausmendel:
Sorry about pushing non working code. I've reviewed it and fixed it accordingly.

vadz:
Thanks for fixing this! (Squash-)merged now.

PR #906, opened 2021-09-09
Added mutex for secure using soci::session in multi-threaded environment

burluckij:
We use soci library in our big project (distributed file system) in many places for communication with DBs. Last week development team received an interesting core dump of one service component, there was an invalid access error – using invalid pointer (typical case for C\C++ apps), but place where the error occurred is pretty interesting, see stack trace below:
```
#0  0x0000000000bb9dc5 in **soci::session::rollback** (this=<optimized out>) ... /soci/src/core/session.cpp:248
#1  0x0000000000bbf40e in soci::transaction::rollback (this=0x80ab18450)  ... /soci/src/core/transaction.cpp:51
#2  0x00000000007abb00 in GenericDao<AlertDao>::genericPreparedInsert<ComponentError*> ... /mgmt/dblayer/soci/GenericDao.H:329
#3  0x00000000007ab417 in GenericDao<AlertDao>::genericPreparedInsert<ComponentError*> ... /mgmt/dblayer/soci/GenericDao.H:273
#4  0x000000000078f815 in GenericDao<AlertDao>::genericPeriodicRefresh<ComponentError*> ... /mgmt/dblayer/soci/GenericDao.H:643
#5  0x000000000076a3eb in AlertDao::periodicRefresh (alerts=std::vector of length 28 = {...})...cc/dblayer/soci/AlertDao.C:105
#6  0x000000000061e917 in Topology::refreshPostgres (this=0x80210f400) ... cc/base/Topology.C:2387
#7  0x0000000000620293 in Topology::postgresRefreshThread (this=0x80210f400) ... /cc/base/Topology.C:2727
#8  0x00000000006181da in postgres_refresh_thread_wrapper (arg=0x0)
#9  0x0000000001109082 in pan_sys_thread_create_start_proc (arg=0x805976b20)
```
Where _soci::rollback()_ is not thread-safe:
```
1: void session::rollback()
2: {
3:    _ensureConnected_(backEnd_);
4:   backEnd_->rollback();  // _Our service crashed in that certain place!_
5: }
```
image

It is not enough to have only _ensureConnected()_ function which actually verifies passed to it pointer on _NULL_, here we need to have mutex (or any similar lock).

Our service has a lot of threads, lets imagine that one thread called _session::rollback()_ and was interrupted after successful _backEnd__ pointer verification on 3rd line and right after that started another thread which called _session::reconnect()_ or _session::close()_ (it is not important which one, because each of them changes value of _backEnd__ pointer), after some time first thread (earlier interrupted) continued its execution and tried to touch _backEnd__ pointer (4 line) and **ACCESS_VALIDATION** occured.

In that pull request I've used mutex (_was idea to use std::mutex, but found that in soci library it is not used at all_) for simple protection _backEnd__ pointer.

### DIFF
--- a/include/soci/session.h
+++ b/include/soci/session.h
@@ -14,12 +14,41 @@
 #include "soci/connection-parameters.h"
 #include "soci/logger.h"
 
+#ifndef _MSC_VER
+#include <stdint.h>
+#endif
+
 // std
 #include <cstddef>
 #include <memory>
 #include <ostream>
 #include <sstream>
 #include <string>
+
+#ifdef _WIN32
+
+#include <windows.h>
+
+typedef CRITICAL_SECTION soci_mutex_t;
+
+#define LOCK(x) EnterCriticalSection(x)
+#define UNLOCK(x) LeaveCriticalSection(x)
+#define MUTEX_INIT(x) InitializeCriticalSection(x)
+#define MUTEX_DEST(x) DeleteCriticalSection(x)
+
+#else // linux
+
+#include <pthread.h>
+#include <dlfcn.h>
+
+typedef pthread_mutex_t soci_mutex_t;
+
+#define LOCK(x) pthread_mutex_lock(x)
+#define UNLOCK(x) pthread_mutex_unlock(x)
+#define MUTEX_INIT(x) pthread_mutex_init(x, NULL)
+#define MUTEX_DEST(x) pthread_mutex_destroy(x)
+
+#endif // _WIN32
 
 namespace soci
 {
@@ -202,6 +231,7 @@ private:
 
     bool uppercaseColumnNames_;
 
+	soci_mutex_t lock_;
     details::session_backend * backEnd_;
 
     bool gotData_;
--- a/include/soci/session.h
+++ b/include/soci/session.h
@@ -231,7 +231,7 @@ private:
 
     bool uppercaseColumnNames_;
 
-	soci_mutex_t lock_;
+    soci_mutex_t lock_;
     details::session_backend * backEnd_;
 
     bool gotData_;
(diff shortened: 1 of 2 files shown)


vadz:
Thanks, but the reason there is no mutex in this code now is that `soci::session` has never been advertised as being thread-safe. I'm not sure if locking mutex in all its methods is really sufficient to make it thread-safe and it's not without its own costs: this will make the calls slower even for single-threaded applications and might result in deadlocks if there is ever any way to call one session method from inside another one.

Moreover, avoiding crashes is one thing, but how are you going to avoid problems due to accessing database concurrently from multiple threads? E.g. what if one thread inserts something into the database while another one deletes from it. Surely you will have problems if this happens.

So IMO the only good approach is to avoid using the same `session` object from multiple threads in the first place instead of trying to make the session object itself thread-safe and I don't think this PR should be merged, sorry.

burluckij:
> Thanks, but the reason there is no mutex in this code now is that `soci::session` has never been advertised as being thread-safe. I'm not sure if locking mutex in all its methods is really sufficient to make it thread-safe and it's not without its own costs: this will make the calls slower even for single-threaded applications and might result in deadlocks if there is ever any way to call one session method from inside another one.
> 
> Moreover, avoiding crashes is one thing, but how are you going to avoid problems due to accessing database concurrently from multiple threads? E.g. what if one thread inserts something into the database while another one deletes from it. Surely you will have problems if this happens.
> 
> So IMO the only good approach is to avoid using the same `session` object from multiple threads in the first place instead of trying to make the session object itself thread-safe and I don't think this PR should be merged, sorry.

Thanks for quick and informative response.

PR #911, opened 2021-09-29
fix #908: sqlite3 bulk operation can not deal with vector size 1

### LINKED ISSUE #908: sqlite3 bulk operation can not deal with vector size 1
```
  soci::session sql(soci::sqlite3, "/tmp/soci_bulk_test.db");
  sql << R"(
CREATE TABLE TEST(
    ID INT PRIMARY KEY NOT NULL,
    NAME TEXT
)
)";

  std::vector<int> ids;
  ids.emplace_back(1);
  ids.emplace_back(2);
  sql << "INSERT INTO TEST(ID) VALUES(:ID)", soci::use(ids);

  ids = std::vector<int>(1);
  sql << "SELECT ID FROM TEST", soci::into(ids);
  BOOST_TEST(ids.size() == 1);
```
gives: `soci_bulk_test.cpp(36): error: in "soci_bulk_test": check ids.size() == 1 has failed [0 != 1]`</br></br>

530967061:


src/backends/sqlite3/statement.cpp
```diff
@@ -343,13 +344,13 @@ sqlite3_statement_backend::execute(int number)
...
+        if (hasVectorIntoElements_ || number == 0)
         {
-            retVal = load_one();
+            retVal = load_rowset(number);
         }
         else
         {
-            retVal = load_rowset(number);
+            retVal = load_one();
         }
```
vadz, line 354: Shouldn't we just call `fetch()` here if the code is identical in both places now?

vadz (changes requested):
Thanks for the fix, but the new test fails when using SQL Server, probably because all the statement results have to be consumed before the connection is reuse for something else. Could you please try to fix this? TIA!

vadz:
Thanks for the update! Looks good to me now and I'll merge it once all the CI builds pass.

vadz:
Sorry, no idea what has happened here, I've restarted the CI builds, let's see how it goes now. Thanks for the ping.

vadz:
Oops, sorry, I do know what's going on here. I've completely forgotten that we still run several CI builds under Ubuntu 16.04 which is not supported by GitHub any longer, so they can't run any more.

I'll try to fix this a.s.a.p....

Issue #913, opened 2021-10-10
soci::row get throw bad_cast when compiled use clang with c++abi


530967061:
When compile soci use clang with c++abi and 'SOCI_VISIBILITY' ON, then many soci::row dynamic binding test cases failed with std::bad_cast, for example
```
-------------------------------------------------------------------------------
Dynamic row binding
-------------------------------------------------------------------------------
/app/yacbs/softwear/source/soci-4.0.2/tests/common-tests.h:2321
...............................................................................

/app/yacbs/softwear/source/soci-4.0.2/tests/common-tests.h:2365: FAILED:
  CHECK( r.get<std::tm>(3).tm_year == 105 )
due to unexpected exception with message:
  std::bad_cast

/app/yacbs/softwear/source/soci-4.0.2/tests/common-tests.h:2321: FAILED:
  {Unknown expression after the reported line}
due to unexpected exception with message:
  std::bad_cast

-------------------------------------------------------------------------------
Reading rows from rowset
-------------------------------------------------------------------------------
/app/yacbs/softwear/source/soci-4.0.2/tests/common-tests.h:2862
...............................................................................

/app/yacbs/softwear/source/soci-4.0.2/tests/common-tests.h:2862: FAILED:
  {Unknown expression after the reported line}
due to unexpected exception with message:
  std::bad_cast

-------------------------------------------------------------------------------
```
</br>
This is cause by RTTI usage when sharing C++ objects across binary boundaries. 

[Here](https://www.russellmcc.com/posts/2013-08-03-rtti.html) is a basic explanation. </br></br>
Please check it, thanks!

vadz:
Sorry, I'm unlikely to have time to look at this in the near future. If you can check casting to which class exactly fails, it could help us to fix the problem.

530967061:
I used soci_sqlite3_test for test and:</br>
```
ldd soci_sqlite3_test
	linux-vdso.so.1 =>  (0x00007fffaede6000)
	libpthread.so.0 => /lib64/libpthread.so.0 (0x00007f41158a5000)
	libdl.so.2 => /lib64/libdl.so.2 (0x00007f41156a1000)
	libsqlite3.so.0 => /lib64/libsqlite3.so.0 (0x00007f41153ec000)
	libsoci_sqlite3.so.4.0 => /app/yacbs/softwear/source/soci-4.0.2/build/lib/libsoci_sqlite3.so.4.0 (0x00007f4115c9e000)
[...]

PR #918, opened 2021-11-21
fix: #913 soci::row use union type data_holder instead of RTTI

530967061:
Because it is dangerous to use RTTI when sharing C++ objects across binary boundaries, so we'd better to use hard code type instead of RTTI for soci::row. Besides, when we need to change from one database to another, it is difficult to use the correct type for soci::row::get(). So we use boost::numeric_cast<> to do auto type conversion when WITH_BOOST is 'ON'.


src/core/statement.cpp
```diff
@@ -682,28 +644,42 @@ void statement_impl::describe()
...
         {
         case dt_string:
-            bind_into<dt_string>();
-            break;
         case dt_blob:
-            bind_into<dt_string>();
-            break;
         case dt_xml:
-            bind_into<dt_string>();
+        {
```
vadz, line 649: This is pretty minor, but the indentation here is weird, `break` should normally be aligned with the `}`, i.e. I'd indent this block (and all the similar ones below).

src/core/statement.cpp
```diff
@@ -682,28 +644,42 @@ void statement_impl::describe()
...
+            std::pair<std::string*, indicator*> holder = row_->alloc_data_holder_string();
+            exchange_for_row(into(*holder.first, *holder.second));
```
vadz, line 651: Instead of repeating this several times, would it make sense to add some helper function taking this `holder` directly and allowing to write something like this instead:
```suggestion
            exchange_for_row_holder(row_->alloc_data_holder_string());
```
?

include/soci/row.h
```diff
@@ -51,31 +65,25 @@ class SOCI_DECL row
...
     column_properties const& get_properties(std::size_t pos) const;
     column_properties const& get_properties(std::string const& name) const;
 
     template <typename T>
     T get(std::size_t pos) const
     {
         typedef typename type_conversion<T>::base_type base_type;
-        base_type const& baseVal = holders_.at(pos)->get<base_type>();
+        base_type baseVal;
+        get_(pos, baseVal);
```
vadz, line 76: Could we please rename this `get_` to `get_at()` or `do_get()` or something else, but to without `_` at the end of the name? As it is, this is a bit confusing, as this naming convention is used for the member variables in SOCI, but this one is a function, not a variable.

include/soci/row.h
```diff
@@ -126,9 +134,77 @@ class SOCI_DECL row
...
+    void get_(std::size_t pos, std::string &baseVal) const;
+    void get_(std::size_t pos, std::tm &baseVal) const;
+    template <typename T>
+    typename soci::enable_if<std::numeric_limits<T>::is_specialized, void>::type
+    get_(std::size_t pos, T &baseVal) const
+    {
+        return get_number(pos, baseVal, numeric_cast_t<T>());
+    }
+    template <typename T, typename NumericCast>
+    void get_number(std::size_t pos, T &baseVal, const NumericCast &cast) const
```
vadz, line 185: It looks like this function is only useful if we're using Boost, as otherwise it never performs any non-trivial casts? If this is correct, I don't know if it's a good idea and I'd rather never cast between the different numbers rather than only doing it when Boost is available but not without it.

Does it already work like this in the current version of SOCI? If so, I had no idea about it...

src/core/row.cpp
```diff
@@ -109,3 +137,156 @@ std::size_t row::find_column(std::string const &name) const
...
+        holder->type = dt_string;
+        holder->data.p = s;
+        holder->ind = i_ok;
+        holders_.push_back(holder);
+        return std::make_pair(s, &holders_.back()->ind);
+    } catch (...) {
+        delete holder;
+        delete s;
+        throw;
+    }
```
vadz, line 157: And this (and all the other similar blocks below) should be removed once the pointers above are replaced with `auto_ptr`.

OTOH, to be honest, I don't really see what exceptions could possibly happen here anyhow (`bad_alloc`? this one is not really recoverable from in any case...), so I'm not even sure why was this done in the first place?

src/core/row.cpp
```diff
@@ -109,3 +137,156 @@ std::size_t row::find_column(std::string const &name) const
 
     return it->second;
 }
+
+std::pair<std::string*, indicator*> row::alloc_data_holder_string()
+{
+    data_holder *holder = NULL;
+    std::string *s = new std::string();
```
vadz, line 144: This should be replaced with
```suggestion
    cxx_details::auto_ptr<data_holder> holder(new data_holder());
    cxx_details::auto_ptr<std::string> s(new std::string());
```
see `include/soci/soci-platform.h` and then `holder.release` and `s.release()` should be used below to avoid managing memory manually.

src/core/row.cpp
```diff
@@ -16,6 +16,26 @@
 using namespace soci;
 using namespace details;
 
+namespace soci
+{
+    namespace details
+    {
+        struct data_holder
+        {
```
vadz, line 24: Why not give this struct a ctor (taking `type` value and initializing `ind` to `i_ok`) and dtor (cleaning up `data.p` for strings and times)? It looks like it could simplify the code quite a bit.

vadz (commented):
Thanks for your work on this!

Globally it looks good, and it looks like it ought indeed solve the original problem, but it would be nice if it could be improved a bit, notably these `catch(...) { throw }` clauses are really not very nice.

Could you please try to do it? TIA!

vadz:
The CodeQL CI failure is unrelated to the changes here, but the other ones need to be fixed before this PR can be merged:

- For the warning in the Valgrind job, please use `Approx()`, see the existing example of its use in `tests/sqlite3/test-sqlite3.cpp` or CATCH documentation.
- For MSVS warnings in the AppVeyor build, you may find `SOCI_DUMMY_RETURN()` useful (but I didn't look at them in the details).
 
TIA!

include/soci/row.h
530967061: I have tried to implement numeric_cast<> use only c++ standard library. But the compiler warnings are not easy to fix. Although  SOCI::into doesn't suffer the number conversion problem, I should use SOCI::into instead, but ORM depends on SOCI::row and some functions in different databases generate different data types. So soci::row with auto numeric conversion is very implement to me. I also know the difference between WITH_BOOST 'ON' and WITH_BOOST 'OFF'  is ugly, but I have not found a better way, so I use number_cast just for BOOST 'ON' and leave numeric_cast<> for BOOST 'OFF'  the future work.<br/>

Do you have any good ideas for me? Because number_cast<> is only in row.h, So I think remove number_cast from row.h and patch this only when I use it in my project may be a compromise way.

530967061:
I have no idea why MSVC 2015 gives warning 'unreachable code', so I have only fixed other problems. Please help to see it again, thanks.

vadz:
Sorry, I don't think it's worth making all these changes just to fix #913, as there is a much simpler solution there, please see #975.

If it's worth returning to this because of some other considerations (performance?), please let me know but otherwise I don't think it's needed.

Issue #922, opened 2021-12-26
soci::RowSet support for social::Blob


schvarcz:
Hello everyone.

First, thanks for this very useful library. We are really looking to use this library on our company due the support to several backends.

However, I see some room for improvement where I hope I can help.

On our case, we have heavily use the `soci::blob`. Right now, I am in a situation where I have to use it with `soci::rowset`. However, when I do something like `m_rowset.get<blob>("m_blob")`, `soci` returns the error:
```
/usr/local/include/soci/row.h:70:11: error: no matching function for call to ‘soci::blob::blob()’
         T ret;
           ^~~
```

I do understand the nature of the problem and why `soci::blob` needs to be initialised with `soci::session`. I also can see that similar [issues](https://github.com/SOCI/soci/issues/476) were raised in the past, with the same nature.

Since I can't see another way to retrieve `soci::blob` objects in a bulk way, I had to dig into it.

I believe that we have 3 main points to check when we are talking about retrieving information with blob:
1. soci::rowset
2. soci::into
3. ORM 

For the first case, the solution seems to rely direct on the `core`, since `soci::rowset` [has a pointer](https://github.com/SOCI/soci/blob/b976b3cbe3df050d537c9fcfdda28371cdc60762/include/soci/rowset.h#L151) to `soci:statement`; and `soci::statement` [has a public pointer](https://github.com/SOCI/soci/blob/b976b3cbe3df050d537c9fcfdda28371cdc60762/include/soci/statement.h#L87) to `soci::session`.

If we can rearrange `soci::rowset` and `soci::row` so the `soci::session` pointer can reach the final class,  we might solve the issue for the first case. (Ps: we also should be careful for the cases of `soci::rowset<soci::blob>`.

**I would like to know what do you think about this solution.** It seems to me that we are supposed to have a method to retrieve blob data in a bulk manner, since it is available to all other types and it seems to be a recurrent issue report here. **If you agree with this implementation, I can work on that this week and submit a PR. Otherwise, I am open to other solutions.**

Very likely I will dig into the case 2. in the near future.

Kind regards

schvarcz:
Hello guys.

@vadz answered me on the PR:

> Thanks for working on this!
> 
[...]

PR #924, opened 2021-12-30
soci::RowSet support for soci::Blob 

schvarcz:
As I have suggested [here](https://github.com/SOCI/soci/issues/922) currently it seems impossible to use `rowset` to loop into a query result containing `blob` fields.

This pull request aims to provide such functionality as described on the reported issue.

I am creating the PR in advance any response there to make use of your CI and test to all supported backends.

Kind regards.

### DIFF
--- a/include/soci/blob.h
+++ b/include/soci/blob.h
@@ -9,9 +9,9 @@
 #define SOCI_BLOB_H_INCLUDED
 
 #include "soci/soci-platform.h"
-#include "soci/type-holder.h"
 // std
 #include <cstddef>
+#include <vector>
 
 namespace soci
 {
@@ -27,36 +27,57 @@ class blob_backend;
 class SOCI_DECL blob
 {
 public:
+    blob();
+
+    // Keeping it for back compatibility
+    // [[deprecated]]
     explicit blob(session & s);
+
     blob(const blob& b) { this->backEnd_ = b.backEnd_; }
     ~blob();
 
-    void assign(details::holder* h);
     std::size_t get_len();
+
+    std::size_t size() { return data_.size(); }
+    void resize(std::size_t new_size) { data_.resize(new_size); }
+    char& operator[](int& i) { return data_[i]; }
+    const char& operator[](int& i) const { return data_[i]; }
+    char& operator[](std::size_t i) { return data_[i]; }
+    const char& operator[](std::size_t i) const { return data_[i]; }
 
     // offset is backend-specific
     std::size_t read(std::size_t offset, char * buf, std::size_t toRead);
 
     // offset starts from 0
-    std::size_t read_from_start(char * buf, std::size_t toRead,
-        std::size_t offset = 0);
+    inline std::size_t read_from_start(char * buf, std::size_t toRead,
+        std::size_t offset = 0)
+    {
+        return this->read(offset, buf, toRead);
+    }
 
     // offset is backend-specific
     std::size_t write(std::size_t offset, char const * buf,
         std::size_t toWrite);
 
-    // offset starts from 0
-    std::size_t write_from_start(const char * buf, std::size_t toWrite,
-        std::size_t offset = 0);
+    inline std::size_t write_from_start(const char * buf, std::size_t toWrite,
+        std::size_t offset = 0)
+    {
+        return this->write(offset, buf, toWrite);
+    }
 
     std::size_t append(char const * buf, std::size_t toWrite);
 
     void trim(std::size_t newLen);
 
-    cxx_details::shared_ptr<details::blob_backend> get_backend() { return backEnd_; }
+    // Keeping it for back compatibility
+    // [[deprecated]]
+    details::blob_backend* get_backend() { return backEnd_; }
 
 private:
-    cxx_details::shared_ptr<details::blob_backend> backEnd_;
+    details::blob_backend* backEnd_;
+
+    // buffer for BLOB data
+    std::vector<char> data_;
 };
 
 } // namespace soci
--- a/include/soci/blob.h
+++ b/include/soci/blob.h
@@ -9,6 +9,7 @@
 #define SOCI_BLOB_H_INCLUDED
 
 #include "soci/soci-platform.h"
+#include "soci/type-holder.h"
 // std
 #include <cstddef>
 
@@ -30,6 +31,7 @@ public:
     blob(const blob& b) { this->backEnd_ = b.backEnd_; }
     ~blob();
 
+    void assign(details::holder* h);
     std::size_t get_len();
 
     // offset is backend-specific
--- a/include/soci/blob.h
+++ b/include/soci/blob.h
@@ -27,6 +27,7 @@ class SOCI_DECL blob
 {
 public:
     explicit blob(session & s);
+    blob(const blob& b) { this->backEnd_ = b.backEnd_; }
     ~blob();
 
     std::size_t get_len();
@@ -50,10 +51,10 @@ public:
 
     void trim(std::size_t newLen);
 
-    details::blob_backend * get_backend() { return backEnd_; }
+    cxx_details::shared_ptr<details::blob_backend> get_backend() { return backEnd_; }
 
 private:
-    details::blob_backend * backEnd_;
+    cxx_details::shared_ptr<details::blob_backend> backEnd_;
 };
 
 } // namespace soci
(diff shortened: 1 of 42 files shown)


tests/sqlite3/test-sqlite3.cpp
```diff
@@ -105,6 +105,35 @@ TEST_CASE("SQLite blob", "[sqlite][blob]")
...
+        sql << "insert into soci_test(id, img) values(4, ?)", use(b);
+
+    }
+    {
+        rowset<row> rs = (sql.prepare << "select * from soci_test");
+        for(rowset<row>::iterator rsit = rs.begin(); rsit != rs.end(); rsit++)
+        {
+            // row &r = *rsit;
+            // soci::blob b = r.get<soci::blob>("img");
+            // CHECK(b.get_len() == sizeof(buf));
```
vadz, line 132: Sorry, why is this commented out? The test doesn't seem to really test much as currently written.

tests/sqlite3/test-sqlite3.cpp
```diff
@@ -105,6 +105,35 @@ TEST_CASE("SQLite blob", "[sqlite][blob]")
     }
 }
 
+TEST_CASE("SQLite blob on a rowset", "[sqlite][blob][rowset]")
```
vadz, line 108: Why should this be SQLite-specific? I think this should work for all backends, are there any known problems with this?

include/soci/blob.h
```diff
@@ -50,10 +51,10 @@ class SOCI_DECL blob
 
     void trim(std::size_t newLen);
 
-    details::blob_backend * get_backend() { return backEnd_; }
+    cxx_details::shared_ptr<details::blob_backend> get_backend() { return backEnd_; }
```
vadz, line 54: Do we really have to change this? It's a backwards-incompatible change and it seems like it could be avoided by keeping the existing return type here and adding some new `get_backend_ptr()` if it's needed.

include/soci/row.h
```diff
@@ -41,6 +42,7 @@ class SOCI_DECL row
 {
 public:
     row();
+    row(session *s);
```
vadz, line 45:
```suggestion
    explicit row(session *s);
```

include/soci/row.h
```diff
@@ -51,6 +53,9 @@ class SOCI_DECL row
     indicator get_indicator(std::size_t pos) const;
     indicator get_indicator(std::string const& name) const;
 
+    void set_session(session *s) { session_ = s; };
```
vadz, line 56: Having both a ctor taking `session*` and the default ctor and a setter for session seems strange, could we avoid having either the first or the last one of them?

include/soci/soci-platform.h
```diff
@@ -124,6 +124,119 @@ namespace cxx_details
     using std::auto_ptr;
 #endif
 
+#if defined(SOCI_HAVE_CXX11) || (defined(_MSC_VER) && _MSC_VER >= 1800)
```
vadz, line 127: I think we should just start using C++11 (or, better, C++14) in SOCI master, see [this thread](https://sourceforge.net/p/soci/mailman/soci-devel/thread/E1lYsSp-0000NS-0C%40smtp.tt-solutions.com/#msg37265445).

OTOH I don't know if we need MT-safety provided by `std::shared_ptr` here, perhaps we should always use our simpler (and potentially more efficient) class instead?

At any rate, it would be better to have just version of it.

include/soci/soci-platform.h
```diff
@@ -124,6 +124,119 @@ namespace cxx_details
...
+    inline shared_ptr<T> dynamic_pointer_cast(shared_ptr<U> const& r)
+    { return std::dynamic_pointer_cast<T>(r); }
+
+    template<class T, class U>
+    inline shared_ptr<T> const_pointer_cast(shared_ptr<U> const& r)
+    { return std::const_pointer_cast<T>(r); }
+
+#else // std::shared_ptr<> not available
+    template <typename T>
+    class shared_ptr
```
vadz, line 146: If we do decide to use our own shared pointer class, we could make it intrusive, which would be both more efficient and a bit simpler.

include/soci/type-conversion-traits.h
```diff
@@ -36,6 +37,32 @@ struct type_conversion
...
+    }
+
+    static void to_base(blob & in, base_type & out, indicator & ind)
+    {
+        char *tmp = new char[in.get_len()];
+        out.resize(in.get_len());
+        in.read(0, tmp, in.get_len());
+        
+        out = tmp;
+        ind = i_null;
```
vadz, line 62: Shouldn't this be `i_ok`?

vadz (commented):
Thanks for working on this!

I don't use rowset API myself, so I'm not too familiar with it and could be missing something here, but I'm a bit surprised by the need to modify the code not clearly related to rowsets to implement support for blobs there when they do work just fine with just `use()` and `into()` already (which I do use in my own code). It would be best if this could be avoided, i.e. if the backend files could remain unmodified.

Also, thank you for adding a test, but ideally the test should work for all backends, not just SQLite, and actually test that retrieving the blobs works, too.

Thanks again!

schvarcz:
Hello @vadz, thank you for your reply here!

I will answer you on the issue itself, since the debate might be longer than a simple fix on this code.

I created the PR here just to test on VS for windows, since I have no means to test it by my own here. I hope it is ok for you.

All your comments will be answered on the [issue 922](https://github.com/SOCI/soci/issues/922). 

If it is ok for you, I will mark you here again when the PR is actually ready for revision.

Thank you!

---------------

***OBS 1:*** If you wanna follow the CI tests, I am running them [here](https://github.com/IWFGA/soci/pull/1). This repo has a good CI pipeline implemented for GitHub Actions, idk why it is not running on my PR here. So I forked the project to run it there.
***OBS 2:*** Just so you know, I wanna work on the final backend (Oracle) tonight or tomorrow. If everything goes fine, a PR will be ready to evaluation on Monday's morning.

include/soci/blob.h
schvarcz: It is a good suggestion indeed.

If we really need to keep this around, I will have this in mind. However, if we normalize the use of `SOCI` for `blob` data, as explained on the [issue 922](https://github.com/SOCI/soci/issues/922), and separate `class blob;` from `class blob_backend;` neither the `shared_ptr` neither this function will be really necessary anymore.

include/soci/row.h
schvarcz: I think it is possible. I will check.

include/soci/soci-platform.h
schvarcz: Yes, please!

Since I am new around here and not a core maintainer, I would even talk about C++17. Since I am young, passionate about the new, and adventurous!

But very happy with C++11.

include/soci/type-conversion-traits.h
schvarcz: Yes, tks for the observation. I was just checking what this would cause. I think I forgot to change back.

tests/sqlite3/test-sqlite3.cpp
schvarcz: Yes, it should and I am working on that (more info on the comments below here or on the issue 922).

At the time I pushed it here, it was working only on SQLite3. My goal was to have the VS integration compiling before moving further.

Talking about tests, what do you think about centralising the tests that should work on every backend? I would happily do it next, if you think it is appropriate.

tests/sqlite3/test-sqlite3.cpp
schvarcz: My goal was not really testing here, just check if it was still compiling for VS.

I am sorry if it was not clear enough.

include/soci/soci-platform.h
schvarcz: I am not sure if I understand what you mean by "intrusive" here.

Anyways, I would like to highlight just two points before we go deeper: 

1.    currently I am using this class on 2 different points where they are really needed (could not use `unique_ptr` or `auto_ptr`). 
2.    If we solve the confusion between `class blob;` and `class blob_backend;` as reported on the issue 922, I will likely not need this class anymore.

vadz:
I'm unsure what to do about this one, almost a year has passed and I don't think it's still useful to keep it open, even though it's also a pity to close it because a lot of work has been done here.

But, unfortunately, I don't see what else can be done in the current state of this PR neither...

Krzmbrzl:
@vadz I'll include this in my PR #992 and see if I can salvage any work done here already. I need this functionality anyway and since this PR has stalled it's probably easier to simply include the feature in #992 than wrapping my head around this PR and then rebasing it on top of #992.

vadz:
Thanks for taking over it, I'll close this one then and wait until #992 stabilizes.

PR #925, opened 2022-01-12
Add documentation for using SOCI as a CMake subdirectory

### LINKED ISSUE #792: SOCI CMake Sub Project help.
I need Some Hellp about instruction how to use SOCI from cmake as subporject "add_subdirectory(SOCI)"
I have long time try use It but SOCI::* not work. I know this about CMake but no one try to help. And How I can do in way from SOCI CMake script Variables. How I can do that ?

zann1x:
Fixes #792.

As explained [in my answer](https://github.com/SOCI/soci/issues/792#issuecomment-1008140358) to issue #792, placing SOCI in a subdirectory of your project and simply including it with CMake is possible. This PR adds documentation how to do it.

I haven't used SOCI extensively enough to say whether both static and dynamic linking is possible this way. In my tests, I was only able to get the statically linked library running.


vadz:
Nice, thanks! I can merge it like this but I wonder if we could (should?) build this example as part of the CI builds to make sure that it not only works now, but also continues working in the future? If you can modify `appveyor.yml` and/or `.github/workflows/ci.yml` to do it, it would be great!

zann1x:
Sure thing :) I tried to add it as non-invasive as possible.

vadz:
Looks great, thanks! Will merge in a moment.

PR #927, opened 2022-01-19
proper implementation of the get_table_names for MySQL

dmarkh:
Currently, the default implementation of the get_table_names method returns nothing as it seems to use a Postgres-specific query. I have added a MySQL-specific query that returns a list of table names for the selected database.


vadz:
Thanks! Will merge this once the CI builds finish running.

PR #935, opened 2022-02-15
Use correct option names

Krzmbrzl:
The documentation was referring to non-existing WITH_* options when in fact
they are (now) called SOCI_*

Source:
https://github.com/SOCI/soci/blob/175c716f2b0b78ca3c907aae60605d0a65127beb/cmake/SociBackend.cmake#L80-L82

### DIFF
--- a/docs/installation.md
+++ b/docs/installation.md
@@ -112,7 +112,7 @@ Some other build options:
 
 #### IBM DB2
 
-* `WITH_DB2` - boolean - Should CMake try to detect IBM DB2 Call Level Interface (CLI) library.
+* `SOCI_DB2` - boolean - Should CMake try to detect IBM DB2 Call Level Interface (CLI) library.
 * `DB2_INCLUDE_DIR` - string - Path to DB2 CLI include directories where CMake should look for `sqlcli1.h` header.
 * `DB2_LIBRARIES` - string - Full paths to  `db2` or `db2api` libraries to link SOCI against to enable the backend support.
 * `SOCI_DB2` - boolean - Requests to build [DB2](backends/db2.md) backend. Automatically switched on, if `WITH_DB2` is set to ON.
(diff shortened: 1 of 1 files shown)


vadz:
I'm not sure if this is correct because `-DWITH_XXX=OFF` definitely works, e.g. I get output like this:
```
-- MySQL: disabled, since WITH_MYSQL=OFF
-- ODBC: disabled, since WITH_ODBC=OFF
-- Oracle: disabled, since WITH_ORACLE=OFF
-- PostgreSQL:
-- POSTGRESQL_INCLUDE_DIRS                  = /usr/include/postgresql
-- POSTGRESQL_LIBRARIES                     = /usr/lib/x86_64-linux-gnu/libpq.so
-- POSTGRESQL_VERSION                       = 11.14 (Debian 11.14-0+deb10u1)
-- SQLite3: disabled, since WITH_SQLITE3=OFF
-- Firebird: disabled, since WITH_FIREBIRD=OFF
-- DB2: disabled, since WITH_DB2=OFF
```
if I use `-DWITH_{DB2,FIREBIRD,MYSQL,ODBC,ORACLE,SQLITE3}=OFF`, so it looks like these options exist (probably provided by CMake standard macros?).

What problem exactly does this fix?

Krzmbrzl:
Interesting - for me the WITH_* macros didn't do anything 👀

vadz:
What exactly did you try, i.e. what didn't work for you?

Krzmbrzl:
Okay I found
https://github.com/SOCI/soci/blob/b8560cca413a370fc7019d18908d5612a8066fe8/cmake/SociDependencies.cmake#L74-L84

Thus it appears as if the `WITH_` macros are a way to prevent SOCI from locating the necessary backends. Howver, that doesn't appear to be the front-end interface intended for users. Especially since the `SOCI_` macros have the explicit comment
https://github.com/SOCI/soci/blob/b8560cca413a370fc7019d18908d5612a8066fe8/cmake/SociBackend.cmake#L79-L82

vadz:
I guess we should document both then? I'm a bit reluctant to remove the current options from the docs because they seem to have existed for a long time and, at least IME, work just fine.

Krzmbrzl:
🤷

Your call :)

vadz:
Sorry, I had somehow completely failed to see that `SOCI_$BACKEND` variables are *already* documented and the documentation even explains their relationship to `WITH_$BACKEND` ones, e.g.
```
* `WITH_FIREBIRD` - boolean - Should CMake try to detect Firebird client library.
...
* `SOCI_FIREBIRD` - boolean - Requests to build [Firebird](backends/firebird.md) backend. Automatically switched on, if `WITH_FIREBIRD` is set to ON.
```
so there is really nothing to do here.

Krzmbrzl:
Oops, seems I missed that too 👀

PR #941, opened 2022-03-22
Fix possible segfault and missing FAIL(...) call in unit test "Insert error"

zann1x:
The unit test "Insert error" contains an incorrect break condition in the for loop. Checking for the validity of the pointer address always results in a truthy value, possibly walking the whole `names` array. When trying to assign the referenced NULL value to a std::string, this could result in a segfault.

Also, a call to FAIL(...) was missing in case the expected exception did not occur in the test.


vadz:
Thanks for the fixes, will merge in a moment!

PR #947, opened 2022-04-01
DOCS: Fix used varname in example

Krzmbrzl:


Krzmbrzl:
Additionally, I think it might be nice, if that page could also give an example of how to exchange multiple parameter. Is that a simple matter of
```
auto stmt = sql.prepare = "INSERT INTO dummy(a,b) VALUES(:a, :b)";
stmt.exchange(soci::use(var1));
stmt.exchange(soci::use(var2));
```
?

vadz:
Thanks for fixing the typos, will merge in a moment.

`exchange()` indeed works like this, either with `use()` or with `into()`.

PR #950, opened 2022-04-02
Transactions

sirlordt:
Hello there.

session and transaction objects are linked. Is possible known from session object the transaction associated and is it active. And from transaction object the session owner.

Fix 1.


vadz:
Sorry, I don't understand what is the goal of this PR. Could you please explain whether this is fixing a problem or adding a new feature? If the former, what is the problem exactly, i.e. how can it be reproduced? If the latter, what is the feature for?

PR #953, opened 2022-04-03
Add transaction::is_active() and current_session()

sirlordt:
Added two new getters methods to transaction object

transaction::is_active();        //Allow check if transaction is alive and prevent throw a c++ exception
transaction::current_session();  //Allow to get the db session from the current transaction. Very usefull to generate a new transaction if is commited.

```cpp
//Example

void process_a( soci::transaction * trans ) 
{
  soci::transaction * transaction_ = trans;

  bool is_local_transaction = false;

  try 
  {
      if ( transaction_->is_active() == false ) //The transaction is not alive
      {
          transaction_ = new soci::transaction( trans.current_session() ); //Generate new one

          is_local_transaction = true;
      }

      //Process data in db

      //...
      
      if ( is_local_transaction == true ) //Only apply if is local transaction
      {
          transaction_->commit(); //Apply the local transaction 
      }
  }
  catch ( ... ) 
  {
      if ( transaction_->is_active() &&  //We save another nested try / catch
           is_local_transaction == true )  
      {
          transaction_->rollback();
      }
  }

  if ( is_local_transaction ) 
  {
      delete transaction_;
  }
}

void process_b( soci::transaction * trans ) 
{
  soci::transaction * transaction_ = trans;

  bool is_local_transaction = false;

  try 
  {
      if ( transaction_->is_active() == false ) //The transaction is not alive
      {
          transaction_ = new soci::transaction( trans->current_session() ); //Generate new one

          is_local_transaction = true;
      }

      //Process data in db

      process_a( transaction_ ); //Pass the transaction. But process_a not apply the transaction because is still alive

      //More db stuff

      //...
      
      if ( is_local_transaction == true ) //Only apply if is local transaction
      {
          transaction_->commit(); //Apply the local transaction 
      }
  }
  catch ( ... ) 
  {
      if ( transaction_->is_active() &&  //We save another nested try / catch
           is_local_transaction == true )  
      {
          transaction_->rollback();
      }
  }

  if ( is_local_transaction ) 
  {
      delete transaction_;
  }
}

int main( int argv, char *argc[] ) {

  soci::session db_connection = ...;

  soci::transaction trans( db_connection );

  //process_a not apply the transaction, is not local created
  process_a( &trans ); 

  trans->commit(); //Now the transaction is not active

  //Still work. process_a generate a local transaction by self, 
  //and apply your local generated transaction
  process_a( &trans ); 

  //Still work. process_b generate a local transaction by self, 
  //and pass to nested call in process_a, but process_a not apply that. 
  //process_b must apply your local generated transaction
  process_b( &trans ); 

}

```

### DIFF
--- a/include/soci/transaction.h
+++ b/include/soci/transaction.h
@@ -24,7 +24,12 @@ public:
     void commit();
     void rollback();
 
+    inline session& current_session() const;
+
+    bool is_active() const;
+
 private:
+
     bool handled_;
     session& sql_;
 
--- a/src/core/transaction.cpp
+++ b/src/core/transaction.cpp
@@ -51,3 +51,13 @@ void transaction::rollback()
     sql_.rollback();
     handled_ = true;
 }
+
+inline session& transaction::current_session() const
+{
+    return sql_;
+}
+
+bool transaction::is_active() const
+{
+    return this->handled_ == false;
+}


sirlordt:
Hello @vadz. Are you there?

vadz:
Sorry, I've somehow missed this one.

Returning to it now, I understand what this does, but still not why do we need. The logic of the example is quite convoluted and I just don't see why would we want to encourage writing code like this.

If you really need it, it looks like you could easily just wrap `transaction` in your own class or even use the `session` transaction-related functions directly. But there doesn't seem to be any need for them in the SOCI class itself.

Please let me know what I am missing if you disagree.

sirlordt:
thanks for the feedback.

Is not possible wrap soci transaction class, because the data i need access is private and not protected. Before need change to protected section.

Issue #926, opened 2022-01-18
Bug: Bad bounds check when retrieving a uint32_t value into an optional


seelabs:
If a `uint32_t` with a value of `numeric_limits<std::uint32_t>::max();` is stored in a database and then retrieved into a variable of type `optional<uint32_t>` an exception is thrown:
```
        Value outside of allowed range for the parameter number 1
```
Note that retrieving the value into a variable of type `std::uint32_t` works correctly. The following small program demonstrates the behavior:

```
#define SOCI_USE_BOOST

#include <soci/soci.h>
#include <soci/sqlite3/soci-sqlite3.h>

#include <boost/optional.hpp>

#include <cassert>
#include <iostream>

void
testSQLiteSelect()
{
    std::uint32_t const uid = std::numeric_limits<std::uint32_t>::max();

    {
        soci::backend_factory const& backEnd = *soci::factory_sqlite3();
        soci::session s(backEnd, "db_test.sqlite3");

        s << "CREATE TABLE T (UI INTEGER UNSIGNED);";
        s << "INSERT INTO T (UI) VALUES(:idu);", soci::use(uid);

        // Retrieving an std::uint32_t::max value into a std::uint_t variable
        // works.
        try
        {
            std::uint32_t uig = 0;
            s << "SELECT UI from T;", soci::into(uig);
            assert(uig == uid);
        }
        catch (std::exception& e)
        {
            std::cerr << "Got exception for std::uint32_t: " << e.what();
            assert(0);
        }

        // Retrieving an std::uint32_t::max value into a
        // boost::optional<std::uint_t> variable throws an exception:
        // "Value outside of allowed range for the parameter number 1"
        try
        {
            boost::optional<std::uint32_t> uig;
            s << "SELECT UI from T;", soci::into(uig);
        }
        catch (std::exception& e)
        {
            std::cerr << "Got exception for optional<std::uint32_t>: "
                      << e.what();
            assert(0);
        }
    }
}

int
main()
{
    testSQLiteSelect();
    return 0;
}
```

Issue #920, opened 2021-12-05
Lack of support for unsigned integers


szymonwieloch:
soci fails to make correct queries with unsigned integers. In the tested scenario, we had a table with a column of unsigned integer type:

```sql
CREATE TABLE 'users' (
    'userid' INT(11) UNSIGNED NOT NULL
) ENGINE InnoDB
```

Both insert and select queries fail:

### Insert

```c++
uint32_t userid = 3000000000;
...
use(userid)
```
soci internally converts the number into a signed integer - then it becomes negative. Inserting a negative value into an unsigned column obviously fails.

### Select

```c++
uint32_t userid;
...
into(userid);
```

This works fine as long the number in the database is small enough to fit into the signed int32_t. Once it is too big (i.e. 3000000000) it throws an exception with a parameter conversion error.

The problems have been found with the latest mariadb. I am not sure if this is a general soci problem or just a problem with MySQL backend.

vadz:
Sorry, I can only confirm that this is a known issue. Generally speaking, C++ numeric types don't really map 1-to-1 to the database types and something needs to be done about it, but it's not totally clear what nor when/who will have time to do it.

zann1x:
Regarding the "what to do", I guess looking at what [poco](https://github.com/pocoproject/poco/tree/master/Data) does could be a step in the right direction as it also has a very similar API to SOCI.
But that's only my two cents.

willw-git:
For the benefit of fellow Googlers who arrived here seeking a fix for this problem:

I have inherited code which just encountered this bug, again with a Maria back end, and Microsoft Visual C++ 2017.  An unsigned 32-bit key just exceeded the signed positive integer limit. 

I have hacked a work-around as follows:

- To retrieve values, I read them into a signed 32-bit int and static_cast them to unsigned.
- To write values, I passed signed 32-bit values and ANDed them with 0xffffffff in the SQL expression. This promotes the input to 64 bits and then removes the sign extension in one hit.

Ugly, but it got the application back on its feet.

Issue #90, opened 2013-03-04
unexpected behavior in soci::row::get for unsigned and long types


ricardofandrade:
Hey guys,

Is there any special reason to force all `unsigned` types to be `long long` while getting a value from a row?

I expected `unsigned char`, `unsigned short` to be an `int` and `unsigned int` to be `long long`.

Besides that, there's no type traits support for neither `long` nor `unsigned long`.

I saw how you select types for exchange and `long` is there, including the support for both 4-byte and 8-byte `long` variations.

Do you see any problem in changing/improving that?

Thanks,
Ricardo

Cleroth:
Similarly, I'm not able to use `unsigned long` to insert data (I have to manually cast it to a `unsigned long long`).

Cleroth:
Got excited when I saw 4.0 is released. Except... there is still no change in something as basic as this.

PR #954, opened 2022-04-07
Complete support for C++ integer types

zann1x:
As often discussed (e.g. #90, #920, #926), SOCI doesn't have a complete support for all C++ integer types. This PR aims to fix this.

To remove ambiguity and confusion in cross-platform cases I switched from usage of `short`, `int`, `long` etc. to the fixed-width integer types `int16_t`, `int32_t` etc. which are part of the type support library as of C++11. Each of those types got its corresponding value in the `exchange_type` and `data_type` enum. This also meant renaming the enumerators to x_[u]intN and dt_[u]intN (e.g. `x_uint32` and `dt_uint32`).
The mapping of the newly introduced types `int8_t`, `uint8_t`, `int16_t`, `uint16_t` and `uint32_t` to database column types was done to the best of my knowledge and can surely be judged better by someone working intensely with the specific database backends.

I also added additional test cases in `common-tests.h` to test the fixed-width types in general as well as their min and max boundaries. All existing tests should stay pretty much untouched and should still pass when this PR is finished.

The commits are divided into the needed changes per backend and tests. All commit messages not starting with "WIP: ..." are more or less done and ready for review.


vadz:
Thanks a lot for working on this! It will take me some time to review this (and the CI failures will need to be fixed, of course), but I'll try to get to it a.s.a.p.

zann1x:
A big part is still work in progress anyways, so take your time :)

zann1x:
Okay, I think I'm finally done now.
For the most part, the API stayed the same. Some database specifics (e.g. MS SQL not supporting unsigned types) actually get noticed now (and might throw errors) and the naming of the exposed `dt_*` types changed (as described above). So that might require changes in some of the user's code.
Other than that all previously existing tests still pass. So generally speaking, SOCI still behaves the same as before.

vadz:
Thanks for your work on this, but this is going to take some time to review, there are really many changes here.

It's a bit surprising that there are no changes at all to the docs, I hoped to start by reading them to have a precise idea of what changed. It would be great if you could please update the documentation to describe the new behaviour there. TIA!

zann1x:
Updating the docs just kind of fell of the table, I'm sorry! I updated them now to reflect the changes.

zann1x:
I rebased my branch on the current master.
Is there any progress on the review of this PR?

vadz:
Thanks for your work on this, I'll try to finally get to it soon, starting from testing my own code using SOCI with this branch.

include/soci/soci-backend.h
```diff
@@ -22,8 +22,19 @@ namespace soci
 // data types, as seen by the user
 enum data_type
 {
-    dt_string, dt_date, dt_double, dt_integer, dt_long_long, dt_unsigned_long_long,
```
vadz, line 25: Sorry, this is a major problem IMO: we need to keep `dt_integer` and `dt_[unsigned_]long_long` names as removing them would break existing code using SOCI (see e.g. [this project](https://cs.github.com/rstudio/rstudio/blob/2a3eead88a128805c249a6aa7a3dbff8e9ae9faa/src/cpp/session/modules/zotero/ZoteroCollectionsLocal.cpp#L88) for a random example).

For `dt_long_long` it's pretty simple as it can be just a synonym for `dt_int64`. For `dt_integer` it might be more complicated, but it needs to be mapped to `dt_int32` or `dt_int64` depending on the actual `sizeof(int)`. But it still needs to exist and continue to work.

include/soci/soci-backend.h
```diff
@@ -40,10 +51,14 @@ enum exchange_type
 {
     x_char,
     x_stdstring,
-    x_short,
-    x_integer,
-    x_long_long,
-    x_unsigned_long_long,
```
vadz, line 46: Same as above, we need to keep the old names.

include/private/firebird/common.h
```diff
@@ -16,6 +16,7 @@
 #include <limits>
 #include <sstream>
 #include <iomanip>
+#include <stdint.h>
```
vadz, line 19: It would be more consistent to include `<cstdint>` (here and elsewhere).

tests/sqlite3/test-sqlite3.cpp
```diff
@@ -664,6 +665,11 @@ class test_context : public test_context_base
         return true;
     }
 
+    bool has_full_uint64_support() const SOCI_OVERRIDE
+    {
+        return false;
```
vadz, line 670: This is rather unexpected, what's wrong with SQLite uint64 support?

src/backends/odbc/statement.cpp
```diff
@@ -352,6 +352,17 @@ void odbc_statement_backend::describe_column(int colNum, data_type & type,
...
     columnName.assign(name, std::strlen(name));
 
+    SQLLEN is_unsigned = 0;
+    SQLRETURN rc_colattr = SQLColAttribute(hstmt_, static_cast<SQLUSMALLINT>(colNum),
+                                           SQL_DESC_UNSIGNED, 0, 0, 0, &is_unsigned);
+
+    if (is_odbc_error(rc_colattr))
+    {
+        std::ostringstream ss;
+        ss << "getting col attribute of column at position " << colNum;
```
vadz, line 362:
```suggestion
        ss << "getting \"unsigned\" attribute of the column at position " << colNum;
```

tests/sqlite3/test-sqlite3.cpp
```diff
@@ -470,10 +471,10 @@ TEST_CASE("SQLite type inference", "[sqlite][sequence]")
...
         CHECK(r1.get<std::string>(0) == cvc);
         CHECK(r1.get<double>(1) == Approx(cdec));
-        CHECK(r1.get<long long>(2) == cll);
-        CHECK(r1.get<unsigned long long>(3) == cull);
-        CHECK(r1.get<long long>(4) == cll);
-        CHECK(r1.get<unsigned long long>(5) == cull);
+        CHECK(r1.get<int64_t>(2) == cll);
+        CHECK(r1.get<uint64_t>(3) == cull);
+        CHECK(r1.get<int64_t>(4) == cll);
+        CHECK(r1.get<uint64_t>(5) == cull);
```
vadz, line 477: This is an example of a change I'd prefer to undo, it doesn't do anything useful but just adds lines of code to review and results in unnecessary code churn in Git history.

vadz (changes requested):
First of all, thanks for all this work!

I couldn't even read all of it yet, but globally it looks very good. However we really should try to avoid breaking compatibility by defining `dt_integer` as either `dt_int32` or `dt_int64` and allowing it to continue to work (which, BTW, should make a bunch of changes replacing it with `dt_int32` in this PR unnecessary which would make it smaller and easier to review too).

One other thing I'm a bit concerned about is the unsigned types support with MS SQL. Can we use unsigned types with it at all? And, on a similar topic, why neither SQLite nor PostgreSQL have "full uint64 support"? And, if they don't, who does, and is it even worth propose something that is not supported by the two arguably most popular databases?

include/private/firebird/common.h
zann1x: `<cstdint>` is a C++11 header, which is why I used `<stdint.h>`. But as you're working on dropping C++98 support in #996 , I'll change that of course.

include/soci/soci-backend.h
zann1x: Good point. Internally though, I would prefer to exclusively use the fixed-size names (dt_int32, dt_int64 etc.). Do you have a different opinion on that?

vadz: I'd prefer not to change the tests, to avoid the code churn and to make sure the old constants still work, but using fixed width types in the implementation Indeed seems like a good idea.

tests/sqlite3/test-sqlite3.cpp
zann1x: SQLite doesn't have native support for uint64. That's why you e.g. only have a `sqlite_bind_int64()` but no `sqlite_bind_uint64()` (https://www.sqlite.org/capi3ref.html).
To store a uint64 you'd have to use some kind of workaround and insert the value in a string column or similar.

zann1x:
> One other thing I'm a bit concerned about is the unsigned types support with MS SQL. Can we use unsigned types with it at all?

There's no real data type for unsigned values in MS SQL. Instead, one could use numeric with a large enough precision.

> And, on a similar topic, why neither SQLite nor PostgreSQL have "full uint64 support"? And, if they don't, who does, and is it even worth propose something that is not supported by the two arguably most popular databases?

E.g. MySQL and Sybase support uint64 in the form of `unsigned bigint` columns.

TBH I struggled to decide between

1. trying to squeeze every database into all the data types that SOCI exposes and
2. making it dependent on the database backend. I.e. only supporting those of the exposed SOCI data types that the individual database "officially" supports (as in you can use `unsigned ...` columns there).

I'm leaning towards option 2. But now that we're talking about it, it really doesn't sound like the smartest idea.

vadz:
We had never really discussed what we actually wanted to do, but I think there can be different goals here, e.g. right now, without thinking much about it, I can see at least the following ones:

1. Make sure all C++ types can be mapped to the database types losslessly, i.e. if you have something in your program, you can write it to the database and then read it back as the same type (and, maybe, can _not_ read it back as a different type?).
2. Make sure that all database types can be read and, ideally, written in some way, because, ideally, a program using SOCI should be able to work with the existing databases and not just the ones it creates.

I think the goal of this PR is to do (1) -- please correct me if I'm wrong -- but if neither MS SQL nor PostgreSQL nor SQLite can support (all) unsigned types, then I'm not even sure if it's the right goal. Maybe we should just document instead that unsigned integers have to be stored as signed ones for portability?

include/soci/soci-backend.h
zann1x: > For dt_integer it might be more complicated, but it needs to be mapped to dt_int32 or dt_int64 depending on the actual sizeof(int).

If we had a `dt_long` before, I would understand your concern. But a `dt_integer` a.k.a. `int` is never 64-bit and therefore should always be mapped to `dt_int32`, shouldn't it?

vadz: I think there are ILP64 platforms, where int is 64 bit, but maybe it's indeed an overkill to worry about those as I've never heard of SOCI being used on them, so I agree, let's just make it a synonym for `dt_int32` to keep things simple, sorry for overthinking this initially.

tests/sqlite3/test-sqlite3.cpp
zann1x: A problem I'm having with this is that the `soci::row` API seems to enforce the type that is allowed for `T` in `row.get<T>(index)` as defined in the respective backend. Undoing changes like these results in failing tests because of `std::bad_cast` exceptions.

zann1x: Nevermind, I fixed it and reverted the changes here.

tests/sqlite3/test-sqlite3.cpp
zann1x: I have to correct myself there: you can store a uint64, but not the full range of this type (only up to INT64_MAX). Returning false here just disables tests where a value bigger than INT64_MAX is worked with.

zann1x:
> I think the goal of this PR is to do (1)

You're correct.

> but if neither MS SQL nor PostgreSQL nor SQLite can support (all) unsigned types

It's not that those database can't support something like a `uint64_t`, it's just that they can't store the full range of `uint64_t` values. Sorry for the confusion here.

zann1x:
I added back the old dt_* types to make the changes backwards compatible. This is _mostly_ working. A breaking change I think we can't avoid is the updated mapping of some database column types.
You can see an example of this in test-mysql.cpp where `unsigned int` columns are now getting mapped to a `dt_uint32` instead of a `dt_long_long`.

zann1x:
@vadz, could you also please have a look at the failing installation step in the pipelines?

vadz:
> @vadz, could you also please have a look at the failing installation step in the pipelines?

I have no idea what's going on with it, it looks like an actual problem with the repository. Let's retry rerunning the builds again later today or tomorrow, maybe it will just be fixed on its own. If not, we're going to need to look into using another Ubuntu version, probably.

zann1x:
Rebased the branch to be mergeable again.

zann1x:
As I said [earlier](https://github.com/SOCI/soci/pull/954#issuecomment-1287391267), I don't see a way to merge this without breaking *some* backwards compatibility. So how about splitting this PR in two?
PR (1) adds complete integer support for x_* types, keeps backwards compatibility and can be merged for an upcoming 4.x release.
PR (2) adds complete integer support for dt_* types, but breaks backwards compatibility and gets postponed to a 5.x release.

vadz:
I'm not sure how much of a problem this compatibility break is going to be in practice, but I need to at least find time to test my own code with the PR before merging it. But, unless I find something really bad while doing it, I'd rather merge everything at once and release 4.1 with them -- I don't think the incompatibility is bad enough to warrant a major version change, but please let me know if I'm missing something.

zann1x:
> But, unless I find something really bad while doing it, I'd rather merge everything at once and release 4.1 with them -- I don't think the incompatibility is bad enough to warrant a major version change, but please let me know if I'm missing something.

Sounds good to me like that 👍🏻

src/backends/odbc/standard-use-type.cpp
```diff
@@ -23,50 +24,70 @@ void* odbc_standard_use_type_backend::prepare_for_bind(
...
+    case x_int64:
         if (use_string_for_bigint())
         {
           sqlType = SQL_NUMERIC;
           cType = SQL_C_CHAR;
           size = max_bigint_length;
           buf_ = new char[size];
           snprintf(buf_, size, "%" LL_FMT_FLAGS "d",
-                   exchange_type_cast<x_long_long>(data_));
+                   static_cast<long long>(exchange_type_cast<x_int64>(data_)));
```
vadz, line 65: I think the casts here (and below) are unnecessary as well.

src/backends/odbc/statement.cpp
```diff
@@ -352,6 +352,17 @@ void odbc_statement_backend::describe_column(int colNum, data_type & type,
...
     columnName.assign(name, std::strlen(name));
 
+    SQLLEN is_unsigned = 0;
+    SQLRETURN rc_colattr = SQLColAttribute(hstmt_, static_cast<SQLUSMALLINT>(colNum),
+                                           SQL_DESC_UNSIGNED, 0, 0, 0, &is_unsigned);
+
+    if (is_odbc_error(rc_colattr))
+    {
+        std::ostringstream ss;
+        ss << "getting \"unsigned\" col attribute of the column at position " << colNum;
```
vadz, line 362: As this message could be shown to the user, it would be better to avoid abbreviations in it:
```suggestion
        ss << "getting \"unsigned\" column attribute of the column at position " << colNum;
```

src/backends/postgresql/statement.cpp
```diff
@@ -768,15 +768,18 @@ void postgresql_statement_backend::describe_column(int colNum, data_type & type,
         type = dt_double;
         break;
 
-    case 16:   // bool
     case 21:   // int2
+        type = dt_int16;
+        break;
+
+    case 16:   // bool
```
vadz, line 775: Is `bool` really 4 bytes in PostgreSQL? [The docs](https://www.postgresql.org/docs/current/datatype-boolean.html) seem to be saying that it's 1, but I could be mixing up something here.

tests/odbc/test-odbc-postgresql.cpp
```diff
@@ -212,6 +212,11 @@ class test_context : public test_context_base
         return !m_verDriver.is_initialized() || m_verDriver < odbc_version(9, 3, 400);
     }
 
+    bool has_full_uint64_support() const override
```
vadz, line 215: I'm still surprised that PostgreSQL doesn't have it, it's the most standard compliant database, so are we really sure that it's wrong here and not us?

tests/common-tests.h
```diff
@@ -1511,6 +1521,78 @@ TEST_CASE_METHOD(common_tests, "Use type conversion", "[core][use]")
...
+        int8_t s = 123;
+        sql << "insert into soci_test(id) values(:id)", use(s);
+
+        int8_t s2 = 0;
+        sql << "select id from soci_test", into(s2);
+
+        CHECK(s2 == 123);
+    }
+
+    SECTION("int8_t min")
```
vadz, line 1535: Is it really worth having 3 sections for each type? I think we could combine all the tests for the same type in a single section, which would make the test slightly shorter and faster to run.

Also, I'd really like to know what happens when trying to insert an out of range value into a column too small (or "too signed") to contain it, it would be great to add some tests checking that this behaves as expected too.

vadz (commented):
First of all, thanks a lot for the incredible work you've done on this!

Globally it looks very good and I don't see any regressions when using this branch with my own code, which is great.

The only non-great part is that the current code doesn't support Win32 and duplicates the same or very similar code between 3 branches. I think we could solve both of these problems relatively easily, but please let me know if I'm missing something.

Finally, I'm still not sure which databases actually have full `uint64_t` support. If PostgreSQL doesn't have it, who does?

Thanks again for working on this and I'll try to merge it a.s.a.p. if/when the problem with Win32 is fixed.

jsonn:
> It's not that those database can't support something like a `uint64_t`, it's just that they can't store the full range of `uint64_t` values. Sorry for the confusion here.

They all support NUMERIC, don't they?

vadz:
It would be really great to finish and merge this and make a new SOCI release. Please let me know if you can do it, TIA!

zann1x:
Sorry for the delay on this. I'll look into it this week.

include/soci/exchange-traits.h
```diff
@@ -43,66 +44,94 @@ struct exchange_traits
...
 template <>
-struct exchange_traits<long int>
+struct exchange_traits<uint64_t>
 {
     typedef basic_type_tag type_family;
-    enum { x_type = long_traits_helper<sizeof(long int)>::x_type };
+    enum { x_type = x_uint64 };
+};
+
+#if defined(_WIN64)
```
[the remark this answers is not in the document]
zann1x: I don't quite like the idea of defining a "new" type for this. Instead of it, I added more descriptive defines. I _think_ that these can also handle the 32-bit case now.
Let me know what you think about this.

src/backends/mysql/standard-use-type.cpp
```diff
@@ -64,38 +65,76 @@ void mysql_standard_use_type_backend::pre_use(indicator const *ind)
...
-        case x_short:
+        case x_int8:
             {
                 std::size_t const bufSize
-                    = std::numeric_limits<short>::digits10 + 3;
+                    = std::numeric_limits<int8_t>::digits10 + 3;
                 buf_ = new char[bufSize];
                 snprintf(buf_, bufSize, "%d",
-                    static_cast<int>(exchange_type_cast<x_short>(data_)));
+                    static_cast<int>(exchange_type_cast<x_int8>(data_)));
```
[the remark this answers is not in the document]
zann1x: Yes, you're right.

src/backends/odbc/standard-use-type.cpp
zann1x: This is the one of the only places where the cast is necessary. `int64_t` isn't defined as `long long` on every platform. The format specifier `long long` is used as this should be the common ground for 64-bit integers across all platforms. If we don't cast here, the compiler complains with a warning.

src/backends/postgresql/statement.cpp
zann1x: You're right, I didn't check that.

tests/odbc/test-odbc-postgresql.cpp
zann1x: TBH I'm confused. IIRC I added these opt-out methods because I had multiple failing tests regarding this during implementation of this PR. As I can't reproduce the errors anymore, I removed them again. That's good news I guess.

The only backends that still have issues with some types are
* Firebird which can't store values bigger INT64_MAX correctly. It instead stores an overflown integer value. This mixes up the ordering of values in the uint64_t vector test.
* SQLite which processes integers as 8-byte signed values. It basically has the same issue as Firebird.
* MSSQL ODBC which seems to be the only one of our tested ODBC backends that needs to use a bigger SQL type in order to store unsigned values (i.e. SQL_BIGINT instead of SQL_INTEGER for uint32_t et cetera).
All of those are either fixed or explicitly tested instead of using an opt-out, though.

tests/common-tests.h
zann1x: > Is it really worth having 3 sections for each type? I think we could combine all the tests for the same type in a single section, which would make the test slightly shorter and faster to run.

I don't really care if it's multiple small sections or a single bigger one, so we can combine that, of course.

> Also, I'd really like to know what happens when trying to insert an out of range value into a column too small (or "too signed") to contain it, it would be great to add some tests checking that this behaves as expected too.

See above, it's not really a problem anymore. But I added the checks where necessary.

include/soci/fixed-size-ints.h
```diff
@@ -0,0 +1,156 @@
...
+{
+
+// Completion of dt_[u]int* bindings for all architectures.
+// This allows us to extract values on types where the type definition is
+// specific to the underlying architecture. E.g. Unix defines a int64_t as
+// long. This would make it impossible to extract a dt_int64 value as both
+// long and long long. With the following type_conversion specializations,
+// this becomes possible.
+
+#if defined(_WIN64)
```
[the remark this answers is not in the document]
zann1x: See above, the same unified defines are now used here.

src/backends/mysql/standard-into-type.cpp
```diff
@@ -89,17 +90,37 @@ void mysql_standard_into_type_backend::post_fetch(
                 dest.assign(buf, lengths[pos]);
             }
             break;
-        case x_short:
-            parse_num(buf, exchange_type_cast<x_short>(data_));
+        case x_int8:
+            {
+                int32_t tmp = 0;
+                parse_num(buf, tmp);
+                exchange_type_cast<x_int8>(data_) = static_cast<int8_t>(tmp);
```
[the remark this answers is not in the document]
zann1x: Well yes, the value could indeed overflow.
I don't know the reasoning behind the custom `parse_num()` method, but as it uses `std::stringstream` for conversion, we can't use it directly for `int8_t` values as it interprets these as characters. I could add explicit range checks at those places to emulate the same behavior. Or maybe we could also replace `parse_num()` (here and at all the other places) with the more general `cstring_to_integer()` method, which already does these range checks?

zann1x:
Rebased on master and added the requested changes. Sorry again for the delay.

vadz:
Thanks again for your work! I'll try to look at it a.s.a.p. but maybe not before the next week-end (i.e. I definitely won't be able to do it today, sorry). For now I see that MySQL tests fail in the AppVeyor CI builds, would you have any idea about why does this happen?

zann1x:
No, I don't quite know why this is happening, especially as it's not happening on all builds. Am I wrong to assume that the MySQL and ODBC driver version are the same on all AppVeyor builds?

vadz:
The drivers are the same, please see `.appveyor.yml`. I haven't had time to look at what's going on there, but this needs to be fixed somehow...

zann1x:
Ahh, the VS2019 and VS2022 builds pass because they don't have MySQL ODBC drivers installed. The MySQL ODBC tests don't even run there 😅 

On a quick search I found this bug: https://bugs.mysql.com/bug.php?id=6111 - it may be related. But I'll need to have another look at it in the next few days.

zann1x:
It seems like the failing MySQL ODBC test is a bug in the ODBC driver itself. This bug might have already been fixed in the latest versions of the driver, though (see the above links). But as the AppVeyor builds use an older driver version, I have added it as a special case in the unit tests.

vadz:
Thanks! I'm fine with disabling these tests with the known buggy versions of MySQL ODBC driver, so I'll try to merge this soon, after having one last look at it.

Thanks again for all your work on this!

zann1x:
@vadz did you find time to review the latest changes yet?

vadz:
Sorry, not yet :-( I'll really, really try to do it soon -- and hopefully make 4.1.0 release after merging it.

zoman2000:
Just curious, what is wrong with current data types? I woulds say we have one to be removed - dt_integer because it doesn't have its' unsigned counterpart.

What we don't have - is real support for DECIMAL type. It gets converted to DOUBLE which is more wrong than not having  int_8, int_16, .... 

DECIMAL is not DOUBLE

however

123 is a dt_unsugned_long_liong or dt_long_long or even dt_integer

As a teaser:
create a table with a colum type decimal(65,30)
insert this value: 111111111111111111111111111111.222222222222222222222222222222
then select into double and print.
You will get this value: 1.11111e+29

This is a disaster.

This is what comes out from mysql:
select `dec` from cookie where id = 1;
+---------------------------------------------------------------+
| dec                                                           |
+---------------------------------------------------------------+
| 111111111111111111111111111111.222222222222222222222222222222 |
+---------------------------------------------------------------+
1 row in set (0.000 sec)

vadz:
@zoman2000 I think the PR description explains this quite well, I don't have much to add. Of course, this doesn't mean that there are no other improvements to make, better `DECIMAL` support would clearly be nice to have too.

alex-kuzza-aaa:
> @zoman2000 I think the PR description explains this quite well, I don't have much to add. Of course, this doesn't mean that there are no other improvements to make, better `DECIMAL` support would clearly be nice to have too.

My question was about why we are doing a change that is not critically necessary,. Could it be possible to address reported unsigned integer bug(s) instead? Then leave the types intact and map to dt_integer only "compatible" types. 

Otherwise it looks it is following the path of mariadb library with all the types  like signed/unsigned[8,16,32,64], or 8 types in total that could lead to a domain or type error during data exchange.

vadz:
> My question was about why we are doing a change that is not critically necessary

I am not sure what does "critically" mean exactly, but for me not being able to round trip many integer types is pretty bad, so I think it's rather important to fix this.

> Could it be possible to address reported unsigned integer bug(s) instead?

I don't see/know how.

> Then leave the types intact and map to dt_integer only "compatible" types.
> 
> Otherwise it looks it is following the path of mariadb library with all the types like signed/unsigned[8,16,32,64], or 8 types in total that could lead to a domain or type error during data exchange.

Sorry, not sure what does this mean. Do you think this PR introduces incompatible changes affecting the behaviour of the existing code? AFAIK it's not supposed to, but one of the reasons I still haven't merged it is that I'd like to test it more to be more certain about it (and I'm really sorry that I still haven't found time to do it...). So if you have any precise examples of such changes, could you please share them?

alex-kuzza-aaa:
> 
> > My question was about why we are doing a change that is not critically necessary
> 
> I am not sure what does "critically" mean exactly, but for me not being able to round trip many integer types is pretty bad, so I think it's rather important to fix this.
> 
> > Could it be possible to address reported unsigned integer bug(s) instead?
> 
> I don't see/know how.

sounds like you some meeting is required to elaborate on this

> 
> > Then leave the types intact and map to dt_integer only "compatible" types.
> > Otherwise it looks it is following the path of mariadb library with all the types like signed/unsigned[8,16,32,64], or 8 types in total that could lead to a domain or type error during data exchange.
> 
> Sorry, not sure what does this mean. Do you think this PR introduces incompatible changes affecting the behaviour of the existing code? AFAIK it's not supposed to, but one of the reasons I still haven't merged it is that I'd like to test it more to be more certain about it (and I'm really sorry that I still haven't found time to do it...). So if you have any precise examples of such changes, could you please share them?

1) This change will break a code that uses the  column type for dynamic binding in the some ORM engine.
Something like this, for example:
      switch (t) {
      ......
        case soci::data_type::dt_double: {
          val = r.get<double>(i);
        } break;
        case soci::data_type::dt_integer: { // YES, we have no support for unsigned :(
          val = r.get<int>(i);
        } break;
        case soci::data_type::dt_long_long:
        case soci::data_type::dt_unsigned_long_long: {
          val = r.get<long long>(i);
        } break;
       .......

     This code will compile, because previous dt_integer and dt_long_long retained in place, however may not work and will create a "hidden problem". Maybe you should stop using dt_integer & dt_long_long completely. Code won't compile and will require a review instead  of just not working.

2)  The other point to mention is the type mapping on the database API level. For example if you expect to have "long" as your C++ type it is not always true for mariadb, you may get an internal type like signed64 and of course you will get a "bad cast". Think also about types like tinyint, int, bigint and possibility to alter a table to change tinyint to let's say int. Then your code will fail because you were using something like r.get<char>(col_index).....

3) So the third problem - you have to recompile your code if you alter the database column types (up or down).

I don't want to say that if you change "signed" to "unsigned" get<char>(col_index) may also fail.

For current design you just have your C++ type as long long (or unsigned long long)  and do not care about all the above.

zann1x:
> This code will compile, because previous dt_integer and dt_long_long retained in place, however may not work and will create a "hidden problem". Maybe you should stop using dt_integer & dt_long_long completely. Code won't compile and will require a review instead of just not working.

As @vadz mentioned in https://github.com/SOCI/soci/pull/954#discussion_r973624874, we would break backwards compatibility with that change. Admittedly, I didn't think that defining `dt_integer` and `dt_long_long` as aliases for some of the new types would actually create a problem. But I don't quite know how to avoid the API breakage in a different way, so it might indeed be worth reconsidering their removal.

> Think also about types like tinyint, int, bigint and possibility to alter a table to change tinyint to let's say int. Then your code will fail because you were using something like r.get<char>(col_index).....
> [...]
> So the third problem - you have to recompile your code if you alter the database column types (up or down).

Shouldn't a change in a table schema almost always result in also updating the corresponding application code for it? Otherwise you might create a "hidden problem" yourself in the long run.

> For current design you just have your C++ type as long long (or unsigned long long) and do not care about all the above.

IMO it's preferable to have exact type mappings to avoid any uncertainties even if that means a little more work in application code.

vadz:
> > > Could it be possible to address reported unsigned integer bug(s) instead?
> > 
> > I don't see/know how.
> 
> sounds like some meeting is required to elaborate on this

Sorry, we can't really decide to organize a meeting. If you have an alternative proposal, please formulate it or, better, try to provide a PR implementing it and let's discuss it.

But please note that lack of support of unsigned types is not the only problem. I.e. I don't think it's a good idea to (always) represent all integer types as `long long` or `unsigned long long`.

>     1. This change will break a code that uses the  column type for dynamic binding in the some ORM engine.

Yes, this is indeed true, thanks for bringing this up. I admit I haven't thought about this case at all because I don't use dynamic binding myself, but it's a valid use case.

>        Something like this, for example:
> ```cpp
>        switch (t) {
>        ......
>        case soci::data_type::dt_double: {
>        val = r.get<double>(i);
>        } break;
>        case soci::data_type::dt_integer: { // YES, we have no support for unsigned :(
>        val = r.get<int>(i);
>        } break;
>        case soci::data_type::dt_long_long:
>        case soci::data_type::dt_unsigned_long_long: {
>        val = r.get<long long>(i);
>        } break;
> ```
>        This code will compile, because previous dt_integer and dt_long_long retained in place, however may not work and will create a "hidden problem". Maybe you should stop using dt_integer & dt_long_long completely. Code won't compile and will require a review instead  of just not working.

I also agree that silent breakage is bad, but OTOH there must be quite a lot of code using `dt_xxx` constants, so breaking all of it would be bad too.

I think ideal solution would be to preserve preserve backwards-compatible behaviour in the existing API, but deprecate the functions returning the `dt_xxx` (and the field of struct `column_info` containing it), and add new ones (and new field in `column_info`) containing the new "precise" integer types. Like this the existing code should keep working, while newly written one could use new API to get the more appropriate type. The main, and obvious, problem I see with this approach is that it would require quite a bit more work and I'm not sure if @zann1x has the possibility to work on it. Of course, there might be other problems with this too, but right now I don't see any.

>     2. The other point to mention is the type mapping on the database API level. For example if you expect to have "long" as your C++ type it is not always true for mariadb, you may get an internal type like signed64 and of course you will get a "bad cast". Think also about types like tinyint, int, bigint and possibility to alter a table to change tinyint to let's say int. Then your code will fail because you were using something like r.get<char>(col_index).....

Sorry, I don't understand this part. If it wouldn't be addressed by the change proposed above, could you please explain this in more details?

>     3. So the third problem - you have to recompile your code if you alter the database column types (up or down).

I think it's not unreasonable to require updating the code if the database schema changes.

> For current design you just have your C++ type as long long (or unsigned long long) and do not care about all the above.

With this PR you could just handle all integer types in the same way by casting them to `long long` if you wish, it just also gives you the possibility to not do it, which is not really possible right now.

To summarize, I think the problem you point out with the change of behaviour in the code using `dt_integer` is real and indeed bad, because it's a silent change.But I believe we could address it by letting `dt_integer` work as before (and maybe, but not even necessarily, if people really find the current behaviour useful, deprecating it) and adding new functions/fields using the new constants alongside it.

What do you think?

zann1x:
> I think ideal solution would be to preserve preserve backwards-compatible behaviour in the existing API, but deprecate the functions returning the dt_xxx (and the field of struct column_info containing it), and add new ones (and new field in column_info) containing the new "precise" integer types. Like this the existing code should keep working, while newly written one could use new API to get the more appropriate type.

How about adding a compile definition called e.g. `SOCI_USE_FIX_WIDTH_TYPES`? When defined, SOCI uses the new types `dt_[u]int<n>`, which new users should be encouraged to use. Ultimately this could become the default behavior in some future version of the library. This way, nothing breaks for existing applications and anyone can opt in for the more precise types at any time.

> The main, and obvious, problem I see with this approach is that it would require quite a bit more work and I'm not sure if @zann1x has the possibility to work on it.

I won't be able to make any promises about the time it will take me to do it, but I can work on it for sure.

vadz:
> How about adding a compile definition called e.g. `SOCI_USE_FIX_WIDTH_TYPES`?

Sorry, I realize the appeal of doing this, but I'd be strongly against it because every time I had done something like this in the past it turned out to be a bad decision. The most immediately practical reason I can give against doing it for SOCI is that it's included in different distributions as a package and distributions will have to build it either with or without this symbol defined. If they choose to not define it, it means that any code using system libraries will still be affected by backwards-incompatible changes, which seems bad as it would encourage people to not use the system libraries. If they choose, as it seems more likely, to define it, it means that you'd have to stop using system libraries in order to use the new preferred API which seems even worse.

Thinking more about this, I wonder if it could perhaps be better/less confusing to add a completely new `db_type` enum (the name is subject to change, it's just the fact that it would be different from the existing `data_type` which matters) and `db_intNN` and `db_uintNN` constants rather than trying to keep the existing `dt_xxx`? This would require slightly more changes when upgrading the existing code to use newer API, but it shouldn't be too bad and, of course, you wouldn't need to do it as the old API would still keep working (at least for now). What do you think?

zann1x:
> Sorry, I realize the appeal of doing this, but I'd be strongly against it because every time I had done something like this in the past it turned out to be a bad decision. The most immediately practical reason I can give against doing it for SOCI is that it's included in different distributions as a package and distributions will have to build it either with or without this symbol defined. If they choose to not define it, it means that any code using system libraries will still be affected by backwards-incompatible changes, which seems bad as it would encourage people to not use the system libraries. If they choose, as it seems more likely, to define it, it means that you'd have to stop using system libraries in order to use the new preferred API which seems even worse.

Alright, I understand that. Thanks for the explanation.

> Thinking more about this, I wonder if it could perhaps be better/less confusing to add a completely new db_type enum (the name is subject to change, it's just the fact that it would be different from the existing data_type which matters) and db_intNN and db_uintNN constants rather than trying to keep the existing dt_xxx? This would require slightly more changes when upgrading the existing code to use newer API, but it shouldn't be too bad and, of course, you wouldn't need to do it as the old API would still keep working (at least for now). What do you think?

I don't know how to feel about this. On the one hand it sounds intriguing because we wouldn't interfere with the old API in any way. On the other hand I'm unsure how much of the library internals we would have to change because I feel like the dt_* and x_* types are quite interlocked with the current implementation.
Maybe we could just define new methods for the cases the user actually gets to see a dt_* type? `column_properties::get_data_type()` comes to my mind here. A new method like `column_properties::get_mapped_data_type()` could return the fixed width types and the existing `column_properties::get_data_type()` could map the fixed width types to the respective old dt_* types. Though I'm unsure if that's feasible for all cases where dt_* types are visible to a user.

Also, how do we want to proceed with the x_* types? If they are visible to users (which I suppose is possible _somewhere_?), they probably need a rework as well.

alex-kuzza-aaa:
@vladz   2&3 use cases.

This is mostly for up/down alter table scenarios, when a native data type may change from int8 to int16...int64. If you use a specific  int8 accessor/binding that will break your code. So for new development will be very important to maintain your code along with the database update, which may happen when you have your code running. This is too tightly coupled solution.

Regarding MariaDB problems I mentioned. Imagine you have a type  UNSIGNED INT(1), that is basically should be a "boolean" , but it is still unsigned int32 and you can insert any integer value (nice feature).

zann1x:
> > > Thinking more about this, I wonder if it could perhaps be better/less confusing to add a completely new db_type enum (the name is subject to change, it's just the fact that it would be different from the existing data_type which matters) and db_intNN and db_uintNN constants rather than trying to keep the existing dt_xxx? This would require slightly more changes when upgrading the existing code to use newer API, but it shouldn't be too bad and, of course, you wouldn't need to do it as the old API would still keep working (at least for now). What do you think?
> > 
> > 
> > I don't know how to feel about this. On the one hand it sounds intriguing because we wouldn't interfere with the old API in any way. On the other hand I'm unsure how much of the library internals we would have to change because I feel like the dt_* and x_* types are quite interlocked with the current implementation.
> 
> I'm not totally sure about this neither, but my hope was that the new type would be almost (or maybe even exactly?) the same as `exchange_type`, which would hopefully also make thing less confusing as the interplay between it and `data_type` right now is not always obvious.
> 
> > Maybe we could just define new methods for the cases the user actually gets to see a dt_* type?
> 
> Yes, just to be clear, this was my original suggestion in a comment above, i.e. this is exactly what I meant by
> 
> > I think ideal solution would be to preserve preserve backwards-compatible behaviour in the existing API, but deprecate the functions returning the `dt_xxx` (and the field of struct `column_info` containing it), and add new ones (and new field in `column_info`) containing the new "precise" integer types.
> 
> And this could still be the best idea. But I think we would only become sure about it when trying one or the other approaches, right now, when just looking at it from very high level, it isn't obvious (to me, at least) whether it is or not.

Summing up all of this: I'll try to find some time to work on it soon and try out the different approaches.

> > Also, how do we want to proceed with the x_* types? If they are visible to users (which I suppose is possible _somewhere_?), they probably need a rework as well.
> 
> I'm pretty sure that these ones are supposed to be only used internally. And considering that they are in `soci::details` namespace, I feel like we would be justified in changing them incompatibly. And a simple [code search](https://github.com/search?q=soci%3A%3Adetails%3A%3Ax_&type=code) for them doesn't find anything which is encouraging.

The only repo I could find was [lib-ledger-core](https://github.com/LedgerHQ/lib-ledger-core/blob/ad1e271b5f61e92f663ff08c337acc8b107fc85f/core/src/database/proxy_backend/standard-use-type.cpp#L68). But I agree with you that the `soci::details` namespace should normally such usage.

zann1x:
A quick update from my side so you know that I haven't forgotten about this PR:

I've rebase the branch on the current master, jotted down something that introduces a new type `db_type` and reverted `data_type` back to its original state. Every user-facing method/variable that previously dealt with `data_type` got its respective overload or addition for `db_type`. The silent breakage should be gone now. I haven't updated the docs yet, but if you're okay with the changes, I'll update them accordingly.

Note that users can still encounter `std::bad_cast` exceptions because of the stricter type mapping. This can result in the weird circumstance that the old API e.g. returns `dt_long_long`, which, despite the name, you now have to retrieve as `uint32_t` instead of `long long`. So far, this worked because of the `type_conversion`s in `unsigned-types.h`. Examples for that would be [1](https://github.com/SOCI/soci/pull/954/files#diff-3c7d5f688a808dc7d9b8b8e3cf70533fc76947df9a87c24a03ec963230b42782R703-R706), [2](https://github.com/SOCI/soci/pull/954/files#diff-3c7d5f688a808dc7d9b8b8e3cf70533fc76947df9a87c24a03ec963230b42782R715-R717), [3](https://github.com/SOCI/soci/pull/954/files#diff-3c7d5f688a808dc7d9b8b8e3cf70533fc76947df9a87c24a03ec963230b42782R726-R728).
We could possible relax the type check in `row::get(size_t)` in order to avoid some (or even all?) of the `std::bad_cast` exceptions if you're unhappy with these. The implicit type casts from https://github.com/SOCI/soci/pull/918 may be re-integrated here.

Next to that I noticed that I haven't added the bounds check in `mysql::standard_into_type` that we once talked about. This is now fixed as well.

> > Also, how do we want to proceed with the x_* types? If they are visible to users (which I suppose is possible somewhere?), they probably need a rework as well.
> 
> I'm pretty sure that these ones are supposed to be only used internally. And considering that they are in soci::details namespace, I feel like we would be justified in changing them incompatibly. And a simple [code search](https://github.com/search?q=soci%3A%3Adetails%3A%3Ax_&type=code) for them doesn't find anything which is encouraging.

Another comment on this one: if they are supposed to be only used internally, can't we remove the old names and spare us type aliases like `x_integer = x_int32`?

vadz:
Thanks a lot for the update and I'll try to get back to this a.s.a.p. but I'm on vacation right now, so it won't happen immediately -- sorry for yet another delay.

In the meanwhile, any testing and, in particular, reports of any compatibility problems would be welcome!

include/soci/sqlite3/soci-sqlite3.h
```diff
@@ -318,9 +325,14 @@ struct sqlite3_session_backend : details::session_backend
...
-            case dt_unsigned_long_long:
+            case dt_int8:
+            case dt_uint8:
+            case dt_int16:
+            case dt_uint16:
+            case dt_int32:
+            case dt_uint32:
+            case dt_int64:
+            case dt_uint64:
                 return "integer";
```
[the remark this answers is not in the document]
zann1x: SQLite only has the storage class `INTEGER` for integers, so we don't really have a choice on choosing here. Also, this method is used in the API for creating DDL queries, which has nothing to do with retrieving values.

include/soci/column-info.h
```diff
@@ -33,22 +37,36 @@ struct type_conversion<column_info>
     static std::size_t get_numeric_value(const values & v,
         const std::string & field_name)
     {
-        data_type dt = v.get_properties(field_name).get_data_type();
```
Sildra, line 36: We are performing name lookup twice here

vadz (approved):
Thanks a lot once again for all your work on this!

AFAICS there indeed shouldn't be any compatibility problems remaining, and the few minor questions below can be dealt with later, so I think we can merge this -- but you mentioned that you wanted to update the docs further and this would, of course, be very welcome.

Please let me know if you plan to do this in the near future or if I should merge it and maybe try to update them myself.

Thanks again!

src/backends/mysql/statement.cpp
```diff
@@ -402,24 +402,47 @@ int mysql_statement_backend::prepare_for_describe()
...
-    data_type & type, std::string & columnName)
+    data_type & type, db_type & dbtype, std::string & columnName)
 {
     int pos = colNum - 1;
     MYSQL_FIELD *field = mysql_fetch_field_direct(result_, pos);
     switch (field->type)
     {
     case FIELD_TYPE_CHAR:       //MYSQL_TYPE_TINY:
+        type = dt_integer;
+        dbtype = field->flags & UNSIGNED_FLAG ? db_uint8 : db_int8;
```
Sildra, line 413: Should I do get<int> or get<int8_t> ?

tests/postgresql/test-postgresql.cpp
```diff
@@ -1199,7 +1214,7 @@ struct test_enum_with_explicit_custom_type_int_rowset : table_creator_base
 
         try
         {
-            sql << "CREATE TABLE soci_test( Type smallint)";
+            sql << "CREATE TABLE soci_test( Type integer)";
```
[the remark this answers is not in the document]
Sildra: smallint is mapped to integer in the current code and int16 in the new one.

zann1x:
> AFAICS there indeed shouldn't be any compatibility problems remaining, and the few minor questions below can be dealt with later, so I think we can merge this

Nice, that's great to hear!

> but you mentioned that you wanted to update the docs further and this would, of course, be very welcome.
> 
> Please let me know if you plan to do this in the near future or if I should merge it and maybe try to update them myself.

I'll update the docs and address your comments sometime later this week.

src/backends/mysql/statement.cpp
vadz: This is discussed in #1088 so let's keep it there to reduce confusion.

tests/postgresql/test-postgresql.cpp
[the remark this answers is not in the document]
vadz: @zann1x Does this mean that using `into(int_var)` from a column of `smallint` type doesn't work any longer? If so, this would be problematic as it still would be a silent break.

There is a discussion of allowing (at least broadening) implicit conversions in #1088 but can we do something like this for `into()` itself?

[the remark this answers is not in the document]
zann1x: That is basically what I meant [here](https://github.com/SOCI/soci/pull/954#issuecomment-1676396849):

> Note that users can still encounter `std::bad_cast` exceptions because of the stricter type mapping. This can result in the weird circumstance that the old API e.g. returns `dt_long_long`, which, despite the name, you now have to retrieve as `uint32_t` instead of `long long`. So far, this worked because of the `type_conversion`s in `unsigned-types.h`. Examples for that can be found in test-mysql.cpp ([1](https://github.com/SOCI/soci/pull/954/files#diff-3c7d5f688a808dc7d9b8b8e3cf70533fc76947df9a87c24a03ec963230b42782R703-R706), [2](https://github.com/SOCI/soci/pull/954/files#diff-3c7d5f688a808dc7d9b8b8e3cf70533fc76947df9a87c24a03ec963230b42782R715-R717), [3](https://github.com/SOCI/soci/pull/954/files#diff-3c7d5f688a808dc7d9b8b8e3cf70533fc76947df9a87c24a03ec963230b42782R726-R728)).
> We could possible relax the type check in `row::get(size_t)` in order to avoid some (or even all?) of the `std::bad_cast` exceptions if you're unhappy with these. The implicit type casts from #918 may be re-integrated here.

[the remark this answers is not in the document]
Sildra: #918 should be reworked as it declares template metaprog functions that are already standard in c++11.

include/soci/soci-platform.h
```diff
@@ -175,4 +175,23 @@ private: \
     #define SOCI_DUMMY_RETURN(x) return x
 #endif
 
+#define SOCI_OS_LINUX       0x0001
+#define SOCI_OS_FREE_BSD    0x0002
+#define SOCI_OS_APPLE       0x0003
+#define SOCI_OS_WINDOWS     0x0004
```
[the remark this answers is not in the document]
zann1x: `SOCI_OS` is (indirectly) used further down the line in preprocessor checks for enabling specific `exchange_traits` and `type_conversion`s. I don't know how this would play out with `enum class` and/or `constexpr`.

include/soci/soci-backend.h
```diff
@@ -313,15 +347,85 @@ class session_backend
     {
         return "truncate table " + tableName;
     }
+    // DEPRECATED. USE create_column_type(db_type, int, int) INSTEAD.
```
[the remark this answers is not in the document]
zann1x: Done!

tests/postgresql/test-postgresql.cpp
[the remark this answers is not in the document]
zann1x: Adding to my previous answer: when using `smallint`, we would get a `std::bad_cast` exception in the test `test_enum_with_explicit_custom_type_int_rowset` (which uses `row`). The test `test_enum_with_explicit_custom_type_int_into` (which uses `into(...)`) works with both `smallint` and `integer`.

That's something I couldn't avoid for the `row` API. It's also something I wouldn't consider "silent" breakage because this always throws an exception instead of swallowing an unmapped/unhandled data type as it did before my changes in https://github.com/SOCI/soci/pull/954/commits/826fa7753a4a30c22e026847dec785aa133f0a3b.

It works for `into(...)` because `into` uses the enum `exchange_type`, which kind of has a dynamic type handling in its `standard_into_type` handlers already.
`row` on the other hand utilizes the class `type_holder` which stores a concrete type `T`. This concrete type is the type that we're mapping the database types to (via the enum `db_type`).

include/soci/odbc/soci-odbc.h
```diff
@@ -545,6 +548,25 @@ inline void odbc_standard_type_backend_base::set_sqllen_from_value(SQLLEN &targe
...
+    // MSSQL ODBC driver seemingly can't handle the conversion of unsigned C
+    // types to their respective unsigned SQL type because they are out of
+    // range for their supported signed types. This results in the error
+    // "Numeric value out of range (SQL state 22003)".
+    // The only place it works is with tinyint values as their range is
+    // [0..255], i.e. they have enough space for unsigned values anyway.
+    return statement_.session_.get_database_product()
+            != odbc_session_backend::prod_mssql;
+}
+
```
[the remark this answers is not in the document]
zann1x: All of these helper functions are declared and defined in the header, so I just continued this convention.

> I suspect we might need to change/adjust them in the future to account for more ODBC quirks (e.g. test the exact version or something else).

Because of this I would just leave them as is if that's okay with you.

tests/postgresql/test-postgresql.cpp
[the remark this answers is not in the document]
vadz: So I guess we do need something along the lines of #1088 to preserve compatibility for row API uses (BTW, just to fix the definitions: "silent breakage" is code which continues to compile, without warnings, but changes behaviour, so this would be one).

@Sildra Would you agree with reworking your PR to only allow lossless implicit type conversions, i.e. those from smaller to bigger int types (of the same sign?), in the row API? If so, I still think we should merge this one and then merge your PR. TIA!

include/soci/soci-backend.h
Sildra, line 1:
```cpp
    virtual std::string create_column_type(data_type dt,
                                           int precision, int scale);
```

As the `create_column_type(data_type...` maps to `create_column_type(db_type...`, the virtual should be removed as the backends overrides.

Sildra: This could be generalized to all interfaces using `data_type` with a redirection like `return call(to_db_type(data_type), ...);`

tests/postgresql/test-postgresql.cpp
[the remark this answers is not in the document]
Sildra: I think I can allow this by comparing the sizeof of source and target as precheks and separating double. Will work on it tommorow.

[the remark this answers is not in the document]
Sildra: Better idea, I can change the static cast to perform runtime narrowing checks.

Sildra:
As more and more databases are adding a json type, is it possible to add a `db_json` in the enum ?

tests/postgresql/test-postgresql.cpp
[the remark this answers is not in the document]
Sildra: https://github.com/Sildra/soci/commit/0eb519b71004cb71c1ff08ac9f84e3411df8ae4d

Let me know what you think of this implem.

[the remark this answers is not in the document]
vadz: I'm not convinced that we need narrowing conventions, it risks being very confusing if your code works perfectly fine at first and then suddenly starts falling with an exception just because somebody added a value allowed by the column type but out of the C++ type range to the database.

But I'd also prefer to discuss this in its own PR, this one is becoming very difficult to navigate.

vadz:
Sorry for missing my own deadline of merging this in 2023, but I think it should be ready to merge now with just some minor changes from #1116 -- please let me know if you see anything wrong with them, otherwise I'll (squash) merge that one soon.

Thanks again for all your work here!

PR #960, opened 2022-05-02
Fix parameter name doc for the SQLITE3 library dir.

ccosmin:
Passing SQLITE3_LIBRARIES results in cmake not finding sqlite3.

### DIFF
--- a/docs/installation.md
+++ b/docs/installation.md
@@ -163,7 +163,7 @@ Some other build options:
 
 * `WITH_SQLITE3` - boolean - Should CMake try to detect SQLite C/C++ library. As bonus, the configuration tries OSGeo4W distribution if OSGEO4W_ROOT environment variable is set.
 * `SQLITE3_INCLUDE_DIR` - string - Path to SQLite 3 include directory where CMake should look for `sqlite3.h` header.
-* `SQLITE3_LIBRARIES` - string - Full paths to libraries to link SOCI against to enable the backend support.
+* `SQLITE3_LIBRARY` - string - Full paths to libraries to link SOCI against to enable the backend support.
 * `SOCI_SQLITE3` - boolean - Requests to build [SQLite3](backends/sqlite3.md) backend. Automatically switched on, if `WITH_SQLITE3` is set to ON.
 * `SOCI_SQLITE3_TEST_CONNSTR` - string - Connection string is simply a file path where SQLite3 test database will be created (e.g. /home/john/soci_test.db). Check [SQLite3 backend reference](backends/sqlite3.md) for details. Example: `-DSOCI_SQLITE3_TEST_CONNSTR="my.db"` or `-DSOCI_SQLITE3_TEST_CONNSTR=":memory:"`.
 


vadz:
I have no idea about CMake, so I don't understand what's the advantage of using `SQLITE3_LIBRARIES` compared to `SQLITE3_LIBRARY`, but I see that the [official docs](https://cmake.org/cmake/help/latest/module/FindSQLite3.html) seem to use `SQLite3_LIBRARIES`, shouldn't we use this one too?

ccosmin:
I guess, can be done as well. I wasn't familiar with this convention.

vadz:
I just don't have any idea what's the right thing here and whether applying this would make things better or worse. Are any of the people who have contributed to CMake support still here? Any guidance would be welcome, TIA!

zann1x:
It's actually more correct to set the variable `SQLITE3_LIBRARY` than `SQLITE3_LIBRARIES` as a user of the library (see https://cmake.org/cmake/help/latest/manual/cmake-developer.7.html#standard-variable-names):

> Xxx_LIBRARIES
>
>    The libraries to use with the module. These may be CMake targets, full absolute paths to a library binary or the name of a library that the linker must find in its search path. This should not be a cache entry (note that this also means this variable should not be used as the result variable of a [find_library()](https://cmake.org/cmake/help/latest/command/find_library.html#command:find_library) command - see Xxx_LIBRARY below for that).

> The following names should not usually be used in CMakeLists.txt files. They are intended for use by Find modules to specify and cache the locations of specific files or directories. Users are typically able to set and edit these variables to control the behavior of Find modules (like entering the path to a library manually):
>
>Xxx_LIBRARY
>
>    The path of the library. Use this form only when the module provides a single library. It is appropriate to use this as the result variable in a [find_library()](https://cmake.org/cmake/help/latest/command/find_library.html#command:find_library) command.

This is also the approach the official find module uses the variable (see https://github.com/Kitware/CMake/blob/b541073a447a21c9117375e24d2db6af37dd6cfc/Modules/FindSQLite3.cmake). The variable `SQLite3_LIBRARIES` is only the result of the find module and should not be set by the user.

The problem I see is that some of SOCI's find modules follow this convention, others (as is the case with SQLite3) do not.
I can't reproduce the problem, so the reason behind SOCI not finding the SQLite3 libraries when passing `SQLITE3_LIBRARIES` is not quite clear to me. But maybe it has something to do with the combination of

```cmake
set(SQLITE3_LIBRARIES
  ${SQLITE3_LIBRARIES}
  ${SQLITE3_LIBRARY})
```

and caching both `SQLITE3_LIBRARIES` and `SQLITE3_LIBRARY`? But fixing that is a topic for another PR I guess.

Regarding the renaming of the variable: I'd probably not do that for two reasons - for now.
(1) It would break backwards compatibility as CMake variables are case-sensitive.
(2) For consistency, all other variables a user can set and all result variables should probably also be renamed then.
But that's something I'd tackle in a general modernization approach to SOCI's find modules.

TL;DR: I don't see any reason against the changes in this PR.

vadz:
@zann1x Sorry, could you please clarify the meaning of your TL;DR as I'm not completely sure I parse it correctly: do you mean that this PR should be applied or that it shouldn't?

zann1x:
Sorry for not being clear - the PR can be applied.

vadz:
Thanks, will merge in a moment.

PR #966, opened 2022-05-12
Properly support x_xmltype for sqlite3

ccosmin:


vadz:
Thanks, looks good to me but I wonder why do we need SQLite-specific XML tests when there doesn't seem to be anything actually specific to SQLite in them. Am I missing something or should we just override `table_creator_xml()` in `test_context` in `test-sqlite3.cpp` and run the existing XML tests for it instead?

P.S. If we do need/want to keep these tests, the `DropTable` class would need to be made uncopyable to avoid MSVS warnings breaking the AppVeyor CI build.

vadz:
Great, thanks! I'm going to (squash) merge it in a moment.

PR #968, opened 2022-05-18
sqlite3: support the pragma foreign_keys for enforcing foreign keys

ccosmin:
By default sqlite3 will not enforce foreign key (added a test for that).
It's opt-in functionality that must be enabled at each connection open.

Added connection parameter foreign_keys for the sqlite3 backend.

For reference:
https://www.sqlite.org/foreignkeys.html

### DIFF
--- a/docs/backends/sqlite3.md
+++ b/docs/backends/sqlite3.md
@@ -53,6 +53,7 @@ The set of parameters used in the connection string for SQLite is:
 * `synchronous` - set the pragma synchronous flag ([link](http://www.sqlite.org/pragma.html#pragma_synchronous))
 * `shared_cache` - should be `true` ([link](http://www.sqlite.org/c3ref/enable_shared_cache.html))
 * `vfs` - set the SQLite VFS used to as OS interface. The VFS should be registered before opening the connection, see [the documenation](https://www.sqlite.org/vfs.html)
+* `foreign_keys` - set the pragma foreign_keys flag ([link](https://www.sqlite.org/pragma.html#pragma_foreign_keys)). Possible values are "on"/"off"
 
 Once you have created a `session` object as shown above, you can use it to access the database, for example:
 
--- a/docs/backends/sqlite3.md
+++ b/docs/backends/sqlite3.md
@@ -53,7 +53,7 @@ The set of parameters used in the connection string for SQLite is:
 * `synchronous` - set the pragma synchronous flag ([link](http://www.sqlite.org/pragma.html#pragma_synchronous))
 * `shared_cache` - should be `true` ([link](http://www.sqlite.org/c3ref/enable_shared_cache.html))
 * `vfs` - set the SQLite VFS used to as OS interface. The VFS should be registered before opening the connection, see [the documenation](https://www.sqlite.org/vfs.html)
-* `foreign_keys` - set the pragma foreign_keys flag ([link](https://www.sqlite.org/pragma.html#pragma_foreign_keys)). Possible values are "on"/"off"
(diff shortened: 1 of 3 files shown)


docs/backends/sqlite3.md
```diff
@@ -53,6 +53,7 @@ The set of parameters used in the connection string for SQLite is:
 * `synchronous` - set the pragma synchronous flag ([link](http://www.sqlite.org/pragma.html#pragma_synchronous))
 * `shared_cache` - should be `true` ([link](http://www.sqlite.org/c3ref/enable_shared_cache.html))
 * `vfs` - set the SQLite VFS used to as OS interface. The VFS should be registered before opening the connection, see [the documenation](https://www.sqlite.org/vfs.html)
+* `foreign_keys` - set the pragma foreign_keys flag ([link](https://www.sqlite.org/pragma.html#pragma_foreign_keys)). Possible values are "on"/"off"
```
vadz, line 56: I'd remove this part:
```suggestion
* `foreign_keys` - set the pragma foreign_keys flag ([link](https://www.sqlite.org/pragma.html#pragma_foreign_keys)).
```
because there are more possible values than this (any of 1, yes, true and on works) and we document `true` for `shared_cache` above and don't document anything at all for `synchronous`.

If we want to improve this, we should document the full set of possible values for all the options and also accept all of them for `shared_cache`, but for now I'd just avoid giving slightly misleading information.

src/backends/sqlite3/session.cpp
```diff
@@ -126,6 +130,12 @@ sqlite3_session_backend::sqlite3_session_backend(
         execude_hardcoded(conn_, query.c_str(), errMsg.c_str());
     }
 
+    if (!foreignKeys.empty())
+    {
+        std::string const query("pragma foreign_keys=" + foreignKeys);
+        execude_hardcoded(conn_, query.c_str(), "Attempt to set foreign_keys pragma failed");
```
vadz, line 136: I think it would be useful to show the value here, e.g.
```suggestion
        execude_hardcoded(conn_, query.c_str(), "Executing " + query + " failed");
```

tests/sqlite3/test-sqlite3.cpp
```diff
@@ -53,6 +53,62 @@ TEST_CASE("SQLite rowid", "[sqlite][rowid][oid]")
     sql << "drop table test1";
 }
 
+static void SetupForeignKeysTest(soci::session& sql)
```
vadz, line 56: It would be better to make this a class doing this in its ctor and executing `drop table xxx` in its dtor to ensure that the tables don't stay in the database even if an exception is thrown.

tests/sqlite3/test-sqlite3.cpp
```diff
@@ -53,6 +53,62 @@ TEST_CASE("SQLite rowid", "[sqlite][rowid][oid]")
...
+    bool thrown = false;
+    try
+    {
+        sql << "delete from parent where id = 1";
+    }
+    catch (const soci::soci_error&)
+    {
+        thrown = true;
+    }
+    CHECK(thrown);
```
vadz, line 106: This should be replaced with just `CHECK_THROWS_AS()` or, if possible, with `CHECK_THROWS_WITH` to also verify that the _expected_ exception is thrown and not something else.

vadz (commented):
Thanks! No real comments, but I think a couple of minor things could be improved here before merging it.

vadz:
Thanks for the updates but there is no some (valid looking) errors in the CI, could you please check them? The table setup class needs to be made non-copyable, but maybe there is something else...

vadz:
Thanks for the fix, will (squash) merge in a moment!

PR #971, opened 2022-06-15
[sqlite3] Improve get_last_insert_id

ccosmin:
Take the value from the specified table as argument.
If table column was marked AUTOINCREMENT use that by checking the internal sqlite_sequence table. If the AUTOINCREMENT is not present or nothing was inserted into a table, just take the max(rowid) of that table. This has the disadvantage that ids are recycled when rows are deleted (as test shows). However, if this is in anyway important to the caller then the columns will be marked as AUTOINCREMENT given this is the solution the documentation specifies.

The previous version was of reduced utility, it was a simple wrapper around the sqlite3_last_insert_rowid which would return a *global* last insert rowid across the whole database connection. Presumably this was not the original purpose of the API given that the table name parameter is present.

### DIFF
--- a/src/backends/sqlite3/session.cpp
+++ b/src/backends/sqlite3/session.cpp
@@ -52,9 +52,29 @@ void check_sqlite_err(sqlite_api::sqlite3* conn, int res, char const* const errM
 
 } // namespace anonymous
 
+static int CheckSequenceTableCallback(void* ctxt, int valueNum, char**, char**)
+{
+    bool* flag = (bool*)ctxt;
+    *flag = valueNum > 0;
+    return 0;
+}
+
+static bool SequenceTableExists(sqlite_api::sqlite3* conn)
+{
+    char *zErrMsg = 0;
+    bool sequenceTableExists = false;
+    std::string query = "select name from sqlite_master where type='table' and name='sqlite_sequence'";
+    int const res = sqlite3_exec(conn, query.c_str(), &CheckSequenceTableCallback,
+                                 &sequenceTableExists,
+                                 &zErrMsg);
+    check_sqlite_err(conn, res, "Failed checking if the sqlite_sequence table exists");
+
+    return sequenceTableExists;
+}
 
 sqlite3_session_backend::sqlite3_session_backend(
     connection_parameters const & parameters)
+    : sequenceTableExists_(false)
 {
     int timeout = 0;
     int connection_flags = SQLITE_OPEN_READWRITE | SQLITE_OPEN_CREATE;
@@ -133,13 +153,11 @@ sqlite3_session_backend::sqlite3_session_backend(
     if (!foreignKeys.empty())
     {
         std::string const query("pragma foreign_keys=" + foreignKeys);
-        std::string const errMsg("Executing query: " + query + " failed");
-        execude_hardcoded(conn_, query.c_str(), errMsg.c_str());
+        execude_hardcoded(conn_, query.c_str(), "Attempt to set foreign_keys pragma failed");
     }
 
     res = sqlite3_busy_timeout(conn_, timeout * 1000);
     check_sqlite_err(conn_, res, "Failed to set busy timeout for connection. ");
-
 }
 
 sqlite3_session_backend::~sqlite3_session_backend()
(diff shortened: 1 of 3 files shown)


src/backends/sqlite3/session.cpp
```diff
@@ -162,10 +180,58 @@ void sqlite3_session_backend::rollback()
...
+{
+    long long value_;
+    bool filledIn_;
+};
+
+static int GetOneLong(void* ctxt, int valueNum, char** values, char**)
+{
+    SeqCtxt* seqCtxt = (SeqCtxt*)ctxt;
+    seqCtxt->value_ = 0;
+    seqCtxt->filledIn_ = true;
```
vadz, line 193: Shouldn't this be only done inside the `if` below? Or do we intentionally do it even if there are no values? If so, could you please add a comment explaining this and why do we do it?

src/backends/sqlite3/session.cpp
```diff
@@ -162,10 +180,58 @@ void sqlite3_session_backend::rollback()
...
+        {
+            value = seqCtxt.value_;
+            return true;
+        }
+    }
+
+    // Fall-back just get the maximum rowid of what was already inserted in the
+    // table. This has the disadvantage that if rows were deleted, then ids may be re-used.
+    // But, if one cares about that, AUTOINCREMENT should be used anyway.
+    std::string const maxRowIdQuery = "select max(rowid) from " + table;
```
vadz, line 231: Shouldn't table be at least quoted here?

src/backends/sqlite3/session.cpp
```diff
@@ -162,10 +180,58 @@ void sqlite3_session_backend::rollback()
...
+    SeqCtxt seqCtxt;
+    seqCtxt.filledIn_ = false;
+    seqCtxt.value_ = 0;
+    if (sequenceTableExists_ || SequenceTableExists(conn_))
+    {
+        // Once the sqlite_sequence table is created (because of a column marked AUTOINCREMENT)
+        // it can never be dropped, so don't search for it again.
+        sequenceTableExists_ = true;
+
+        std::string const query = "select seq from sqlite_sequence where name ='" + table + "'";
```
vadz, line 213: I'm a bit nervous with using table name as part of the query. It's probably unlikely that it wouldn't be under program control, but if it can ever come from an untrusted source, we've got an SQL injection here.

src/backends/sqlite3/session.cpp
```diff
@@ -162,10 +180,58 @@ void sqlite3_session_backend::rollback()
...
 
+struct SeqCtxt
+{
+    long long value_;
+    bool filledIn_;
+};
+
+static int GetOneLong(void* ctxt, int valueNum, char** values, char**)
+{
+    SeqCtxt* seqCtxt = (SeqCtxt*)ctxt;
```
vadz, line 191:
```suggestion
    SeqCtxt* const seqCtxt = static_cast<SeqCtxt*>(ctxt);
```

src/backends/sqlite3/session.cpp
```diff
@@ -162,10 +180,58 @@ void sqlite3_session_backend::rollback()
     execude_hardcoded(conn_, "ROLLBACK", "Cannot rollback transaction.");
 }
 
+struct SeqCtxt
+{
+    long long value_;
+    bool filledIn_;
```
vadz, line 186: Very minor, but I'd call this `valid` as IMHO this is even more clear.

vadz (commented):
Thanks, this looks globally good, but:

1. Could you please use SOCI naming conventions, AFAICS all SQLite3 backend code, at least, does follow them and so using a different convention in this PR is a bit jarring.
2. Fix `zErrMsg` memory leaks.
3. Fix MSVC warning so that the CI builds pass.

TIA!

src/backends/sqlite3/session.cpp
ccosmin: No, no need, the rowid is part of the table.

vadz: Just to be clear: this is a continuation of my comment about SQL injection above, i.e. I'd prefer not to use `table` as part of the query, but if we use it, it should at least be `"'" + table + "'"`, I think.

ccosmin:
Please review this? I think all tests are green.

vadz:
Thanks for the changes! I've renamed a few more variables/functions and simplified the code a bit by reusing the existing `execute_hardcoded()`.

Please let me know if you have any comments or objections, otherwise I'll merge it soon.

vadz:
Merged now, thanks!

PR #976, opened 2022-06-22
Map dt_date to datetime in SQLite3's DDL API

### LINKED ISSUE #969: Sqlite3 dt_date usage
Creating a table using the generic code
```
soci::ddl_type ddl = db.create_table(name);
ddl.column("TIMESTAMP", soci::dt_date)("not null");
```

a new table is created with the column type INTEGER instead of the expected type "DATE".

When querying data from this column the following code an std::bad_cast error is thrown: 
```
std::tm timestamp;
statement st = (db.prepare << "select TIMESTAMP from name, into(timestamp));
st.execute();
```
This might be due to this part of code: https://github.com/SOCI/soci/blob/master/include/soci/sqlite3/soci-sqlite3.h#L319
I would expect for "dt_date" a mapping to "date" or something similar.

The corresponding std::tm test for sqlite3 succeeds, because it is not using ddl_type but rather a direct sql command for creating the table: https://github.com/SOCI/soci/blob/master/tests/sqlite3/test-sqlite3.cpp#L379

zann1x:
As the datetime data type doesn't exist in SQLite3, the DDL
method create_column_type() mapped dt_date to an integer type. This was
possible because SQLite3 internally handles datetime values as text,
real or integer. When using the soci::row API however, we want to be
able to query datetime columns as std::tm instead of integers.
To do this, create_column_type() must assign dt_date to a type that
gets mapped back to dt_date in get_data_type_map() in
sqlite3/statement.cpp. The most generic one in this case is datetime.

Fixes #969

### DIFF
--- a/include/soci/sqlite3/soci-sqlite3.h
+++ b/include/soci/sqlite3/soci-sqlite3.h
@@ -317,6 +317,12 @@ struct sqlite3_session_backend : details::session_backend
             case dt_double:
                 return "real";
             case dt_date:
+                // The type `datetime` doesn't exist in SQLite3. Instead,
+                // the storage class of date and time data types is either
+                // one of TEXT, REAL or INTEGER. However, when using the
+                // soci::row API, we want to be able to map the column back 
+                // to a dt_date, so that we can query its value as a std::tm.
+                return "datetime";
             case dt_integer:
             case dt_long_long:
             case dt_unsigned_long_long:
--- a/tests/sqlite3/test-sqlite3.cpp
+++ b/tests/sqlite3/test-sqlite3.cpp
@@ -541,6 +541,34 @@ TEST_CASE("SQLite std::tm bind", "[sqlite][std-tm-bind]")
     CHECK(std::mktime(&result.front()) == std::mktime(&datetime));
 }
 
+TEST_CASE("SQLite std::time bind after DDL", "[sqlite][std-tm-bind][ddl]")
+{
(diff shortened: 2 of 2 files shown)


vadz:
Thanks, but just to confirm: is it really fine to create columns with any type at all? What about when using strict option?

zann1x:
You can define STRICT on a per-table basis. If a table is defined as STRICT, the datatypes are restricted to INT, INTEGER, REAL, TEXT, BLOB and ANY (https://www.sqlite.org/stricttables.html). Without STRICT, SQLite is very liberal about the values given and the values stored:

> For example, if a table column has a type of "INTEGER", then SQLite tries to convert anything inserted into that column into an integer. So an attempt to insert the string '123' results in an integer 123 being inserted. But if the content cannot be losslessly converted into an integer, for example if the input is 'xyz', then the original string is inserted instead.

Digging a little deeper into SOCI, I have to admit that I didn't see that use create_column_type(), add_column() etc. are publicly usable on session_backend. Creating a DDL string that way and then adding STRICT manually would be possible. I suppose that one could even add STRICT when using ddl_type::set_tail().
My change would probably break that behavior :confused:

zann1x:
I'm closing this PR because adding `strict` to the DDL string indeed results in an exception with message

```
sqlite3_statement_backend::prepare: unknown datatype for soci_test.ts: "datetime" while preparing "create table soci_test (ts datetime not null) strict
```

PR #981, opened 2022-07-13
Fix having both "use" and "into" in a vector object of a custom type.

ccosmin:
Previous behavior was that the vector was reset to zero size.

I'm the first to admit that I don't know what was the use of the commented out code in statement.cpp:134.
For my test case both number and rows were equal to 1 so the load_rowset was not executed.

### DIFF
--- a/src/backends/sqlite3/statement.cpp
+++ b/src/backends/sqlite3/statement.cpp
@@ -311,7 +311,7 @@ sqlite3_statement_backend::bind_and_execute(int number)
 
         // Handle the case where there are both into and use elements
         // in the same query and one of the into binds to a vector object.
-        if (1 == rows /* && number != rows */)
+        if (1 == rows && number != rows)
         {
             return load_rowset(number);
         }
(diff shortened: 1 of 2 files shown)


vadz:
Just in case you didn't notice it, there are some build errors.

Also, there are some helpers in `include/private/soci-mktime.h` that can be used instead of testing for `_MSC_VER` directly, which would be nice (although right now I'm not sure at all why do we need timestamps for this test, couldn't we just use something else?).

vadz:
Sorry, I'm somewhat confused about what does this PR do now: it doesn't seem to fix anything, it does add some tests, but it adds them to SQLite test suite, when they don't seem to be really SQLite-specific...

Is there still a problem with vectors (and custom types?) in SQLite? It would be nice to at least document it, if it can't be (easily) fixed, but I just don't understand what it is, could you please explain? TIA!

ccosmin:
Yes, sorry, I don't think this PR is mergeable yet.

I removed the non-test code because it seems reading "into" 1-sized vectors is not supported (test even crashes) and one should do "into" scalar in that case. At least this is my reading of the code.
I couldn't find any documentation on this and I don't know if this specific to sqlite or other backends are affected as well.

vadz:
FWIW I think reading into vectors of size 1 should work, and if it doesn't, it must be a bug (not sure if it happens with SQLite only or with the other backends too?), but as this is not mergeable yet, I'm going to close it for now, but, of course, please reopen it or recreate another PR if you can fix the actual bug. TIA!

PR #982, opened 2022-08-05
Make CMake subdirectory usage easier

samh:
These tweaks will allow users to use SOCI as a subdirectory with CMake the same way as if they had called `find_package` (targets in the `SOCI::` namespace, and no additional call to `target_include_directories`).

Everything should be backwards-compatible - the internal non-namespace targets continue to work as they did before, and if someone used `target_include_directories` as before it wouldn't break anything.

I also updated the documentation and examples to reflect the simplified usage.

Please let me know if you have any comments or want me to change anything. Thanks!


vadz:
The new way to use SOCI in a subdirectory is definitely much nicer to me than the old one, thanks for improving this!

If there are no objections, I'm going to merge this in a couple of days.

Thanks again!

Tectu:
Any news/update on this? :)

vadz:
You're right, this should have been merged already -- done now.

PR #986, opened 2022-09-14
Documentation: Fix typo

Tectu:
Fix a minor typo in `docs/types.md`

### DIFF
--- a/docs/types.md
+++ b/docs/types.md
@@ -142,7 +142,7 @@ The extension mechanism relies on appropriate specialization of the `type_conver
 
 There are three required class members for a valid `type_conversion` specialization:
 
-* the `base_type` type definition, aliasing either one of the base types *or another ser-defined type*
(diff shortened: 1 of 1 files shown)


vadz:
Thanks for the fix! I will merge it soon.

Tectu:
Closing != Merging :p

vadz:
I prefer not to create merge commits for the single commit branches as they seem useless to me and GitHub doesn't provide any way to mark a PR as "merged" if it was cherry-picked from outside the web UI (which I'd rather not use if possible), so, unfortunately, we have to live with this side effect. I don't think it makes any difference in practice, does it?

Tectu:
> I don't think it makes any difference in practice, does it?

Nah, I was just wondering whether you maybe accidentally hit the "comment & close" button which might lead to this PR being forgotten about as it wouldn't show up as an open/active PR anymore :D

PR #990, opened 2022-09-14
Rename library aliases to use shorter and more consistent names

vadz:
Use "Soci::" rather than "SOCI::" as the prefix, for consistency with the other CMake variables defined by FindSoci.cmake.

Do not use "soci_" prefix for the library names which is completely redundant with the other prefix, writing "Soci::soci_" is clearly redundant.

See #982.

### DIFF
--- a/cmake/SociBackend.cmake
+++ b/cmake/SociBackend.cmake
@@ -155,7 +155,7 @@ macro(soci_backend NAME)
           SHARED
           ${THIS_BACKEND_SOURCES}
           ${THIS_BACKEND_HEADERS})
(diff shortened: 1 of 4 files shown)


vadz:
@samh and @Krzmbrzl please let me know if you have any objections to changing it like this.

samh:
I'm neutral on `SOCI` vs `Soci`, so either way is fine with me. I like removing the extra `soci_` prefix; I think it fits with current CMake conventions (e.g. Qt4 had `Qt4::QtDBus`, but for Qt5 it changed to `Qt5::DBus`).

Krzmbrzl (approved):
LGTM

Issue #499, opened 2016-08-30
Blob offset not uniform between backends


msobczak:
Hi,
The read/write operations for BLOB objects expect the "offset" parameter, which is not handled uniformly across backends.
In particular, Oracle expects 1 to denote the first byte of the BLOB object, wherease PostgreSQL expects 0 as the beginning index. This is reflected in the existing unit test cases.
This makes it more difficult to write backend-agnostic applications.

Is this considered to be a problem?

mloskot:
Although I don't recall any issues reported, unification would be nice, I guess.

msobczak:
The big question here is whether the unconditional unification (that is, changing the current behaviour for one of the databases) would break any existing application.
It might be possible to introduce a configuration parameter (in a connection string, perhaps) with the intended starting index for those applications that care to have it unified. The lack of this additional parameter would preserve the existing behaviour, which would protect existing code that relies on one or the other index value.

mloskot:
AFAICT, those involved, developers and users, have accepted SOCI 4 may break compatibility in many places. That is one of the reasons SOCI 4 (current `master`) has been running for long time now without packaged release.

vadz:
API breakage is one thing and acceptable in a major release. Silent breakage like this is another matter entirely and I'm less sure about it.

mloskot:
@vadz I consider parameters of public API functions are part of the API. I also consider a parameter domain (e.g. valid values `0` or `1`) part of the API. Hence, I wouldn't call such change a silent breakage (I also assume it would be documented).

vadz:
@mloskot Sorry, I should have made it clear that I use "silent breakage" in its usual, well-defined, meaning of "change in behaviour not resulting in any compilation error or warnings". And this would be exactly this, which is rightly considered to be the worst possible kind of incompatibility: it's much better to prevent the code from compiling, than to let it compile, but do the wrong thing silently.

Because of this, unless we have a very good reason to believe that this API is not used at all, my ideal choice would be to add some new consistent API and deprecate the existing one. If this is considered too drastic/time-consuming, then I prefer @msobczak's proposal of keeping the current behaviour by default and adding an option. This is, IMHO, less elegant than a new API but simpler to implement.

But personally I'd always do just about anything else rather than silently change an API behaviour. I realize that this is not a universally shared point of view and SOCI may use a different approach, of course.

mloskot:
@vadz 
> "change in behaviour not resulting in any compilation error or warnings"

[...]

PR #991, opened 2022-09-15
[read|write]_from_start implemented for multiple backends

Krzmbrzl:
#508 added the functions `read_from_start` and `write_from_start` to the BLOB interface in order to work around the issue that the offsets in the plain `read` and `write` functions in the interface are backend-dependent (e.g. some start from 0 and some from 1).

However, that PR only implemented these functions for Postgresql and Oracle and for all other backends, the default implementation throws an exception stating that the function was not implemented/available.

Looking at existing test-cases involving read/write operations on BLOBs, I extracted the expected offset specification for a couple more backends and implemented the `read_from_start` and `write_from_start` functions accordingly.

Note: As I see it, these functions are now implemented for all backends that Soci supports BLOBs for in the first place.

TODO:
- [x] Make sure all test still pass (will have to use CI for that)
- [x] ~~Extend BLOB documentation to mention backend-specific offsets and that read/write from start is preferred (where available)~~ Current docs already covers everything

### DIFF
--- a/include/soci/blob.h
+++ b/include/soci/blob.h
@@ -32,6 +32,7 @@ public:
     std::size_t get_len();
 
     // offset is backend-specific
+    [[deprecated("Use read_from_start instead")]]
     std::size_t read(std::size_t offset, char * buf, std::size_t toRead);
 
     // offset starts from 0
@@ -39,6 +40,7 @@ public:
         std::size_t offset = 0);
 
     // offset is backend-specific
+    [[deprecated("Use write_from_start instead")]]
     std::size_t write(std::size_t offset, char const * buf,
         std::size_t toWrite);
 
--- a/include/soci/db2/soci-db2.h
+++ b/include/soci/db2/soci-db2.h
@@ -229,8 +229,8 @@ struct db2_blob_backend : details::blob_backend
     ~db2_blob_backend() override;
 
     std::size_t get_len() override;
-    std::size_t read(std::size_t offset, char* buf, std::size_t toRead) override;
-    std::size_t write(std::size_t offset, char const* buf, std::size_t toWrite) override;
+    std::size_t read_from_start(char* buf, std::size_t toRead, std::size_t offset = 0) override;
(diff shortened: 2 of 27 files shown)


vadz:
Thanks, I don't see anything wrong with these changes, but I'm rather confused by the (co)existence of both `read`/`write` and their `_from_start` versions -- does it really make sense to have both? Should we perhaps remove or deprecate `read` and `write`, if they can't be used portably anyhow?

4b0dbab3e removed the mention of these functions from `docs/exchange.md`, which seems to imply that they're not supposed to be used/exist any more, but I'm not really sure...

Krzmbrzl:
It is indeed confusing. Based on https://github.com/SOCI/soci/issues/499 the original functions were not removed in order to maintain backwards compatibility with old code.
I don't have a problem with removing the old functions, though :shrug:

vadz:
Thanks, I completely forgot about the discussion there. It's true that we had to preserve compatibility in the public API, but we still probably should deprecate `read()` and `write()` now, i.e. in 4.1.

And from this point of view it would actually make more sense to implement `blob_backend::read()` in terms of `read_from_start()` by default, rather than vice versa (and the same thing for `write()` too, of course), what do you think? This would require more changes, but would allow us to deprecate `read()` more easily in the future (or even now).

But even if we don't do it like this but keep implementing `read_from_start()` in terms of `read()`, it would still be better to do it in `details::blob_backend` itself instead of doing it in all backends except for Oracle IMO.

Krzmbrzl:
I adapted my code to your review. I have now deprecated all old plain read/write functions and made sure there always is a *_from_start alternative implemented instead.

Furthermore, I took the liberty of making sure that non-implemented BLOB backends now consistently throw exceptions instead of just returning dummy values to make it immediately obvious to users that what they are using is not implemented.

Krzmbrzl:
Adapted to the new C++14 availability

Krzmbrzl:
@vadz is there still something that you want me to do,  before this can be merged?

vadz:
Sorry for the delay with this, I've finally looked at and tested this and don't see any problems with it, so I'll merge it soon.

Thanks again!

PR #993, opened 2022-09-16
Make soci::row movable

### LINKED ISSUE #900: std::vector<soci::row>::resize()  The specified type does not meet the requirements of Cpp17MoveInsertable
```
      using namespace soci;
      constexpr int BATCH_SIZE = 30;
      std::vector<row> rows(BATCH_SIZE);
      statement st = (sql.prepare << "select * from table", into(rows));
      st.execute();
      while (st.fetch()) {
         std::vector<int>::iterator pos;
         for (pos = valsOut.begin(); pos != valsOut.end(); ++pos) {
            cout << *pos << '\n';
         }
         rows.resize(BATCH_SIZE); //ERROR
      }
```
```
In template: static_assert failed due to requirement '__is_cpp17_move_insertable<std::allocator<soci::row>, false>::value' "The specified type does not meet the requirements of Cpp17MoveInsertable"
```

Krzmbrzl:
Fixes #900

----

Note: I haven't actually verified that this fixes the linked issue. However, based on the comment in there that all that is needed was to make `soci::row` movable, this PR should indeed fix it.

### DIFF
--- a/include/soci/row.h
+++ b/include/soci/row.h
@@ -43,6 +43,9 @@ public:
     row();
     ~row();
 
+    row(row &&other) = default;
+    row &operator=(row &&other) = default;
+
     void uppercase_column_names(bool forceToUpper);
     void add_properties(column_properties const& cp);
     std::size_t size() const;
--- a/tests/common-tests.h
+++ b/tests/common-tests.h
@@ -2469,6 +2469,16 @@ TEST_CASE_METHOD(common_tests, "Dynamic row binding", "[core][dynamic]")
 
         CHECK(r.get_properties(0).get_data_type() == dt_string);
         CHECK(r.get_properties(1).get_data_type() == dt_integer);
+
+        // Check if row object is movable
+        row moved = std::move(r);
+
+        CHECK(moved.size() == 2);
+        // We expect the moved-from row to become empty after the move operation
+        CHECK(r.size() == 0);
+
+        CHECK(moved.get_properties(0).get_data_type() == dt_string);
+        CHECK(moved.get_properties(1).get_data_type() == dt_integer);
     }
 }
 


include/soci/row.h
```diff
@@ -43,6 +44,11 @@ class SOCI_DECL row
     row();
     ~row();
 
+#ifdef SOCI_HAVE_CXX11
+    row(row &&other);
+    row &operator=(row &&other);
```
vadz, line 49: Wouldn't just the following work?

```suggestion
    row(row &&other) = default;
    row &operator=(row &&other) = default;
```

vadz (commented):
It would be useful to have a unit test for this to check that it works as expected.

include/soci/row.h
Krzmbrzl: No, because it would just perform a member-wise move. However, the moved-from object remains in an indeterminate state (not guaranteed to be empty in case of vectors) and since the destructor accesses these objects again, we would run into undefined behavior. Thus, we have to explicitly clear the vectors to being them into a defined state again.

Krzmbrzl:
Adapted this PR to the latest Cpp14 availability.

include/soci/row.h
vadz: Oh, you're right in principle, but this is so, so annoying: we now need to write much more code just to do the same thing that the compiler is almost certainly doing by default anyhow. I also find [Howard's analysis here](https://stackoverflow.com/a/17735913/15275) convincing and it seems to me that as long as we're not using different allocators (and we don't and never will) this can't actually happen in practice.

Am I missing anything here?

src/core/row.cpp
```diff
@@ -26,6 +26,31 @@ row::~row()
     clean_up();
 }
 
+#ifdef SOCI_HAVE_CXX11
```
vadz, line 29: _If_ we keep this code at all, we don't need this test any longer in master.

Krzmbrzl: Oh, I intended to remove this as I did in the header file.

include/soci/row.h
Krzmbrzl: I wouldn't want to rely on interpreting something like this and rely on it. Especially not as the workaround to simply conform to the standard is so trivial.

vadz: I disagree. The workaround is verbose and needs maintenance (e.g. will need to be updated if another member is added to the class). Using default implementations is simple and I'm willing to eat my (non existent, but that's a detail) hat if a vector remains non-empty in any existing or future implementation.

Krzmbrzl: Alright - your call (and responsibility :P) :shrug:

Krzmbrzl:
There you go

Krzmbrzl:
Ha! It seems that the test case that I added proofs that it is not a good idea to rely on a specific state of a moved-from object.
Apparently the `_holders` vector (which is queried in the `size` function) is not empty after having been moved.

Krzmbrzl:
@vadz what do you want me to do about the failing test cases?

vadz:
I'm sorry but I don't understand what happened here: you've added a check that that the moved-from vector is _not_ empty, and it fails, just as expected.

Why did you add this check instead of `CHECK( r.empty() )`?

Krzmbrzl:
Lol - apparently I trusted my intention and my comment more than the actual code that I wrote. This is a C&P error that slipped in. Obviously, this check was nonsense :facepalm: 

> Why did you add this check instead of CHECK( r.empty() )?

Because afaict `soci::row` does not have a function `empty`. I changed it to `size() == 0` instead.

PR #994, opened 2022-09-16
soci::blob: Use unique_ptr 

### LINKED ISSUE #957: Blob and copy/move constructor
Blob has no explicit copy/move constructor but a destructor which calls delete. So making a copy/move can result in double free or use after free.
example:
```cpp
  soci::session sql (soci::sqlite3, databaseName);
  auto originalBlob = soci::blob{ sql };
  auto copyBlob = originalBlob;
```
```
AddressSanitizer:DEADLYSIGNAL
=================================================================
==46681==ERROR: AddressSanitizer: SEGV on unknown address 0x00000700000a (pc 0x55f275330f88 bp 0x7fff806d8ef0 sp 0x7fff806d8ee0 T0)
==46681==The signal is caused by a READ memory access.
    #0 0x55f275330f88 in soci::blob::~blob() /home/walde/.conan/data/soci/4.0.2/werto87/stable/build/7ef31e0413043c7eb48a64bb0d0adb2c77e9639e/source_subfolder/src/core/blob.cpp:23:5
    #1 0x55f2752e611f in main /home/walde/projects/sandbox/cxx_template/main.cxx:129:1
    #2 0x14f42d6f030f in __libc_start_call_main (/usr/lib/libc.so.6+0x2d30f)
    #3 0x14f42d6f03c0 in __libc_start_main_impl (/usr/lib/libc.so.6+0x2d3c0)
    #4 0x55f2751fdd14 in _start (/home/walde/projects/sandbox/cxx_template/build/bin/project+0x32d14)

AddressSanitizer can not provide additional info.
SUMMARY: AddressSanitizer: SEGV /home/walde/.conan/data/soci/4.0.2/werto87/stable/build/7ef31e0413043c7eb48a64bb0d0adb2c77e9639e/source_subfolder/src/core/blob.cpp:23:5 in soci::blob::~blob()
==46681==ABORTING
```
Possible solution:
remove destructor of blob and replace "details::blob_backend * backEnd_;" with std::unique_ptr < details::blob_backend >  backEnd_;

Problem:
This is api breaking

Krzmbrzl:
If a blob object was copied, on destruction a double-free would occur.

Fixes #957

Also make blobs movable in CXX11 or more recent

### DIFF
--- a/src/core/blob.cpp
+++ b/src/core/blob.cpp
@@ -14,13 +14,8 @@
 using namespace soci;
 
 blob::blob(session & s)
+	: backEnd_{s.make_blob_backend()}
 {
-    backEnd_ = s.make_blob_backend();
-}
-
-blob::~blob()
-{
-    delete backEnd_;
 }
 
 std::size_t blob::get_len()
--- a/include/soci/blob.h
+++ b/include/soci/blob.h
@@ -11,6 +11,7 @@
 #include "soci/soci-platform.h"
 // std
 #include <cstddef>
+#include <memory>
 
 namespace soci
 {
@@ -27,7 +28,10 @@ class SOCI_DECL blob
 {
 public:
     explicit blob(session & s);
-    ~blob();
+    ~blob() = default;
+
+    blob(blob &&other) = default;
+    blob &operator=(blob &&other) = default;
 
     std::size_t get_len();
 
@@ -50,10 +54,12 @@ public:
 
     void trim(std::size_t newLen);
 
-    details::blob_backend * get_backend() { return backEnd_; }
+    details::blob_backend * get_backend() { return backEnd_.get(); }
 
 private:
-    details::blob_backend * backEnd_;
+    SOCI_NOT_COPYABLE(blob)
+
+    std::unique_ptr<details::blob_backend> backEnd_;
 };
 
 } // namespace soci


src/core/blob.cpp
```diff
@@ -23,6 +23,25 @@ blob::~blob()
...
+#ifdef SOCI_HAVE_CXX11
+
+blob::blob(blob &&other)
+{
+    *this = std::move(other);
+}
+
+blob &blob::operator=(blob &&other)
+{
+    backEnd_ = std::move(other.backEnd_);
```
vadz, line 35: `move()` seems useless on a pointer:
```suggestion
    backEnd_ = other.backEnd_;
```

vadz (commented):
I really think we should just use `unique_ptr backEnd_` instead as then we could implement these methods by using `= default`.

src/core/blob.cpp
Krzmbrzl: It is, but I added it for semantic reasons. Plus, should we ever switch this to a smart pointer, we don't have to adapt the code here.

Krzmbrzl:
Redid this PR with the new C++14 availability in mind

src/core/blob.cpp
```diff
@@ -14,13 +14,12 @@
 using namespace soci;
 
 blob::blob(session & s)
+	: backEnd_(std::unique_ptr<details::blob_backend>(s.make_blob_backend()))
```
vadz, line 17: Why not just
```suggestion
	: backEnd_{s.make_blob_backend()}
```
?

src/core/blob.cpp
```diff
@@ -14,13 +14,12 @@
...
 {
-    backEnd_ = s.make_blob_backend();
 }
 
 blob::~blob()
 {
-    delete backEnd_;
 }
```
vadz, line 23: I'd leave a comment here just to prevent somebody from removing this "useless" dtor later, and I'd also use `= default`, so:
```cpp
// Note that this dtor must be defined here, where blob_backend class is fully declared.
blob::~blob() = default;
```

src/core/blob.cpp
Krzmbrzl: Because it doesn't really matter and I'm not used to using brace initialization :shrug:

src/core/blob.cpp
Krzmbrzl: Yeah defaulting the dtor sounds good :+1:

src/core/blob.cpp
vadz: Well, using braces was just part of it (and is a good practice as it avoids narrowing conversions, so I don't agree that it doesn't really matter), but this was mostly about omitting constructing a temporary object unnecessarily.

Krzmbrzl: > constructing a temporary object unnecessarily

Ah yeah - that will be optimized away anyway, but yeah it's a bit verbose. I'll change it.

Krzmbrzl:
Adapted to review

Krzmbrzl:
@vadz is there still something missing in this PR?

vadz:
Sorry, I have a couple of other urgent things to finish right now, so I'm unable to look at SOCI for the time being. I do remember about this and the other PRs and will try to get them a.s.a.p.

Krzmbrzl:
Okay, no worries then :+1:

vadz:
Thanks, I'm merging this one, but with one small but important change: dtor must not be inline as `details::blob_backend` is not guaranteed to be fully defined in the header (even though it is, apparently, right now, as otherwise things wouldn't work).

vadz:
Applied now.

Issue #907, opened 2021-09-09
Update SOCI to use C++14


vadz:
The earliest compilers I'd personally like to still be supported are gcc 6 and MSVS 2017.

vadz:
This was done back in c1a4b7606f9b134b3b7e459cba8de32903fab550.

PR #996, opened 2022-09-17
Require C++14 for building SOCI

vadz:
Update the CI builds and the code as discussed in #907.


include/soci/once-temp-type.h
```diff
@@ -31,7 +31,7 @@ class SOCI_DECL once_temp_type
     once_temp_type(once_temp_type const & o);
     once_temp_type & operator=(once_temp_type const & o);
 
-    ~once_temp_type() SOCI_NOEXCEPT_FALSE;
+    ~once_temp_type() noexcept(false);
```
Krzmbrzl, line 34: Afaik `noexcept(false)` hass absolutely no effect. The boolean argument to `noexcept` only has meaning in a metaprogramming context where a function is conditionally `noexcept`. But if your ahrd-code `false` here, it seems it would be better to have no noexcpet statement at all :thinking:

include/soci/rowset.h
```diff
@@ -148,8 +148,8 @@ class rowset_impl
 
     unsigned int refs_;
 
-    const cxx_details::auto_ptr<statement> st_;
-    const cxx_details::auto_ptr<T> define_;
+    const std::unique_ptr<statement> st_;
+    const std::unique_ptr<T> define_;
     SOCI_NOT_COPYABLE(rowset_impl)
```
Krzmbrzl, line 153: This is no longer needed as the `unique_ptr` members prevent copying anyway.
Dunno if you want to keep this around to make it more explicit, though

include/soci/session.h
```diff
@@ -43,7 +43,7 @@ class SOCI_DECL session
 {
 private:
 
-    void set_query_transformation_(cxx_details::auto_ptr<details::query_transformation_function>& qtf);
+    void set_query_transformation_(std::unique_ptr<details::query_transformation_function>&& qtf);
```
Krzmbrzl, line 46: `unique_ptr`s should be passed by value to signify ownership transfer. And if you don't want to transfer ownership, you should pass the raw pointer instead of the smart-pointer (for semantic and performance reasons)

include/soci/soci-platform.h
```diff
@@ -105,45 +105,24 @@ namespace std {
...
-{
-
-#if defined(SOCI_HAVE_CXX11) || (defined(_MSC_VER) && _MSC_VER >= 1800)
-    template <typename T>
-    using auto_ptr = std::unique_ptr<T>;
-#else // std::unique_ptr<> not available
-    using std::auto_ptr;
+    #if __cplusplus < 201402L
+        #error This version of SOCI requires C++14.
+    #endif
```
Krzmbrzl, line 115: This could be dangerous, as afaik a couple of older compiler versions used to set this macro to garbage and thus hard-erroring if this macro contains an invalid value seems a bit too much. :thinking:

src/core/session.cpp
```diff
@@ -73,15 +73,15 @@ class standard_logger_impl : public logger_impl
 } // namespace anonymous
 
 session::session()
-    : once(this), prepare(this), query_transformation_(NULL),
+    : once(this), prepare(this),
       logger_(new standard_logger_impl),
       uppercaseColumnNames_(false), backEnd_(NULL),
```
Krzmbrzl, line 78: While transitioning to a modern cpp standard, it would probably also be good to replace all `NULL` with `nullptr`

include/soci/once-temp-type.h
vadz: Sorry, this is simply not true. Destructors are noexcept by default in C++11 and we absolutely need this here as this dtor can throw.

include/soci/rowset.h
vadz: I haven't decided what to do about `SOCI_NOT_COPYABLE` yet, but I'm probably going to remove it and declare the special member functions directly. It doesn't need to be done as part of this PR however.

include/soci/soci-platform.h
vadz: I've never seen any compilers setting `__cplusplus` to garbage, and if there really are any, they certainly can't be used for compiling C++11 code. Perhaps you were thinking about MSVC, which traditionally didn't set it correctly, which is why it's handled specifically above.

src/core/session.cpp
vadz: Yes, I'm going to use clang-tidy to do this later, but let's merge this one first.

include/soci/session.h
vadz: Passing by value or r-value reference is the same thing for `unique_ptr` and, also, this is a private function and so we don't really care about what it takes, so I don't see any really compelling reason to change this. And of course we do want ownership transfer here.

vadz:
I think this can be merged now. There are certainly more things to do, e.g. using `nullptr` mentioned above and more, and I'll use clang-tidy modernization checks to change this later, but this already will allow any future PRs to not have to care about C++98 compatibility, and the sooner this can happen the better.

Sorry if this results in any conflicts with the still open PRs, but hopefully they won't be difficult to resolve.

If anybody has any objections to merging this, please comment here soon. Thanks!

include/soci/once-temp-type.h
Krzmbrzl: Ah whoops - I missed that this is a dtor.

PR #1001, opened 2022-10-13
Add support for std optional

mkaes:
Support for std::optional in soci.
The handling of the compiler switches needs to be defined.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -14,7 +14,9 @@ cmake_minimum_required(VERSION 2.8...3.20 FATAL_ERROR)
 
 project(SOCI)
 
-set(CMAKE_CXX_STANDARD 14)
+if(NOT DEFINED CMAKE_CXX_STANDARD OR CMAKE_CXX_STANDARD LESS 14)
+  set(CMAKE_CXX_STANDARD 14)
+endif()
 set(CMAKE_CXX_STANDARD_REQUIRED ON)
 
 ###############################################################################
--- a/tests/common-tests.h
+++ b/tests/common-tests.h
@@ -20,6 +20,9 @@
 #include "soci/boost-fusion.h"
 #endif // BOOST_VERSION
 #endif // SOCI_HAVE_BOOST
+#if SOCI_HAS_CXX17
+#include "soci/std-optional.h"
+#endif
 
 #include "soci-compiler.h"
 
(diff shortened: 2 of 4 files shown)


CMakeLists.txt
```diff
@@ -14,7 +14,7 @@ cmake_minimum_required(VERSION 2.8...3.20 FATAL_ERROR)
 
 project(SOCI)
 
-set(CMAKE_CXX_STANDARD 14)
+set(CMAKE_CXX_STANDARD 17)
```
vadz, line 17: I think we need to do
```cmake
if(NOT "${CMAKE_CXX_STANDARD}")
  set(CMAKE_CXX_STANDARD 14)
endif()
```
and add a CI build using `-DCMAKE_CXX_STANDARD=17` to test C++17 stuff.

tests/common-tests.h
```diff
@@ -20,6 +20,9 @@
 #include "soci/boost-fusion.h"
 #endif // BOOST_VERSION
 #endif // SOCI_HAVE_BOOST
+#if defined(_WIN32) && _MSVC_LANG > 201402L || __cplusplus > 201402L
```
vadz, line 23: I think we ought to define `SOCI_HAS_CXX11` once in some header instead of repeating this unwieldy expression multiple times.

vadz (commented):
Thanks, but we don't want to force the C++ standard to 17 just yet, so this needs to be addressed in some way.

Tectu:
@mkaes Are you planning on updating the PR with the necessary changes any time soon? I too would really like to have support for `std::optional` when C++17 is available on the target.

CMakeLists.txt
vadz: Apparently this doesn't work, sorry for proposing to do it without testing, see the failure in the clang build. It looks like `CMAKE_CXX_STANDARD` is always defined, so we need to check if it's `< 14` instead, apparently, and set it to 14 then.

We also still need a C++17 CI build as otherwise the new code is never tested.

mkaes: No Problem I can change the `CMAKE_CXX_STANDARD`.
But I have no idea how to setup a CI build for c++17.

tests/common-tests.h
mkaes: I just added this define. It is currently not used.
I am not sure how to create such a define in cmake depending on the other cmake variables.

CMakeLists.txt
vadz: You can add C++17 build to `.github/workflows/ci.yml`.

tests/common-tests.h
vadz: You don't need to do it in CMake, you can just use exactly the same condition as above in some SOCI header (`soci-platform.h` probably).

vadz:
Thanks, I've merged this and added a CI workflow using C++17.

Tectu:
Thank you for your efforts, guys!

PR #1012, opened 2022-10-31
Add imported targets to FindSoci.cmake

samh:
This adds targets to the `FindSoci.cmake` module so it can be used the same way as if you used the internal CMake / exported targets. That is, for example:

```cmake
target_link_libraries(MyExecutable PUBLIC
    Soci::core
    Soci::postgresql)
```

The second commit fixes some renaming that seems to have been missed in af6155c3ccc9ba23fd0f518704555badbfa44893.

The 3rd and 4th commits are just stylistic fixes to make it more consistent, and they shouldn't change anything functionally, but let me know if want me to take those out.

It is still missing the option for static libraries. I considered trying to add that, but I wasn't sure about all the details (I think we would want to call `find_library` a second time for each library with `CMAKE_FIND_LIBRARY_SUFFIXES` altered). Still, it's possible to use `FindSoci.cmake` to find static libraries like this:

```cmake
# Temporarily override CMAKE_FIND_LIBRARY_SUFFIXES to try to find static libs
set(_soci_ORIG_CMAKE_FIND_LIBRARY_SUFFIXES ${CMAKE_FIND_LIBRARY_SUFFIXES})
set(CMAKE_FIND_LIBRARY_SUFFIXES .a)
find_package(Soci REQUIRED COMPONENTS postgresql)
set(CMAKE_FIND_LIBRARY_SUFFIXES ${_soci_ORIG_CMAKE_FIND_LIBRARY_SUFFIXES})
```


vadz:
Thanks, this looks fine (and useful) to me but, as usual, I'd appreciate any reviews from CMake-literate people, e.g. @Krzmbrzl 

But if there are no comments/objections, I'll merge it soon, thanks again!

Krzmbrzl (approved):
LGTM

vadz:
OK, thanks for looking at this (and thanks again for the fixes themselves too, of course).

PR #1016, opened 2022-11-10
Fix typo

Tectu:
Nothing too crazy, just a simple typo fix in a comment.


vadz:
Sorry, I don't think it's worth applying this, both MSVC (compiler name) and MSVS (Visual Studio, i.e. the product name) are used several places in the comments and other SOCI files and are basically interchangeable when version is not specified (compiler and product versions are different).

Tectu:
No need to apologize - all good from my side :)

PR #1022, opened 2022-11-30
Restore a Win32 build on AppVeyor CI

vadz:
The changes of 03af258a (Use more modern compiler versions for AppVeyor CI builds, 2022-09-18) made all AppVeyor CI builds 64 bits, but we still need to support 32 bit builds too, so change one of the build jobs to use 32 bits.


vadz:
Actually we never had Win32 builds on AppVeyor and it doesn't have 32 bit PotsgreSQL/MySQL libraries installed.

PR #1023, opened 2022-12-06
Include std::optional support automatically

Tectu:
This change automatically includes the `soci/std-optional.h` header file if the C++ standard is at least 17.

`std::optional` is more or less a "core feature" as of C++17 which in my opinion justifies auto-including this header. After all, similarly boost infrastructure is also included automatically (if present/enabled).

### DIFF
--- a/include/soci/soci.h
+++ b/include/soci/soci.h
@@ -55,4 +55,9 @@
 #include "soci/boost-gregorian-date.h"
 #endif // SOCI_USE_BOOST
 
+// C++17
+#ifdef SOCI_HAVE_CXX17
+#include "soci/std-optional.h"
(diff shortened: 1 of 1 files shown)


Tectu:
Any news on this?

vadz:
I'm still not sure if it's really a good idea to do it by default, to be honest, but you're right that doing it for Boost optional but not the standard version is even stranger... And as we can't stop doing it for Boost without breaking the existing code relying on it, we should probably do it for the latter too. And we could always provide some `SOCI_NO_OPTIONAL` to avoid including this header by default, I guess.

Anyhow, merged now, sorry for the delay.

PR #1026, opened 2022-12-15
Fix for supporting enum types with rowset in Postgresql backend

parvit:
Hello everyone,
i've been develpoing a feature on a project called ZoneMinder which wants to introduce SOCI to support at this time the MySQL and Postgresql databases (https://github.com/ZoneMinder/zoneminder/pull/3644).

During development i've encountered an issue when combining the rowset and custom enum types features, when they are used together a strongly typed enum fails to be selected with error:
```
unknown data type with typelem: 104811 for colNum: 1 with name: type while
executing "SELECT Type FROM soci_test;"
```

This PR addresses that issue and i hope it is can be included in the main project, let me know if there is any improvement / change that would need to be addressed.

### DIFF
--- a/include/soci/postgresql/soci-postgresql.h
+++ b/include/soci/postgresql/soci-postgresql.h
@@ -21,6 +21,7 @@
 #include "soci/connection-parameters.h"
 #include <libpq-fe.h>
 #include <vector>
+#include <unordered_map>
 
 namespace soci
 {
@@ -312,7 +313,7 @@ struct postgresql_statement_backend : details::statement_backend
 
     // the following map is used to keep the results of column
     // type queries with custom types
-    typedef std::map<unsigned long, char> CategoryByColumnOID;
+    typedef std::unordered_map<unsigned long, char> CategoryByColumnOID;
     CategoryByColumnOID categoryByColumnOID_;
 };
 
--- a/include/soci/postgresql/soci-postgresql.h
+++ b/include/soci/postgresql/soci-postgresql.h
@@ -309,6 +309,11 @@ struct postgresql_statement_backend : details::statement_backend
 
     typedef std::map<std::string, char **> UseByNameBuffersMap;
     UseByNameBuffersMap useByNameBuffers_;
+
+    // the following map is used to keep the results of column
+    // type queries with custom types
+    typedef std::map<unsigned long, char> CategoryByColumnOID;
+    CategoryByColumnOID categoryByColumnOID_;
 };
 
 struct postgresql_rowid_backend : details::rowid_backend
--- a/src/backends/postgresql/statement.cpp
+++ b/src/backends/postgresql/statement.cpp
@@ -783,7 +783,7 @@ void postgresql_statement_backend::describe_column(int colNum, data_type & type,
     {
         int form = PQfformat(result_, pos);
         int size = PQfsize(result_, pos);
-        if (form == 0 && size == -1)
+        if (form == 0 && (size == -1 || size == 4))
         {
             type = dt_string;
         }
--- a/src/backends/postgresql/statement.cpp
+++ b/src/backends/postgresql/statement.cpp
@@ -725,6 +725,21 @@ int postgresql_statement_backend::prepare_for_describe()
     return columns;
 }
 
+void throw_soci_type_error(Oid typeOid, int colNum, char category, const char* typeName )
+{
+    std::stringstream message;
+    message << "unknown data type"
+        << " for column number: " << colNum
+        << " with type oid: " << (int)typeOid;
+    if( category != '\0' )
+    {
+        message << " with category: " << category;
+    }
+    message << " with name: " << typeName;
+
+    throw soci_error(message.str());
+}
+
 void postgresql_statement_backend::describe_column(int colNum, data_type & type,
     std::string & columnName)
 {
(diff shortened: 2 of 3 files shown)


src/backends/postgresql/statement.cpp
```diff
@@ -783,7 +783,7 @@ void postgresql_statement_backend::describe_column(int colNum, data_type & type,
     {
         int form = PQfformat(result_, pos);
         int size = PQfsize(result_, pos);
-        if (form == 0 && size == -1)
+        if (form == 0 && (size == -1 || size == 4))
```
vadz, line 786: Isn't it a bit dangerous to assume that any unknown OID of length 4 is a string? In my Postgres installation
```sql
select typname from pg_type where typlen=4;
```
returns ~30 types some of them don't look very stringy.

I think that for the proper enum support we really ought to query `pg_type` and check if `typcategory=='E'` (see https://www.postgresql.org/docs/current/catalog-pg-type.html#CATALOG-TYPCATEGORY-TABLE).

vadz (commented):
Thanks for the PR and separate thanks for including the tests in it!

However I'm not sure if simply checking for the type length is the right thing to do. Or, rather, I'm sure that it isn't and it would be preferable to perform a `pg_type` lookup, but I don't know if there are any real drawbacks to it in practice or if we could still use this until somebody has time to implement a better solution.

Also, there seem to be some (new) memory leaks now, please see the failing CI builds. Could you please check this? TIA!

parvit:
@vadz 
Thanks for the quick response, the leak is actually from the tests, i didn't think it would be reported as a leak being a test but i'll fix it no problem.

I concur with you on the size check, but not knowing in depth the postgres internals it was the best i could get. I'll see if i can implement an equivalent check using the solution you indicated, if it works i'll push it also.

parvit:
@vadz could you check the updates? i think i have provided the changes you indicated.

src/backends/postgresql/statement.cpp
```diff
@@ -781,19 +781,44 @@ void postgresql_statement_backend::describe_column(int colNum, data_type & type,
...
+                message << "unknown data type with typelem: " << (int)typeOid
+                    << " for colNum: " << colNum
+                    << " with name: " << PQfname(result_, pos);
+                throw soci_error(message.str());
+            }
+
+            char* typeVal = PQgetvalue(res, 0, 0);
+            categoryByColumnOID_.insert( std::pair<unsigned long, char>( typeOid, typeVal[0] ) );
+
+            typeCategoryIt = categoryByColumnOID_.find(typeOid);
```
vadz, line 802: Probably doesn't matter much in practice (the map is small...), but it's unnecessary to call `find()` after `insert()` which already returns the iterator.

And if you wanted the code to be 100% optimal, you'd use `lower_bind()` instead of `find()` above and then insert using the iterator found by it as the hint.

include/soci/postgresql/soci-postgresql.h
```diff
@@ -309,6 +309,11 @@ struct postgresql_statement_backend : details::statement_backend
 
     typedef std::map<std::string, char **> UseByNameBuffersMap;
     UseByNameBuffersMap useByNameBuffers_;
+
+    // the following map is used to keep the results of column
+    // type queries with custom types
+    typedef std::map<unsigned long, char> CategoryByColumnOID;
```
vadz, line 315: Not a big deal, but should use `unordered_map` here IMO -- not only it's (marginally, for the maps of this size) more efficient, but it's also more clear as we don't really need this map to be sorted.

BTW, the reason for using `map` above is that this code had been written before SOCI could use C++11, but this is not the case any longer.

src/backends/postgresql/statement.cpp
```diff
@@ -781,19 +781,44 @@ void postgresql_statement_backend::describe_column(int colNum, data_type & type,
...
+            std::stringstream query;
+            query << "SELECT typcategory FROM pg_type WHERE oid=" << (int)typeOid;
+
+            PGresult* res = PQexec(session_.conn_, query.str().c_str());
+            if (PQresultStatus(res) != PGRES_TUPLES_OK) {
+                std::stringstream message;
+                message << "unknown data type with typelem: " << (int)typeOid
+                    << " for colNum: " << colNum
+                    << " with name: " << PQfname(result_, pos);
+                throw soci_error(message.str());
```
vadz, line 796: It would be nice if we could extract this into a separate function (or even lambda if you prefer) to avoid duplicating almost the same code below. We could pass `\0` as category to indicate that it's unknown here.

src/backends/postgresql/statement.cpp
```diff
@@ -781,19 +781,44 @@ void postgresql_statement_backend::describe_column(int colNum, data_type & type,
...
-        int size = PQfsize(result_, pos);
-        if (form == 0 && size == -1)
+        auto typeCategoryIt = categoryByColumnOID_.find(typeOid);
+        if( typeCategoryIt == categoryByColumnOID_.end() ) 
         {
-            type = dt_string;
+            std::stringstream query;
+            query << "SELECT typcategory FROM pg_type WHERE oid=" << (int)typeOid;
+
+            PGresult* res = PQexec(session_.conn_, query.str().c_str());
```
vadz, line 790: You should use `soci::postgresql_result` here to ensure that `PQclear()` is called. The best proof of its advantage is that you forgot to do it here, which explains the tests failures, I think (I didn't look in the details, but it seems likely).

vadz (commented):
Thanks for the update! This looks like a much better approach to me, but the memory leak needs to be fixed and, if possible, a couple of other things should be improved.

Also, could you please reformat the code using SOCI style, notably putting all the braces on their own lines? If not, I'd have to do it when applying your PR, which will result in the conflicts for you later.

And yes, I should add a `.clang-format` to this project...

include/soci/postgresql/soci-postgresql.h
parvit: Yes my thought was exactly to use the same type to avoid doing it differently, but i agree with unordered_map being better.

src/backends/postgresql/statement.cpp
parvit: sure

parvit:
@vadz I've included your comments in the change, this time i've run it with asan so its sure to not have leaks anymore.

vadz:
Thanks, looks good now, will merge soon!

vadz:
Thanks, I've (squash) merged this now with some cosmetic changes.

parvit:
@vadz thanks, just for information, can i assume already the version this will be released in will be 4.1.0 ?

vadz:
Yes, unless we decide to call the next release 5.0 but this looks unlikely for now.

PR #1032, opened 2023-01-14
adapt cmake find for dl-library and sqlite3-library

AHeimberger:
Hey, thanks for this 'database access layer'. I would like to use it within an embedded linux project, but currently face a build error due to yocto and the [rm_work bb-class](https://docs.yoctoproject.org/ref-manual/classes.html?highlight=rm_work#rm-work-bbclass). Whenever I try to build the provided [open-embedded recipe](https://layers.openembedded.org/layerindex/recipe/33349/) the build fails because it looks up for the `libdl.a` and `libsqlite3.so` in the sysroot directory of soci. To fix this I looked up the current cmake implementation and came up with the two commits.

**It is still a Work in Progress:** I have problems identifying how to make the CI and the Yocto-Project passing the build. find_package(SQLite3) uses the delivered Module from the directory which has preferred paths for the CI and I have not found a way to disable CMAKE_MODULE_PATH specifically for this SQLite3 module.


AHeimberger:
I tried different ideas to make `libsqlite3.so`  available for the ci and yocto builds
- disabling the provided FindSQLite3.cmake file with an option
- extending PATHS within FindSQLite3.cmake by extending it with [CMAKE_SYSROOT](https://cmake.org/cmake/help/latest/variable/CMAKE_SYSROOT.html).

However, the only way I got it to work was by disabling the complete CMAKE_MODULE_PATH in the root CMakeLists.txt file.

vadz:
Sorry, I don't really understand neither the original problem nor the solution.

Do you mean that something references the files created during SOCI build in its build directory after it was deleted? If so, isn't the solution just to install SOCI in some directory which doesn't get deleted? I could be completely missing the point, of course, but please explain what actually happens if I do.

vadz:
I know of yocto but I've never really used it.

Anyhow, thanks for the explanation, but I still don't understand why do the paths for the dependent libraries become invalid. These libraries should be out side of the temporary workspace normally, shouldn't they?

It's probably a good idea to use `CMAKE_DL_LIBS` (although I'd appreciate any confirmation from CMake experts), but how does the standard `FindSQLite3` module differ from the one included in SOCI and what exactly results in the problem occurring with the latter but not the former?

AHeimberger:
It is just a guessing right now. In the past there was an option called `SOCI_LIBDIR`. This option has been removed in commit https://github.com/SOCI/soci/commit/d8632f5aca9042bc5788b462b7fdee14b97ae5be but is still referenced in the old [.bb-recipe](http://cgit.openembedded.org/meta-openembedded/tree/meta-oe/recipes-dbs/soci/soci_4.0.3.bb?h=master). Momentary I am not sure if they solved it this way. 

The difference between the two find_package implementations [soci](https://github.com/SOCI/soci/blob/master/cmake/modules/FindSoci.cmake) vs. [cmake](https://github.com/Kitware/CMake/blob/master/Modules/FindSQLite3.cmake) is I guess the Hint to the paths.

I need a bit more time, I will also try to deliver something so you can reproduce it on your own. BR

vadz:
Sorry, I'm not going to apply this in its current state, so closing.

PR #1035, opened 2023-02-23
Fix for CI ODBC build

vadz:


vadz:
The remaining build failures are bogus and due to apt-get errors (no idea why do they happen so often on GitHub Actions images compared to anywhere else), so I'm going to merge this as is.

PR #1038, opened 2023-03-12
Always initialize a variable to avoid -Wmaybe-uninitialized

### LINKED ISSUE #1037: maybe-uninitialized error
When a user builds SOCI with `-Werror=maybe-uninitialized` enabled and the compiler chooses to inline the below soci code a hard compiler errors results. This occurs even with the soci includes marked as `SYSTEM` in CMake as inlining negates this protection.

I believe changing `base_type baseValue;` to `base_type baseValue{};` at `include/soci/values.h:176` would fix the issue

```bash
In file included from _deps/soci-src/src/core/../../include/soci/column-info.h:13,
                 from _deps/soci-src/src/core/../../include/soci/soci.h:16,
In constructor ‘soci::details::copy_holder<T>::copy_holder(const T&) [with T = int]’,
    inlined from ‘void soci::values::set(const std::string&, const T&, soci::indicator) [with T = boost::optional<int>]’ at _deps/soci-src/src/core/../../include/soci/values.h:180:21:
_deps/soci-src/src/core/../../include/soci/values.h:37:32: error: ‘baseValue’ may be used uninitialized [-Werror=maybe-uninitialized]
   37 |     copy_holder(T const & v) : value_(v) {}
      |                                ^~~~~~~~~
_deps/soci-src/src/core/../../include/soci/values.h: In member function ‘void soci::values::set(const std::string&, const T&, soci::indicator) [with T = boost::optional<int>]’:
_deps/soci-src/src/core/../../include/soci/values.h:176:23: note: ‘baseValue’ was declared here
  176 |             base_type baseValue;
```

vadz:
This warning is sometimes given (e.g. when using optional<int>) and can't be easily suppressed from the application code, so ensure the variable is always initialized to avoid it.

Closes #1037.

### DIFF
--- a/include/soci/values.h
+++ b/include/soci/values.h
@@ -173,7 +173,7 @@ public:
             indicator * pind = new indicator(indic);
             indicators_.push_back(pind);
 
-            base_type baseValue;
+            base_type baseValue{};
             type_conversion<T>::to_base(value, baseValue, *pind);
 
             details::copy_holder<base_type> * pcopy =


andrewkcorcoran (approved):

vadz:
I've somehow completely forgotten to merge this, but it's done now.

Issue #179, opened 2013-07-25
Enable uses of ODBC API with Unicode support


mloskot:
This is second part of improvements initially reported in #163 where small set of necessary fixes have been applied and will be released with SOCI 3.2.2.

All further fixes for the uses of ODBC API with Unicode should be attached here.
CMake updates will be necessary too, see #164.

Issue #164, opened 2013-06-24
Update CMake to Unicode projects for VS


mloskot:
Currently, Visual Studio projects are generated with `Character Set` property set to `Use Multi-Byte Character Set`. So, SOCI is built in non-Unicode mode with VS, whereas default mode in fairly recent VS versions is Unicode.

We should align the configuration with the VS default settings.
This is also related to users wishing to use ODBC in full Unicode mode, however, support for Windows UTF-16/UCS-2 through `std::wstring` has not been implemented, yet.

Krzmbrzl:
I believe this has nothing to do with CMake, does it (i.e. the tag seems wrong)? 🤔

vadz:
As MSVS projects are generated by CMake, I think the tag is correct. Whether it's really worth doing this is another question...

Krzmbrzl:
I have never seen CMake do anything with regards to character encoding in this way - but I guess I also never tried to explicitly generate Visual Studio Projects with specific properties set 🤔:shrug:

PR #1041, opened 2023-04-04
UTF-16 support for ODBC (for MSSQL)

bold84:
This pull request adds support for UTF-16 encoding in the ODBC module. The module previously only supported UTF-8 encoding, which made it difficult to exchange data with non-UTF-8 MSSQL databases properly.

With this new feature, the ODBC module can now properly exchange data with MSSQL databases that use UTF-16 encoding. This is especially important for users who work with databases that use different encodings or require internationalization support.

The implementation uses std::wstring_convert to convert between UTF-8 and UTF-16 encodings. The toUtf16() and toUtf8() functions have been added to handle the conversion between the two encodings.

There seem to have been issues in the past:

https://github.com/SOCI/soci/issues/164
https://github.com/SOCI/soci/issues/179
https://github.com/SOCI/soci/issues/1111


tests/odbc/test-odbc-mssql.cpp
```diff
@@ -75,6 +75,34 @@ TEST_CASE("MS SQL long string", "[odbc][mssql][long]")
...
+            : table_creator_base(sql)
+        {
+            sql << "create table soci_test ("
+                    "str nvarchar(255) null"
+                ")";
+        }
+    } unicode_table_creator(sql);
+
+    // std::string str_in { "สวัสดีชาวโลก!" };
+    std::string str_in { u8"\u0E2A\u0E27\u0E31\u0E2A\u0E14\u0E35\u0E0A\u0E32\u0E27\u0E42\u0E25\u0E01!" };
```
vadz, line 94: I don't think we want to use `u8` here, this makes the literal `char8_t` in C++20 and so this won't compile (because `std::string` expects `char`).

tests/common-tests.h
```diff
@@ -6106,8 +6115,11 @@ TEST_CASE_METHOD(common_tests, "CLOB", "[core][clob]")
     sql << "select s from soci_test where id = 1", into(s2);
 
     CHECK(s2.value.size() == 0);
-
+#ifdef SOCI_ODBC_WIDE
+    s1.value = make_long_xml_string(3000);
+#else
```
vadz, line 6120: Could we handle this inside `make_long_xml_string()`, e.g. divide the length by 2 there in the wide char case?

src/backends/odbc/vector-use-type.cpp
```diff
@@ -181,19 +181,39 @@ void* odbc_vector_use_type_backend::prepare_for_bind(SQLUINTEGER &size,
...
+            sqlType = SQL_WCHAR;
+            cType = SQL_C_WCHAR;
+#else
             char *pos = buf_;
 
             for (std::size_t i = 0; i != vsize; ++i)
             {
                 *pos++ = (*vp)[i];
-                *pos++ = 0;
+                *pos++ = 0;
```
vadz, line 210:
```suggestion
                *pos++ = 0;
```

src/backends/odbc/vector-use-type.cpp
```diff
@@ -181,19 +181,39 @@ void* odbc_vector_use_type_backend::prepare_for_bind(SQLUINTEGER &size,
 
             prepare_indicators(vsize);
 
-            size = sizeof(char) * 2;
+#ifdef SOCI_ODBC_WIDE
+            size = sizeof(SQLWCHAR) * 2; // 1 wchar + 1 null terminator
+#else
+            size = sizeof(char) * 2; // 1 char + 1 null terminator
+#endif // SOCI_ODBC_WIDE
+
```
vadz, line 189: Can't it be written just as

```suggestion
            size = sizeof(SQLTCHAR) * 2; // 1 wchar + 1 null terminator
```

?

src/backends/odbc/vector-use-type.cpp
```diff
@@ -204,31 +224,60 @@ void* odbc_vector_use_type_backend::prepare_for_bind(SQLUINTEGER &size,
...
-            for (std::size_t i = 0; i != vecSize; ++i)
+            for (std::size_t i = 0; i < vecSize; ++i)
             {
                 std::size_t sz = vector_string_value(type_, data_, i).length();
+
+#ifdef SOCI_ODBC_WIDE
+                set_sqllen_from_vector_at(i, static_cast<long>(sz * sizeof(SQLWCHAR)));
+#else
                 set_sqllen_from_vector_at(i, static_cast<long>(sz));
+#endif // SOCI_ODBC_WIDE
```
vadz, line 235: Again, why not just

```suggestion
                set_sqllen_from_vector_at(i, static_cast<long>(sz * sizeof(SQLTCHAR)));
```

?

vadz (commented):
Thanks for working on this, but I don't know how do I feel about introducing a separate build variant for Unicode support. IMO it would be really better to just use UTF-8 in all builds, instead of requiring a special build mode to handle Unicode.

And it would be really nice to have some description of this option in the docs, if only to explain what does enabling it change.

Finally, even if this is relatively trivial, it looks like the use of `SQLTCHAR` could avoid some preprocessor checks in the code.

src/backends/odbc/vector-use-type.cpp
bold84: At first I didn't know of the existence of SQLTCHAR and created my own alias. Because it can get confusing, I later removed that alias and did it this way. But I'm fine either way. I will make the necessary changes everywhere.

bold84:
> Thanks for working on this, but I don't know how do I feel about introducing a separate build variant for Unicode support. IMO it would be really better to just use UTF-8 in all builds, instead of requiring a special build mode to handle Unicode.
> 
> And it would be really nice to have some description of this option in the docs, if only to explain what does enabling it change.
> 
> Finally, even if this is relatively trivial, it looks like the use of `SQLTCHAR` could avoid some preprocessor checks in the code.

The problem I have is that the data has been written already with another application and I have no influence on that. The database(s) are not UTF-8. 

The changes in this PR were the only way I could figure out to prevent reading jibberish:

Screenshot 2024-03-20 at 01 25 50

Perhaps you have a better suggestion on how to solve this problem. I'd be glad to implement a more elegant solution.

bold84:
maybe connection_parameters can be abused to decide at runtime what type of char / string is stored in the database.

vadz:
> The problem I have is that the data has been written already with another application and I have no influence on that. The database(s) are not UTF-8.

I see, thanks.

> Perhaps you have a better suggestion on how to solve this problem.

We do need to add `SQLWCHAR` support to be able to work with the existing database using UTF-16, but I think it should be available in addition to UTF-8 support with plain `SQLCHAR`, with the decision about which one to use being performed at _run-time_ rather than compile-time.

Ideally, this should be automatic, i.e. when exchanging data with `SQLWCHAR` columns, UTF-16 should be read from/written to them, while the same code should use UTF-8 when working with `SQLCHAR` columns, but I don't know if it's possible to implement this easily, so perhaps we need a new `db_wstring` type instead. If we could make this work automatically, it would be really great, however.

The main point is that I'd really, really love to avoid different incompatible builds. The conditional compilation directives in the tests are a good example of how we do _not_ want the code using SOCI to look like. And there are other problems, e.g. we'd need to add wide char builds to the CI too if we do it like this and I'd rather avoid it.

bold84:
> > The problem I have is that the data has been written already with another application and I have no influence on that. The database(s) are not UTF-8.
> 
> 
> 
> I see, thanks.
> 
> 
> 
> > Perhaps you have a better suggestion on how to solve this problem.
> 
> 
> 
> We do need to add `SQLWCHAR` support to be able to work with the existing database using UTF-16, but I think it should be available in addition to UTF-8 support with plain `SQLCHAR`, with the decision about which one to use being performed at _run-time_ rather than compile-time.
> 
> 
> 
> Ideally, this should be automatic, i.e. when exchanging data with `SQLWCHAR` columns, UTF-16 should be read from/written to them, while the same code should use UTF-8 when working with `SQLCHAR` columns, but I don't know if it's possible to implement this easily, so perhaps we need a new `db_wstring` type instead. If we could make this work automatically, it would be really great, however.
> 
> 
> 
> The main point is that I'd really, really love to avoid different incompatible builds. The conditional compilation directives in the tests are a good example of how we do _not_ want the code using SOCI to look like. And there are other problems, e.g. we'd need to add wide char builds to the CI too if we do it like this and I'd rather avoid it.

Okay, I think we're on the same page. Shall there be an implicit conversion to std::string to shall std::wstring be supported in case of SQLWCHAR based columns?

Krzmbrzl:
Also note that wstring_convert has been deprecated and scheduled for removal from the standard (without replacement). It's a shame, but either way everyone here should be aware of this 👀

bold84:
> Also note that wstring_convert has been deprecated and scheduled for removal from the standard (without replacement). It's a shame, but either way everyone here should be aware of this 👀

That could temporarily (until there's a replacement) be solved with icu, libiconv or alike. But I wasn't sure if that's acceptable at this point.
On the other hand, it could be optional.

vadz:
> Shall there be an implicit conversion to std::string or shall std::wstring be supported in case of SQLWCHAR based columns?

I think it would make sense to support `wstring` as people using `SQLWCHAR` in ODBC are most likely to use it too. But ideally I'd like to be able to support `string` to/from `SQLWCHAR` mapping too using UTF-8.

> That could temporarily (until there's a replacement) be solved with icu, libiconv or alike. 

I'd rather not pull in ICU just for this and libiconv is Unix-only. If necessary, I can contribute my own code, written many years ago, converting between UTF-8 and `wchar_t` (i.e. either UTF-16 or UTF-32). It has some lower level functions as well as

```cppp
std::string ToUTF8(const std::wstring& wstr);
std::wstring FromUTF8(const std::string& uft8);
```

bold84:
That sounds good!

Okay, I'll add std::wstring support in a separate PR first.

Afterwards we can add the conversion to std::string.

PR #1042, opened 2023-04-05
Make soci_error nothrow copy constructible

zann1x:
According to the CERT C++ Coding Standard rule [ERR60-CPP. Exception objects must be nothrow copy constructible](https://www.securecoding.cert.org/confluence/display/cplusplus/ERR60-CPP.+Exception+objects+must+be+nothrow+copy+constructible), SOCI's base exception `soci_error` should have `noexcept` copy constructors.
The exception classes derived from `soci_error` don't have copy constructors, i.e. they adhere to this rule automatically.


vadz (commented):
Thanks, I agree with making the special member functions `noexcept`, but I'm not sure why is swap-based implementation of `operator=()` is better than the current one?

If we do decide to do it like this, we should inline `make_safe_copy()` into the ctor as it's now used in a single place only and it isn't really useful to have a separate function for it any more.

But personally I'd just add `noexcept` and not change anything else, as the existing code already shouldn't throw AFAICS.

zann1x:
I switched to the copy-and-swap implementation based on [OOP54-CPP. Gracefully handle self-copy assignment](https://wiki.sei.cmu.edu/confluence/display/cplusplus/OOP54-CPP.+Gracefully+handle+self-copy+assignment). Alternatively, I think that we should at least check for self-assignment in the copy constructor.

I don't have a strong opinion about it, though. So if you think we should keep the old copy logic, I'll happily revert that part.

zann1x:
Sure thing, let's do `&e == this` then.

I don't have access to my work station until monday, so you'll have to wait a bit :)

PR #1053, opened 2023-05-29
add support of bulk operations for ORM in ORACLE and SQLite backends

vyskubov:
Hello everyone,

I am excited to submit this pull request, which (re)introduces support for bulk operations in the SOCI library's ORM (Object-Relational Mapping) functionality. This work is based on the initial [groundwork](https://github.com/SOCI/soci/pull/168) laid by @abc100m, and I have built upon their contributions to enhance the capabilities of the library.

I would appreciate it if the community could review and provide feedback on the changes. Your insights and suggestions are valuable in ensuring the quality and effectiveness of this new feature.

Thank you for your attention and consideration. I look forward to collaborating with you all to further enhance the SOCI library.

Best regards,
Aleksey

### DIFF
--- a/include/soci/row.h
+++ b/include/soci/row.h
@@ -19,6 +19,12 @@
 
 namespace soci
 {
+struct CaseInsensitiveComparator {
+    bool operator()(const std::string& a, const std::string& b) const noexcept
+    {
+        return ::strcasecmp(a.c_str(), b.c_str()) < 0;
+    }
+};
 
 class SOCI_DECL column_properties
 {
@@ -40,7 +46,7 @@ private:
 class SOCI_DECL row
 {
 public:
-    row();
+    row(std::size_t bulk_size = 1);
     ~row();
 
     row(row &&other) = default;
@@ -51,13 +57,17 @@ public:
     std::size_t size() const;
     void clean_up();
 
+    //bulk buffer size
+    std::size_t bulk_size() const { return bulk_size_; }
+    void bulk_size(const std::size_t sz) { bulk_size_ = sz; }
+
     indicator get_indicator(std::size_t pos) const;
     indicator get_indicator(std::string const& name) const;
 
     template <typename T>
-    inline void add_holder(T* t, indicator* ind)
+    inline void add_holder(std::vector<T>* t, std::vector<indicator>* ind)
     {
-        holders_.push_back(new details::type_holder<T>(t));
+        holders_.push_back(new details::vector_type_holder<T>(t));
         indicators_.push_back(ind);
     }
 
@@ -68,17 +78,17 @@ public:
     T get(std::size_t pos) const
     {
         typedef typename type_conversion<T>::base_type base_type;
-        base_type const& baseVal = holders_.at(pos)->get<base_type>();
+        base_type const& baseVal = holders_.at(pos)->get<base_type>(bulk_pos_);
 
         T ret;
-        type_conversion<T>::from_base(baseVal, *indicators_.at(pos), ret);
+        type_conversion<T>::from_base(baseVal, (*indicators_.at(pos))[bulk_pos_], ret);
         return ret;
     }
 
     template <typename T>
     T get(std::size_t pos, T const &nullValue) const
     {
-        if (i_null == *indicators_.at(pos))
+        if (i_null == (*indicators_.at(pos))[bulk_pos_])
         {
             return nullValue;
         }
@@ -98,7 +108,7 @@ public:
     {
         std::size_t const pos = find_column(name);
 
-        if (i_null == *indicators_[pos])
+        if (i_null == (*indicators_[pos])[bulk_pos_])
         {
             return nullValue;
         }
@@ -122,6 +132,13 @@ public:
     void reset_get_counter() const
     {
         currentPos_ = 0;
+        bulk_pos_   = 0;
+    }
+
+    void next(std::size_t num = 1) const
+    {
+        currentPos_  = 0;
+        bulk_pos_   += num;
     }
 
 private:
@@ -130,12 +147,14 @@ private:
     std::size_t find_column(std::string const& name) const;
 
     std::vector<column_properties> columns_;
-    std::vector<details::holder*> holders_;
-    std::vector<indicator*> indicators_;
-    std::map<std::string, std::size_t> index_;
+    std::vector<details::vector_holder*> holders_;
+    std::vector<std::vector<indicator>*> indicators_;
+    std::map<std::string, std::size_t, CaseInsensitiveComparator> index_;
 
     bool uppercaseColumnNames_;
-    mutable std::size_t currentPos_;
+    mutable std::size_t currentPos_;    //column
+    mutable std::size_t bulk_pos_;      //row
+    mutable std::size_t bulk_size_;
 };
 
 } // namespace soci
--- a/include/soci/soci-platform.h
+++ b/include/soci/soci-platform.h
@@ -64,6 +64,8 @@ namespace std {
     }
 }
 #endif // _MSC_VER < 1800
+#define strncasecmp _strnicmp
+#define strcasecmp _stricmp
 #endif // _MSC_VER
 
 #if defined(__CYGWIN__) || defined(__MINGW32__)
--- a/include/soci/sqlite3/soci-sqlite3.h
+++ b/include/soci/sqlite3/soci-sqlite3.h
@@ -60,6 +60,10 @@ private:
     int result_;
 };
 
+
+typedef std::map<std::string, data_type> sqlite3_data_type_map;
+sqlite3_data_type_map get_data_type_map();
+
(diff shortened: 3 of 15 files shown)


vadz (commented):
Thanks for the PR but I have a few questions just after the first very quick look at it:

1. Could you please try to minimize the number of unrelated changes? For examples, in several places `at(pos)` is replaced with `[pos]` which is not obviously right but even if it is, it would be much better to make it in a separate PR. There are also a lot of whitespace-only changes that really should be avoided.
2. Why are the new tests different for Oracle and SQLite? Is there some actual difference between the backends functionality and, if so, what is it? And if not, could we please have these tests in the common code?
3. And, of course, there are the the CI failures, could you please look at them?

I'll try to have a more detailed look at this later, but fixing (1) would be really helpful.

TIA!

vadz:
>     1. I have reverted all the whitespace-only changes.

Thanks!

> Regarding the replacement of `at(pos)` with `[pos]`, these changes were already present in the original PR, so I simply included them as-is.

Sorry for not realizing this.

> However, I have now reverted these changes for the time being. It's worth mentioning that the usage of `at(pos)` and `[pos]` in this file is inconsistent, which could be addressed in a separate PR.

I agree that inconsistencies should be fixed but it's definitely better to do them in a separate PR. Thanks for fixing this up too!

asmwarrior:
What is the status of this PR, I think the bulk adding is a very important feature for soci.

I have a similar discussion issue discussion here: [ORM: What is the correct way to bulk add a lot of rows? · Issue #1066 · SOCI/soci](https://github.com/SOCI/soci/issues/1066), in the issue, I have to use a for loop to add many rows, but I do need a simple way to do that.

Thanks.

vyskubov:
I apologize for abandoning this PR. I would like to complete the work on it; however, I currently have more urgent tasks to attend to.

Furthermore, I do not have a readily available development environment on Windows and FreeBSD. Any assistance would be greatly appreciated.

Thank you.

tests/sqlite3/test-sqlite3.cpp
```diff
@@ -397,6 +397,32 @@ struct longlong_table_creator : table_creator_base
...
+{
+    // we need to have an table that uses autoincrement to test this.
+    session sql(backEnd, connectString);
+
+    test4_table_creator tableCreator(sql);
+
+    {
+        std::vector<std::string> vec_name { "John", "James" };
+        sql << "insert into soci_test(name) values(:name)", use(vec_name);
+        assert(vec_name.size() == 2);
```
vadz, line 410: All these `assert`s really need to be changed to `CHECK()`s.

tests/sqlite3/test-sqlite3.cpp
```diff
@@ -676,6 +702,36 @@ class test_context : public test_context_base
...
+    session sql(backEnd, connectString);
+    table_creator_three one(sql);
+
+    sql << "insert into soci_test values('john', '(404)123-4567')";
+    sql << "insert into soci_test values('doe', '(404)123-4567')";
+
+    std::vector<PhonebookEntry> vec_orm(10);
+    sql << "select * from soci_test", into(vec_orm);
+
+    assert(vec_orm.size() == 2);
```
vadz, line 718: Here too.

tests/sqlite3/test-sqlite3.cpp
```diff
@@ -676,6 +702,36 @@ class test_context : public test_context_base
...
+
+    for (std::vector<PhonebookEntry>::iterator it = vec_orm.begin();
+         it != vec_orm.end();
+         ++it)
```
vadz, line 722: Could use range-for loop here...

include/soci/row.h
```diff
@@ -40,7 +46,7 @@ class SOCI_DECL column_properties
 class SOCI_DECL row
 {
 public:
-    row();
+    row(std::size_t bulk_size = 1);
```
vadz, line 49:
```suggestion
    explicit row(std::size_t bulk_size = 1);
```

src/backends/sqlite3/vector-into-type.cpp
```diff
@@ -132,8 +169,16 @@ void sqlite3_vector_into_type_backend::post_fetch(bool gotData, indicator * ind)
     int const endRow = static_cast<int>(statement_.dataCache_.size());
     for (int i = 0; i < endRow; ++i)
     {
-        sqlite3_column &col = statement_.dataCache_[i][position_-1];
-
+        std::shared_ptr<sqlite3_column> col1;
```
vadz, line 172: Why do we use a shared pointer here? It doesn't seem to be shared with anybody...

src/backends/sqlite3/vector-into-type.cpp
```diff
@@ -132,8 +169,16 @@ void sqlite3_vector_into_type_backend::post_fetch(bool gotData, indicator * ind)
...
+        if (1 == endRow)
```
vadz, line 173: I wish there were a comment here explaining what is this test for. I can't understand what's going on here.

include/soci/sqlite3/soci-sqlite3.h
```diff
@@ -60,6 +60,10 @@ class SOCI_SQLITE3_DECL sqlite3_soci_error : public soci_error
     int result_;
 };
 
+
+typedef std::map<std::string, data_type> sqlite3_data_type_map;
+sqlite3_data_type_map get_data_type_map();
+
```
vadz, line 66: There doesn't seem to be any reason to extract this here, is there?

include/soci/soci-platform.h
```diff
@@ -64,6 +64,8 @@ namespace std {
     }
 }
 #endif // _MSC_VER < 1800
+#define strncasecmp _strnicmp
```
vadz, line 67: This seems to be unused?

include/soci/values.h
```diff
@@ -243,7 +246,7 @@ class SOCI_DECL values
     std::vector<details::standard_use_type *> uses_;
     std::map<details::use_type_base *, indicator *> unused_;
     std::vector<indicator *> indicators_;
-    std::map<std::string, std::size_t> index_;
+    std::map<std::string, std::size_t, CaseInsensitiveComparator> index_;
```
vadz, line 249: Why did this need to be changed to be case-insensitive?

src/backends/oracle/statement.cpp
```diff
@@ -298,6 +298,11 @@ void oracle_statement_backend::describe_column(int colNum, data_type &type,
             {
                 type = dt_double;
             }
+        }        
+        //create table T(t number)--> select t+0 from T; we get: scale=precision=0, size=22
+        else if (dbscale == 0 && dbprec == 0 && dbsize == 22)
+        {
+            type = dt_double;
```
vadz, line 305: This seems unrelated to the topic of this PR.

src/backends/sqlite3/statement.cpp
```diff
@@ -19,12 +19,6 @@
 #pragma warning(disable:4355)
 #endif
 
-// This is used instead of tolower() just to avoid warnings about int to char
-// casts inside MSVS std::transform() implementation.
-char toLowerCh(char c) {
-    return static_cast<char>( std::tolower(c) );
-}
-
```
vadz, line 27: Removing this and inlining it looks a completely gratuitous change, was there really some reason for doing it?

src/backends/sqlite3/statement.cpp
```diff
@@ -353,11 +350,7 @@ sqlite3_statement_backend::execute(int number)
 statement_backend::exec_fetch_result
 sqlite3_statement_backend::fetch(int number)
 {
-    if (hasVectorIntoElements_ || number == 0)
-        return load_rowset(number);
-    else
-        return load_one();
-
+    return (1 == number) ? load_one() : load_rowset(number);
```
vadz, line 353: It looks like `hasVectorIntoElements_` is not used at all now and should be removed?

include/soci/type-holder.h
```diff
@@ -12,13 +12,29 @@
...
 #include <typeinfo>
+#include <vector>
 
 namespace soci
 {
 
 namespace details
 {
 
+// this class is for pass test [commontests.h, test12()],
```
vadz, line 23: This seems completely unrelated to the subject of this PR...

vadz (changes requested):
Sorry, it doesn't seem like it can be merged in the current state, there are just too many unclear/unrelated things and commented out parts.

I also don't know if we really want to unconditionally replace scalar values with vectors, i.e. use vector of size 1 instead of simple values, it just looks like such a strange idea when we already have both scalar and vector versions.

Of course, maybe I just misunderstand things, but it's really hard to guess what is the intention/rationale here without any comments anywhere in sight.

vadz:
It doesn't look like anything is going to happen here, so I have no choice but to close it, sorry.

Issue #1058, opened 2023-07-08
Statement::bind_clean_up keeps deleted indicator entries in vector


cstiborg:
I discovered that my application became slower and slower at inserting records. After chasing wild geese in my database, I did some rudimentary statistics and my results were that the first 10000 records I could insert in 7 seconds, while the last block of 10000 records up to 2 million records took 2 minutes, and the culprit was Statement::bind_clean_up().
After some more investigation I found that entries in the in indicators_ vector were surely removed, but the entries were never deleted. I did some investigation into why this might be the case, but I didn't manage to find any reason.
Finally, I ended up making a simple patch, which clears the vector after its entries are deleted. I've tested the patch in my application and I can't produce any bad behaviour. That's obviously not a mark of quality, particularly given my poor knowledge of the inner workings of SOCI.

[soci.patch](https://github.com/SOCI/soci/files/11992573/soci.patch)

zann1x:
Looking at the method you edited in your patch, it is indeed weird to see that indicators are cleaned differently than the uses and intos above it. Could you share a minimal reproducible example for the issue?

cstiborg:
Yes - I will create one.

cstiborg:
I've created a minimal reproducible example and added it to this message:

- soci-issue-1058.cpp contains a small program that connects to a postgres database, sets up a prepared statement, and then uses it to add one million records to a database table. For every 10,000 records it prints out how far it has gotten with a timestamp.
- For your debugging pleasure I've added a couple of Dockerfiles which setup a postgresql database and creates a table and role as expected by the minimal reproducible example. (You can use them to see how I build the test environment and how I run it)

To build and run either the original or patched code in docker execute the following commands as root:
`# cp Dockerfile.[original|patched] Dockerfile
# docker build -t soci-issue-1058:latest .
# docker run -it soci-issue-1058`

[soci-issue-1058.tar.gz](https://github.com/SOCI/soci/files/12050783/soci-issue-1058.tar.gz)

zann1x:
Thanks a lot! I'll look into it.

zann1x:
After having a look at it, I currently don't see anything else that might be the cause for the degrading performance.
The indicators vector is only filled when using the values API, which I'm unfortunately not deeply familiar with either. Nonetheless, the number of (freed) elements in the vector grows indefinitely, so that definitely seems like a bug needing to be fixed. As you have already provided a solution, feel free to create a PR for that!

PR #1061, opened 2023-07-17
Patch for issue #1058

cstiborg:
This small patch assures that the indicators_ vector is cleared after its content is deleted in bind_clean_up().
If this is not done the indicators_ vector grows indefinitely as it caches the deleted and NULLed indicators.
See issue #1058 for more explanation/comments.

### DIFF
--- a/src/core/statement.cpp
+++ b/src/core/statement.cpp
@@ -173,6 +173,7 @@ void statement_impl::bind_clean_up()
         delete indicators_[i];
         indicators_[i] = NULL;
     }
+    indicators_.clear();
 
     row_ = NULL;
     alreadyDescribed_ = false;


vadz:
Thanks for the PR, I agree that it looks like the right thing to do, i.e. I don't see why should we have an ever-growing `indicators_` vector, but I don't understand how can just clearing it, without adjusting any indices for accessing it anywhere, not break something: it looks like either the indices into it were wrong before (always a possibility, of course) or they must be wrong now.

But OTOH I don't actually understand where is this `indicators_` actually used at all. It is filled in `statement_impl::bind(values & values)`, but where is it used then? I wonder if we could just remove this completely, i.e. could it be a leftover from some earlier version? Or am I missing something obvious here?

vadz:
> That's what I wondered about as well. But I _think_ that `statement_impl::indicators_` serves as kind of cleanup bucket for `values::indicators_` because the entries don't seem to be cleared anywhere else.

Ah, indeed, thanks, you're absolutely right.

> But the reason for this indirection is unclear to me as well.

I have no idea what is the life cycle of `values` supposed to be, so I guess it might be too short?

Anyhow, I'll merge this PR for now, thanks again to both of you!

vadz:
Merged in the commit above.

Issue #764, opened 2019-10-26
RC4.0.0 Bulk operations


irodushka:
Hi

I'm looking at bulk operations and see that 4.0.0 documentation says:
> Bulk operations are supported also for std::vectors of the user-provided types that have appropriate conversion routines defines.

Does it mean that it's possible to insert to a vector of structs now, like `std::vector<Person>`? 
Or this option is still unavailable in SOCI?

Regards
Yuri.

mloskot:
AFAIR this should be supported, but user-defined types need conversion routines. You will find examples in https://github.com/SOCI/soci/blob/master/tests/common-tests.h

BTW, looks like that specific case for vector<Person> is not covered in the tests. Contributions welcome.

irodushka:
If you loоk into my sample code you see that I have a conversion implemented.
Moreover - if I write `use(pv[0])` instead of `use(pv)` it compiles ok.
Maybe I need to create another conversion for vector itself?.. Or I should implement set/get methods for Person class?.. You know, it's hard to guess and a test sample would be very useful here:)

mloskot:
Sorry, I currently have no time to dig the problem myself. That's why I suggested playing with simple test, trying bulk ops with the PhonebookEntry classes in tests. Simply, try DIY

irodushka:
Tried to write additional tests.
It does not compile any way.
If `use(bulk_vector)` changed to `use(bulk_vector[0])` - compiles ok.
What am I doing wrong?:)

**_(Modded common_tests.h in attach)_**

```
TEST_CASE_METHOD(common_tests, "Bulk insert with ORM", "[core][orm]")
{
    soci::session sql(backEndFactory_, connectString_);

    sql.uppercase_column_names(true);
    auto_table_creator tableCreator(tc_.table_creator_3(sql));
    
    std::vector<PhonebookEntry> bulk_vector;
    for( int i = 0; i < 100; i++ ) {
        PhonebookEntry temp;
        std::ostringstream oss;
        oss << "bulk" << i;
        temp.name = temp.phone = oss.str();

        bulk_vector.push_back( temp );
    }

    statement insertStatement = (sql.prepare << "insert into soci_test values (:NAME, :PHONE)", use(bulk_vector));
    insertStatement.execute(true);

    int count = 0;

    sql << "select count(*) from soci_test where NAME like ('bulk%')", into(count);

    CHECK(count == 100);
}
```
    SECTION("Bulk insert from vector of boost::fusion::vector")
    {
[...]

PR #1065, opened 2023-07-27
Added broken unit test for bulk user-defined objects

AndyStieber:
Broken unit test per the discussion here: https://github.com/SOCI/soci/issues/764


vadz:
Sorry, but this example is just misguided, there is no way this could work as you somehow expect SOCI to map multiple values (phone book entries have 2 fields) to a single `value`. This just can't be expected to happen.

PR #1078, opened 2023-09-14
Add support for Network address types for PostgreSQL

Liewyec:
Hello, 
I noticed a missing support for Network address types for PostgreSQL. I added the support and they are returned s strings.

### DIFF
--- a/src/backends/postgresql/statement.cpp
+++ b/src/backends/postgresql/statement.cpp
@@ -766,6 +766,8 @@ void postgresql_statement_backend::describe_column(int colNum, data_type & type,
     case 2950: // uuid
     case 829:  // macaddr
     case 869:  // inet
+    case 650:  // cidr
+    case 774:  // macaddr8
         type = dt_string;
         break;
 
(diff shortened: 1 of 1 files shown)


vadz:
Sorry, I've somehow missed this. I see no problem with applying this, but I wonder if we should add support for the other related types too, see

https://github.com/postgres/postgres/blob/b6a77c6a6ccf698787201b001cbbbf9c89fe5715/src/include/catalog/pg_type.dat#L256-L265

?

Liewyec:
Hello, sorry for the late answer, I was a bit busy and sick. I will look into this and add it, it looks easy and unless I am mistaken there are just 2 more types to add. I will update the pull request probably over weekend.

vadz:
@Liewyec do you still plan to do it? TIA!

Liewyec:
Hello, yes I am planing to do this, I  should have a bit less work in next few weeks. Sorry it takes so long a I had a lot of other work.

Liewyec:
Hello, I added the two more types, sorry it was taking so long.

vadz:
Thanks, will merge soon.

PR #1080, opened 2023-09-29
timegm implementation has been added to build on SunOs and AIX

avpalienko:
timegm function is missing on some platforms, e.g. Sun and AIX. We have to check the existence and use self-made version if not

### DIFF
--- a/include/private/soci-mktime.h
+++ b/include/private/soci-mktime.h
@@ -21,6 +21,24 @@ namespace soci
 namespace details
 {
 
+namespace inner 
+{
+
+time_t selfmade_timegm ( struct tm* tb );
+
+template <typename T>
+auto timegm_impl(T* t) -> decltype(timegm(t))
+{
+    return timegm(t);    
+}
+
+template <typename T>
+auto timegm_impl(T t) -> time_t
+{
+    return selfmade_timegm(t);
+}
+} // namespace inner
+
 // Fill the provided struct tm with the values corresponding to the given date
 // in UTC.
 //
@@ -40,7 +58,7 @@ mktime_from_ymdhms(tm& t,
     t.tm_min  = minute;
     t.tm_sec  = second;
 
-    timegm(&t);
+    inner::timegm_impl(&t);
 }
 
 // Helper function for parsing datetime values.
--- a/include/private/soci-mktime.h
+++ b/include/private/soci-mktime.h
@@ -21,10 +21,7 @@ namespace soci
 namespace details
 {
 
-namespace inner 
-{
-
-time_t selfmade_timegm ( struct tm* tb );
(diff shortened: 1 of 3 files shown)


include/private/soci-mktime.h
```diff
@@ -21,6 +21,24 @@ namespace soci
 namespace details
 {
 
+namespace inner
```
vadz, line 24: Do we really need this namespace? `details` is already a kind of "inner" namespace.

vadz (commented):
Thanks! Is `selfmade_timegm()` your own code? A quick search finds some other versions of this function which seem a bit more clear and don't use macros, e.g. [this one](https://stackoverflow.com/a/58037981/15275).

Especially if it's your own code, it would be really nice to have at least some trivial unit tests for this function to make sure it works correctly now (which it probably does) and keeps working in the future (as it would be easy to break it with the automated tests only running under platforms not using this function at all).

include/private/soci-mktime.h
avpalienko: fixed

include/private/soci-mktime.h
```diff
@@ -21,6 +21,24 @@ namespace soci
 namespace details
 {
 
+namespace inner 
+{
+
+time_t selfmade_timegm ( struct tm* tb );
```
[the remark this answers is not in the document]
avpalienko: fixed

avpalienko:
> Thanks! Is `selfmade_timegm()` your own code? A quick search finds some other versions of this function which seem a bit more clear and don't use macros, e.g. [this one](https://stackoverflow.com/a/58037981/15275).
> 
> Especially if it's your own code, it would be really nice to have at least some trivial unit tests for this function to make sure it works correctly now (which it probably does) and keeps working in the future (as it would be easy to break it with the automated tests only running under platforms not using this function at all).

Thank you for review! 

No.  It's a copy from internet. Your example does not normalize the argument but timegm does.
I've added the unit test and removed the macros

vadz:
>  It's a copy from internet.

It would be nice to point out its provenance, if only to have a confirmation that it's available under a compatible licence.

vadz:
Ugh, sorry, a quick code search finds this code in `mktime.c` file with

```c
/*
  Copyright (c) 1987-1997, Microsoft Corporation. All rights reserved.
*/
```

in it so we really can't include it into SOCI, unfortunately.

Could you please get the version from SO (which is implicitly in public domain AFAIK) or find some other implementation under acceptable licence?

avpalienko:
Ups! :-(
Thank you! 
I've changed implementation to as SO

vadz:
Merged now, thanks for the update!

Begasus:
Thanks for this, was hitting this when checking another project that uses SOCI, fixes the build for Haiku. +1

Issue #1081, opened 2023-10-05
Backend inconsistencies


cstiborg:
I recently got a change request for my application to include support for MySQL. Using SOCI I assumed that would be straight-forward. However, I ran into the issue with the line:

`soci::rowset<soci::row> rowset(sql);`

It constructs the object when using a PostgreSQL backend but when using MySQL it throws a soci::mysql_soci_error stating that the query is empty.

I looked into how it could be that it’s:
- Allowed in PostgreSQL but not MySQL? Simply put, PostgreSQL is happy with an empty query and simply returns an empty set for your foolishness, whereas MySQL wants to educate you on the error of your ways and throws an exception.
- Not found by a test case? All test cases for rowsets in MySQL are on the form: `soci::rowset<soci::row> rowset = (sql.prepare << …);` and no test for an empty rowset is performed. However, in the PostgreSQL tests this is not the case, and constructing an empty rowset is tested.

After realizing this I pondered what should be the right way to fix my problem. Whether I should handle it in my code or if I’m right in expecting that SOCI should behave identically across backends, and that throwing on a constructor, when it's provided what I consider to be valid input, is a bit weird.

My conclusion obviously (otherwise I wouldn’t be writing this) was that I would expect SOCI to behave consistently across backends, though I’m aware that meddling too much with the signalling from the backends may cause unexpected behaviour.

So I set out to produce a minimal patch. I'll attach this work to this issue and also make it available in a branch on my SOCI repo-clone. Please note that I have only patched the MySQL backend code, not the test. Along with the patch I’ve also produced a minimal piece of code to reproduce the problem and a couple of Dockerfiles, both build a test setup, one tests the original code and another tests the patched code.

The code to reproduce can be executed as follows:

`# cp Dockerfile.[original|patched] Dockerfile`
`# docker build -t original:latest .`
`# docker run -it original`

If consensus is made that this approach is sensical I will create a pull request.

[soci-issue-1081.tar.gz](https://github.com/SOCI/soci/files/12820109/soci-issue-1081.tar.gz)

vadz:
Sorry, but could you please show just the diff? Also, it's not really clear to me what happens when you construct a rowset from `sql` object in MySQL case, could you please explain it if you already know it?

All I can say without looking at the details is that I would indeed expect the API to behave the same with all the backends, but I would _not_ expect SOCI from isolating you between the differences in their SQL dialects. And it's unfortunately not totally clear to me if this falls in the former or the latter case.

[...]

PR #1082, opened 2023-10-09
Disallowing constructing rowset from a session

cstiborg:
After discussion on #1081 it was decided to explicitly disallow the implicit constructor `rowset(soci::session const & session)`.


vadz:
Thanks, I've cherry picked the second commit (please don't include merges in the branches from which you make the PRs, you should branch them directly from SOCI master) with just some comments added.

Issue #1057, opened 2023-07-07
How to create an empty rowset?


vatnic:
Good afternoon!
I did not find documentation on how to create an empty rowset, which is necessary for output from the method, when an exception occurs.
```cpp
rowset<row> DbConnector::selectData(wxWindow* win, wxString selectString)
{
    rowset<row> rowSet= (sql.prepare << "");
    try
    {
        rowSet= (sql.prepare << selectString);
        return rowSet;
    }
    catch (const std::exception& e)
    {
        wxMessageDialog messageDialog(win, e.what(), "Ошибка", wxOK);
        messageDialog.ShowModal();
    }
    return rowSet;
}
```
How to create an empty rowset without accessing the database?

vadz:
I'm afraid a `rowset` already has to be valid currently. It wouldn't be difficult to add a default ctor to it and change `begin()` and `end()` to return the (same) invalid iterator for such an empty `rowset`. Any patches doing this would be welcome!

asmwarrior:
I got here by searching. I just want to know how to got the `size` of the `rowset`. Maybe this issue is related, I see there is no function call named `size()`, so I have to iterate all the elements of the `rowset` to get the size of this `rowset`?
`

asmwarrior:
[Add rowset::clear() · Issue #198 · SOCI/soci](https://github.com/SOCI/soci/issues/198)

This is a related discussion, but it is in year 2013, and no response of this issue.



Also, in this issue [ORM and rowset · Issue #1020 · SOCI/soci](https://github.com/SOCI/soci/issues/1020), some one gives a way to translate the `rowset `to `std::vector`.

cstiborg:
I've created a PR which patches `rowset` with a default constructor. Feel free to comment.

asmwarrior:
Is it possible to make a new tag(release) in the git? I just looked at the git commit history, the last git tag is 4.0.3, which is in beginning of the year 2022. Thanks.

Issue #198, opened 2013-09-11
Add rowset::clear() 


mloskot:
The `rowset<T>` is conceptually similar to a standard container like `std::vector`, so may be sensible to be able to clear it with the following post-conditions:
- attached statement resource is cleaned and deallocated
- `rowset<T>::begin() == rowset<T>::end()`

PR #1084, opened 2023-10-11
Soci 1057

cstiborg:
This PR allows the creation of an empty rowset as per #1057.

### DIFF
--- a/include/soci/rowset.h
+++ b/include/soci/rowset.h
@@ -39,8 +39,8 @@ public:
         : st_(0), define_(0)
     {}
 
-    rowset_iterator(statement & st, T & define)
-        : st_(&st), define_(&define)
+    rowset_iterator(statement * st, T * define)
+        : st_(st), define_(define)
     {
         // Fetch first row to properly initialize iterator
         ++(*this);
@@ -113,6 +113,11 @@ public:
 
     typedef rowset_iterator<T> iterator;
 
+    rowset_impl()
+        : refs_(1), st_(nullptr), define_(nullptr)
+    {
+    }
+
     rowset_impl(details::prepare_temp_type const & prep)
         : refs_(1), st_(new statement(prep)), define_(new T())
     {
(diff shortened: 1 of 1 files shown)


include/soci/rowset.h
```diff
@@ -136,7 +141,7 @@ class rowset_impl
     iterator begin() const
     {
         // No ownership transfer occurs here
-        return iterator(*st_, *define_);
+        return iterator(st_.get(), define_.get());
```
vadz, line 144: I don't understand how does this not result in a crash... `rowset_iterator` ctor calls `++` which uses the pointer, which is null here for an invalid rowset.

Am I missing something here or would it actually crash if you added a test exercising the new code?

vadz (changes requested):
Sorry, I don't think this is correct. I'd also rather avoid changing the existing `rowset_iterator` ctor, we already have a ctor for creating an invalid iterator, so we should just use it instead.

Also, could you please rebase your branch on SOCI master to avoid having unrelated changes in this PR?

TIA!

include/soci/rowset.h
cstiborg: I have been testing this and it does work. However, I also wonder why the `++(*this)` doesn't make it explode... I can do a gdb session on it and see where it takes me.

Talking about the design itself, I agree that changing the `rowset_iterator` ctor is not ideal, however, I was taken away by the beauty of the minimal change to `begin()`. I did another version in which I didn't change the `rowset_iterator` instead I had:

`iterator begin() const`
`{`
`    // No ownership transfer occurs here`
`    return (st_.get() == nullptr)`
`        ? iterator()`
`        : iterator(*st_, *define_);`
`}`

vadz: Well, beauty is in the eye of the beholder and personally I find a ctor taking optional and maybe null pointers pretty ugly. And

```cpp
iterator begin() const
{
    // Empty rowset doesn't have any valid begin iterator.
    return st_ ? iterator(*st_, *define_) : iterator();
}
```

seems perfectly fine to me.

cstiborg: It's coming up - I just need to test it - while I'm at it, should I add a clear for #198 and close a 10 year old issue? ;)

PR #1086, opened 2023-10-11
Creating empty constructor for rowset

cstiborg:
This PR allows the creation of an empty `rowset` as per #1057

### DIFF
--- a/include/soci/rowset.h
+++ b/include/soci/rowset.h
@@ -113,6 +113,11 @@ public:
 
     typedef rowset_iterator<T> iterator;
 
+    rowset_impl()
+        : refs_(1), st_(nullptr), define_(nullptr)
+    {
+    }
+
     rowset_impl(details::prepare_temp_type const & prep)
         : refs_(1), st_(new statement(prep)), define_(new T())
     {
@@ -135,8 +140,8 @@ public:
 
     iterator begin() const
     {
-        // No ownership transfer occurs here
-        return iterator(*st_, *define_);
+        // No ownership transfer occurs here. Empty rowset doesn't have any valid begin iterator.
+        return st_ ? iterator(*st_, *define_) : iterator();
     }
 
     iterator end() const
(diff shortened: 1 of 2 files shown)


cstiborg:
I will add a test.

Regarding the "Previous problem" - I did catch a problem. I have tested both the previous code and this version using the test code below which might not be short and sweet, but at least to me it proves that it works. In both cases it compiled and executed as expected - I have it in a docker container, if you want to see it for yourself.

	#include <soci/soci.h>
	#include <string>
	#include <iostream>

	class Test {
	public:
		Test(soci::session& session): session(session) {
		}

	 	void buildIt() {
			session << "delete from soci_test";
			for (int i = 0; i < 5; i++) {
				session << "insert into soci_test (a) values (" << i << ")";
			}
			rowset = (session.prepare << "select * from soci_test");
			readRow = rowset.begin();
		}
 
		bool handleRow() {
			if (readRow != rowset.end()) {
				std::cout << "Got row" << std::endl;
				++readRow;
			}
			return readRow != rowset.end();
		}

	private:
		soci::session& session;
		soci::rowset<soci::row> rowset;
		soci::rowset<soci::row>::const_iterator readRow;
	};

	int
	main(int argc, char *argv[]) {
		soci::session postgresql("postgresql", "user=soci host=127.0.0.1 dbname=postgres password=abcd1234");
		soci::session mysql("mysql", "user=soci host=127.0.0.1 dbname=soci password=abcd1234");

		Test pgTest(postgresql);
		Test myTest(mysql);
		if (pgTest.handleRow()) std::cout << "PG Row holds data" << std::endl;
		if (myTest.handleRow()) std::cout << "MY Row holds data" << std::endl;
		pgTest.buildIt();
		myTest.buildIt();
		while (pgTest.handleRow()) {}
		while (myTest.handleRow()) {}

		return 0;
	}

cstiborg:
Is it possible to re-run the tests? - Something went wrong when fetching from a Ubuntu repo...

vadz:
Thanks, should I merge this or will you add `clear()` to this PR too?

cstiborg:
I've just tested and added it - now waiting for the automated tests to succeed.

cstiborg:
I forgot about the test, sorry - I've added it now.

vadz:
Thanks again for you work! I'll merge it in a moment.

PR #1089, opened 2023-10-12
Implementation of the datetime (with up to nanoseconds) has been added

avpalienko:


vadz:
The build failures must be fixed, of course, but beyond this could I ask you to please _avoid_ mixing up whitespace-only changes and variable renaming with the significant changes? If you really want to, you can make them in separate commits (or maybe separate PRs later), but having them all here makes it much harder to review the changes than necessary, so could you please undo all the cosmetic changes? TIA!

avpalienko:
> The build failures must be fixed, of course, but beyond this could I ask you to please _avoid_ mixing up whitespace-only changes and variable renaming with the significant changes? If you really want to, you can make them in separate commits (or maybe separate PRs later), but having them all here makes it much harder to review the changes than necessary, so could you please undo all the cosmetic changes? TIA!

Sorry, this was not on purpose - these are IDE settings on my project. I've undone it

vadz:
Thanks for the updates, but there still seems to be a big number of unrelated changes (e.g. all MSVC-specific stuff) as well as many changes to the tests which I wouldn't expect (I'd only expect additions, not changes to the existing tests) and the PR is just huge -- it's going to be very hard to find time to review it, so any further efforts to reduce its size would be welcome.

One global question is why exactly do we need the third-party date header? Can't we live with just the functionality that standard chrono provides?

avpalienko:
> Thanks for the updates, but there still seems to be a big number of unrelated changes (e.g. all MSVC-specific stuff) as well as many changes to the tests which I wouldn't expect (I'd only expect additions, not changes to the existing tests) and the PR is just huge -- it's going to be very hard to find time to review it, so any further efforts to reduce its size would be welcome.
> 

I've created new PR as diff -w. I hope this reduces the number of unrelated changes. 

> One global question is why exactly do we need the third-party date header? Can't we live with just the functionality that standard chrono provides?

Third party header is https://howardhinnant.github.io/date/date.html.  I can't do it myself using chrono. I've read this date.h is a part of of C++20

PR #1093, opened 2023-10-16
Implementation of the datetime (with up to nanoseconds) has been added

avpalienko:


vadz (changes requested):
Sorry, but I couldn't finish reviewing the changes, partially because there are just too many of them clearly unrelated to the topic of this PR and I got discouraged when I saw that I covered barely half. It would be really great if you could please leave just the datetime-related parts.

I also still think that we should handle datetime as strings for the backends that don't support them natively, as otherwise this new type doesn't seem very useful.

We also need at least some documentation for the new type, especially if it's only supported in Oracle and PostgreSQL backends (which is not obvious).

vadz:
Sorry, I can't really deal with this on my own and after almost a year of inactivity I have no choice but to close this one. Please redo it with the minimal changes just for the datetime support if you can.

TIA!

PR #1097, opened 2023-10-22
Add automatic widening conversions in dynamic bindings

Sildra:
Add the ability to perform lossless conversions in the dynamic bindings.
Also add an opt-in narrowing cast with the define `SOCI_WIDENING_CASTS` (to be discussed).

Allow for extending the range of casts with the templated struct `soci::soci_cast` with the model of `soci::type_conversion<T>`.

### DIFF
--- a/include/soci/metaprog.h
+++ b/include/soci/metaprog.h
@@ -1,0 +1,77 @@
+//
+// Copyright (C) 2004-2008 Maciej Sobczak, Stephen Hutton
+// Distributed under the Boost Software License, Version 1.0.
+// (See accompanying file LICENSE_1_0.txt or copy at
+// http://www.boost.org/LICENSE_1_0.txt)
+//
+
+#ifndef SOCI_METAPROG_H_INCLUDED
+#define SOCI_METAPROG_H_INCLUDED
+#include <cstdint>
+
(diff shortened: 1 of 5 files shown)


Sildra:
`int64_t` and `long long` are the same under Windows and different under linux. I will either need to wait for #954 or temporarily switch to old types for the build to pass.

Sildra:
I will need to add support for a c++98 version of the metaprog templates and check if adding `constexpr` to the `get` function avoids the weird unreachable code on the statements. I can't seems to reproduce both of those issues under VS2022.

vadz:
You can base this PR on #954 if it makes things simpler.

As for the unreachable code warnings (do you have a direct link?), they can be suppressed explicitly if nothing else works.

Sildra:
Only the Windows builds are missing : for the 2 latest, I will silence the compiler warning, for the 2 oldest, it might be because they don't have `std::numeric_limits<long long>`. Defining a generic alias for `soci::numeric_limit<T>` + providing specialized `soci::numeric_limits<long long>` might do the trick.

I will also add tests in common to check the feature.

Sildra:
I will check the changes proposed by the new types.

PR #1102, opened 2023-11-23
Allow passing rvalues to soci::use()

avpalienko:
It is a useful when we can write soci::use( 3 ) or soci::use (string_holder("once used string"))
So, for rvalue case we have to use a specific type of use_container. And the overload of the exchange() and do_use() functions. As a result we have a "use_type" object of new derived class "by_value_use_type" (or "by_value_conversion_use_type"), which contains the copy of the rvalue

### DIFF
--- a/include/soci/type-conversion.h
+++ b/include/soci/type-conversion.h
@@ -183,6 +183,21 @@ private:
     SOCI_NOT_COPYABLE(conversion_use_type)
 };
 
+template <typename T>
+class by_value_conversion_use_type : value_holder<T>, public conversion_use_type<T>
+{
+public:
+    by_value_conversion_use_type ( T const& t, std::string const& name = std::string () )
+        : value_holder<T> ( t, i_ok ), conversion_use_type<T> ( value_holder<T>::saved_val_, name )
+    {
+    }
+
+    by_value_conversion_use_type ( T const& t, const indicator& ind, std::string const& name = std::string () )
+        : value_holder<T> ( t, ind ), conversion_use_type<T> ( value_holder<T>::saved_val_, value_holder<T>::saved_ind_, name )
+    {
+    }
+};
+
 // this class is used to ensure correct order of construction
 // of vector based into_type and use_type elements that use type_conversion
 
@@ -438,6 +453,18 @@ into_type_ptr do_into(std::vector<T> & t, std::vector<indicator> & ind,
 {
     return into_type_ptr(
         new conversion_into_type<std::vector<T> >(t, ind, begin, end));
+}
+
+template <typename T>
+use_type_ptr do_use ( T& t, std::string const& name, user_type_tag, std::true_type )
+{
+    return use_type_ptr ( new by_value_conversion_use_type<T> ( t, name ) );
+}
+
+template <typename T>
+use_type_ptr do_use ( T& t, indicator& ind, std::string const& name, user_type_tag, std::true_type )
+{
+    return use_type_ptr ( new by_value_conversion_use_type<T> ( t, ind, name ) );
 }
 
 template <typename T>
--- a/include/soci/use.h
+++ b/include/soci/use.h
@@ -18,7 +18,7 @@ namespace soci
 
 namespace details
 {
-template <typename T, typename Indicator>
+template <typename T, typename Indicator, typename ByValueTag = std::false_type>
 struct use_container
 {
     use_container(T &_t, Indicator &_ind, const std::string &_name)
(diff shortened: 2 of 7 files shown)


avpalienko:
CI env problem: 
Starting MySQL 5.7
Start-Service : Cannot find any service with service name 'MySQL57'.
At line:1 char:1
+ Start-Service 'MySQL57'
+ ~~~~~~~~~~~~~~~~~~~~~~~
    + CategoryInfo          : ObjectNotFound: (MySQL[57](https://ci.appveyor.com/project/SOCI/soci/builds/48597466/job/r6lwn67f6ep8jnvb#L57):String) [Start-Service], ServiceCommandException
    + FullyQualifiedErrorId : NoServiceFoundForGivenName,Microsoft.PowerShell.Commands.StartServiceCommand

tests/oracle/test-oracle.cpp
```diff
@@ -180,39 +180,68 @@ struct basic_table_creator : public table_creator_base
     }
 };
 
-TEST_CASE("Oracle nested statement", "[oracle][blob]")
+TEST_CASE("Oracle nested statement", "[oracle][cursors]")
```
vadz, line 183: Is the change to this test case really related to the main topic of the PR?

tests/oracle/test-oracle.cpp
```diff
@@ -386,6 +415,69 @@ TEST_CASE("Oracle null user-defined objects in/out", "[oracle][null][type_conver
     CHECK(ind == i_null);
 }
 
+TEST_CASE ( "Oracle", "[oracle][rvalue]" )
```
vadz, line 418: This test shouldn't Oracle-specific, should it? I.e. using rvalues should work, and be tested, with all backends AFAICS, and so this test should be in `common-tests.h`.

vadz (commented):
Sorry, I don't have time to really review the new code, but it seems clear that at least some comments explaining what it does would be helpful. Could you please try to add them?

TIA!

vadz:
> CI env problem: Starting MySQL 5.7 Start-Service : Cannot find any service with service name 'MySQL57'. At line:1 char:1
> 
>     * Start-Service 'MySQL57'
> 
>     * ```
>         + CategoryInfo          : ObjectNotFound: (MySQL[57](https://ci.appveyor.com/project/SOCI/soci/builds/48597466/job/r6lwn67f6ep8jnvb#L57):String) [Start-Service], ServiceCommandException
>         + FullyQualifiedErrorId : NoServiceFoundForGivenName,Microsoft.PowerShell.Commands.StartServiceCommand
>       ```

AppVeyor probably updated their MySQL version... Also not sure why did we suddenly start getting deprecation warnings in the GitHub CI builds. This is clearly unrelated to this PR, of course.

tests/oracle/test-oracle.cpp
avpalienko: No, this addition is an "incidental correction" )))

tests/oracle/test-oracle.cpp
avpalienko: I've used dual table and implicit conversion int->string. I think it's oracle-specific. But I try add a test like this into common-tests.h

avpalienko: done

include/soci/use.h
```diff
@@ -48,10 +48,16 @@ struct use_container<T, no_indicator>
 
 // soci::use is deleted for rvalues because it will likely lead to subtle stack-use-after-scope bugs.
```
vadz, line 49: This comment is not true any more and really should be updated, it's just too confusing otherwise.

include/soci/type-conversion.h
```diff
@@ -183,6 +183,21 @@ class conversion_use_type
     SOCI_NOT_COPYABLE(conversion_use_type)
 };
 
+template <typename T>
+class by_value_conversion_use_type : value_holder<T>, public conversion_use_type<T>
```
vadz, line 187: It would be really great to have at least some explanation about what this class is used for.

vadz (commented):
I'm sorry but I'm still unable to understand how does this code work. Maybe it seems obvious to you even without any comments, but this is not at all the case for me. Please consider adding some comments (and updating existing ones which are out of date now) to help me -- and the other people reading this code -- understand it.

TIA!

vadz:
I've finally taken time to read this carefully and I understand how it works now, but I'm not sure I like it: for the objects that are expensive to copy, we now implicitly make another copy of them, which might be quite unexpected and have noticeable performance implications.

IMO we could use it for cheap to copy types such as fundamental numeric types and maybe provide some way to opt-in into this by specializing some trait. But I don't think this should be done for strings etc.

So I'm closing this PR because it does it unconditionally, but I'd welcome any PRs doing it for just numbers because this could indeed be useful to have and doesn't have any drawbacks. I also wonder if we could implement this in a simpler way by just specializing `use_container` for such types to store `T` instead of `T&`, it looks like it could be much easier to do and understand than what this PR does, but perhaps there are some problems I fail to see with this approach.

PR #1104, opened 2023-12-07
Added handling of ssl_mode connection parameter to MySQL session

willw-git:
This PR allows SOCI users to pass the ssl-mode parameter described [here](https://dev.mysql.com/doc/refman/8.0/en/connection-options.html#option_general_ssl-mode) when connecting to a MySQL database. The setting is passed as `ssl_mode`, with its value set to one of the name constants given in the above documentation: `DISABLED, PREFERRED, REQUIRED, VERIFY_CA` and `VERIFY_IDENTITY`. Used like this:
    
    auto connection_str = "host=localhost db=mydb user=fred pass='12345' ssl_mode=DISABLED";
    session sql(mysql, connection_str);

This patch implies that the documentation at https://soci.sourceforge.net/doc/master/backends/mysql/ needs to be fixed to include this parameter and its possible values.

**Motivation:** I found that a 32-bit client could not connect to a server, even though the server permitted unencrypted connections. The error returned was 2026. I believe this is because the default value of this parameter is PREFERRED, which implies that TLS protocol negotiations are initiated. Because the 32-bit client is no longer maintained, these fail - and no connection is made. The solution is to turn off SSL entirely (if the security situation permits it).

Of course, the parameter is also generally useful.

### DIFF
--- a/src/backends/mysql/session.cpp
+++ b/src/backends/mysql/session.cpp
@@ -200,7 +200,8 @@ void parse_connect_string(const string & connectString,
     string *charset, bool *charset_p, bool *reconnect_p,
     unsigned int *connect_timeout, bool *connect_timeout_p,
     unsigned int *read_timeout, bool *read_timeout_p,
-    unsigned int *write_timeout, bool *write_timeout_p)
+    unsigned int *write_timeout, bool *write_timeout_p,
+    unsigned int *ssl_mode, bool *ssl_mode_p)
 {
     *host_p = false;
     *user_p = false;
@@ -217,6 +218,7 @@ void parse_connect_string(const string & connectString,
     *connect_timeout_p = false;
     *read_timeout_p = false;
     *write_timeout_p = false;
+    *ssl_mode_p = false;
     string err = "Malformed connection string.";
     string::const_iterator i = connectString.begin(),
         end = connectString.end();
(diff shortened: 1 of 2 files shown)


src/backends/mysql/session.cpp
```diff
@@ -335,6 +337,15 @@ void parse_connect_string(const string & connectString,
...
             *write_timeout = std::strtoul(val.c_str(), &endp, 10);
             *write_timeout_p = true;
+        } else if (par == "ssl_mode" && !*ssl_mode_p)
+        {
+            if (val=="DISABLED") *ssl_mode = SSL_MODE_DISABLED;
+            else if (val=="PREFERRED") *ssl_mode = SSL_MODE_PREFERRED;
+            else if (val=="REQUIRED") *ssl_mode = SSL_MODE_REQUIRED;
+            else if (val=="VERIFY_CA") *ssl_mode = SSL_MODE_VERIFY_CA;
+            else if (val=="VERIFY_IDENTITY") *ssl_mode = SSL_MODE_VERIFY_IDENTITY;
+            else throw soci_error(err);
```
vadz, line 347: I think it would be nice to say that "ssl mode is invalid" here instead of giving just the generic error message.

vadz (commented):
Thanks, I'd be glad to add this, but the documentation indeed needs to be updated, could you please add the new option description to `docs/backend/mysql.md`?

vadz:
Oh, and you need to look into whether we can use `mysql_options()` instead of the deprecated function resulting in CI errors. If we can't, the warning would need to be disabled.

src/backends/mysql/session.cpp
willw-git: Hi,

As you suggested I've fixed the `soci_error()` call to provide a relevant string, and I have modified the docs to describe the addition.

The `mysql_ssl_set()` call I don't understand - it occurs outside the bit I have touched, at line 420 (my version). I used `mysql_options()` (line 459) to do my work. Can't see how I goofed...

PR #1119, opened 2024-01-09
Only prepare for describe if statement type is 'SELECT' for Oracle backend

iqbal-hasprime:
In my use-case I always call statement::exchange_for_rowset, even if the query for the statement does not return any result, since type of query used to create the statement is unknown beforehand.

OCIStmtExecute call with OCI_DESCRIBE_ONLY in oracle_statement_backend::prepare_for_describe will fail (with "ORA-24333: zero iteration count" error) for non-SELECT queries, since there's nothing to describe.

My proposed change to the library is to the call to OCIStmtExecute if the statement type is OCI_STMT_SELECT.

### DIFF
--- a/src/backends/oracle/statement.cpp
+++ b/src/backends/oracle/statement.cpp
@@ -164,21 +164,34 @@ std::string oracle_statement_backend::rewrite_for_procedure_call(
 
 int oracle_statement_backend::prepare_for_describe()
 {
-    sword res = OCIStmtExecute(session_.svchp_, stmtp_, session_.errhp_,
-        1, 0, 0, 0, OCI_DESCRIBE_ONLY);
+    ub2 statementType = 0;
+    sword res = OCIAttrGet(static_cast<dvoid*>(stmtp_),
+        static_cast<ub4>(OCI_HTYPE_STMT), static_cast<dvoid*>(&statementType),
+        0, static_cast<ub4>(OCI_ATTR_STMT_TYPE), session_.errhp_);
     if (res != OCI_SUCCESS)
     {
         throw_oracle_soci_error(res, session_.errhp_);
     }
 
-    int cols;
-    res = OCIAttrGet(static_cast<dvoid*>(stmtp_),
-        static_cast<ub4>(OCI_HTYPE_STMT), static_cast<dvoid*>(&cols),
-        0, static_cast<ub4>(OCI_ATTR_PARAM_COUNT), session_.errhp_);
-
-    if (res != OCI_SUCCESS)
+    int cols = 0;
+    if (statementType == OCI_STMT_SELECT)
     {
-        throw_oracle_soci_error(res, session_.errhp_);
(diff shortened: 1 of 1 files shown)


src/backends/oracle/statement.cpp
```diff
@@ -164,21 +164,29 @@ std::string oracle_statement_backend::rewrite_for_procedure_call(
...
 int oracle_statement_backend::prepare_for_describe()
 {
-    sword res = OCIStmtExecute(session_.svchp_, stmtp_, session_.errhp_,
-        1, 0, 0, 0, OCI_DESCRIBE_ONLY);
-    if (res != OCI_SUCCESS)
-    {
-        throw_oracle_soci_error(res, session_.errhp_);
-    }
+    ub2 statementType = 0;
+    sword res = OCIAttrGet(static_cast<dvoid*>(stmtp_),
```
vadz, line 168: Shouldn't we do something with `res`? I am not actually sure what should happen if an error happens here, but it looks like we ought to handle it somehow.

src/backends/oracle/statement.cpp
```diff
@@ -164,21 +164,29 @@ std::string oracle_statement_backend::rewrite_for_procedure_call(
...
+        static_cast<ub4>(OCI_HTYPE_STMT), static_cast<dvoid*>(&statementType),
+        0, static_cast<ub4>(OCI_ATTR_STMT_TYPE), session_.errhp_);
+
+    int cols = 0;
+    if (statementType == OCI_STMT_SELECT) {
```
vadz, line 173: Minor style nitpick:
```suggestion
    if (statementType == OCI_STMT_SELECT)
    {
```

vadz (commented):
Thanks, I agree that it's harmless to avoid this error, but I'd just like to check what should happen in case of an error?

src/backends/oracle/statement.cpp
iqbal-hasprime: Yes I agree, we should throw throw_oracle_soci_error here if an error occurs.

vadz (approved):
Looks good, thanks!

I'll merge this soon.

vadz (changes requested):
Oops, sorry, I spoke too soon and didn't notice that a test failed.

It looks like this change breaks some other uses of SOCI, so it can't be applied as is... Sorry for missing it originally.

iqbal-hasprime:
> Oops, sorry, I spoke too soon and didn't notice that a test failed.
> 
> It looks like this change breaks some other uses of SOCI, so it can't be applied as is... Sorry for missing it originally.

No worries. I should have checked as well. I'll try and fix the failing tests.

vadz:
OK, I've finally looked at this and the failures are to mistakenly reading the number of columns into the wrong `cols` variable, defined inside the block, instead of the right one defined outside it. I've fixed this and will push soon.

PR #1120, opened 2024-01-17
SQLite update DDL creation to the new db enum

Sildra:


vadz:
Sorry, could you please explain what does this actually change?

tests/common-tests.h
```diff
@@ -172,6 +172,66 @@ std::ostream& operator<<(std::ostream& ostr, const std::vector<MyOptionalString>
     return ostr;
 }
 
+// The next classes are used for DDL testing
```
vadz, line 175: This comment is really not very useful :-( I'd appreciate if it could describe at least a bit what's going on here.

tests/common-tests.h
```diff
@@ -172,6 +172,66 @@ std::ostream& operator<<(std::ostream& ostr, const std::vector<MyOptionalString>
...
+template<typename T>
+struct DdlTester {
+    typedef T val_type;
+    DdlTester(soci::db_type type, T val, bool expectNull)
```
vadz, line 179: Do I understand correctly that `val` is not used if `expectNull` is true? I.e. both these parameters could be replaced with `std::optional<T>`?

tests/common-tests.h
```diff
@@ -172,6 +172,66 @@ std::ostream& operator<<(std::ostream& ostr, const std::vector<MyOptionalString>
...
+        case soci::db_int8:   t.outVal = static_cast<typename T::val_type>(v.get<std::int8_t>(pos));   break;
+        case soci::db_uint8:  t.outVal = static_cast<typename T::val_type>(v.get<std::uint8_t>(pos));  break;
+        case soci::db_int16:  t.outVal = static_cast<typename T::val_type>(v.get<std::int16_t>(pos));  break;
+        case soci::db_uint16: t.outVal = static_cast<typename T::val_type>(v.get<std::uint16_t>(pos)); break;
+        case soci::db_int32:  t.outVal = static_cast<typename T::val_type>(v.get<std::int32_t>(pos));  break;
+        case soci::db_uint32: t.outVal = static_cast<typename T::val_type>(v.get<std::uint32_t>(pos)); break;
+        case soci::db_int64:  t.outVal = static_cast<typename T::val_type>(v.get<std::int64_t>(pos));  break;
+        case soci::db_uint64: t.outVal = static_cast<typename T::val_type>(v.get<std::uint64_t>(pos)); break;
+        case soci::db_double: t.outVal = static_cast<typename T::val_type>(v.get<double>(pos));        break;
+        default: break;
```
vadz, line 231: A comment explaining why do we ignore the other types would be really welcome.

tests/common-tests.h
```diff
@@ -288,8 +348,43 @@ template<> struct type_conversion<PhonebookEntry3>
     }
 };
 
+template<typename T>
+struct type_conversion<DdlTester<T>>
+{
+    typedef soci::values base_type;
+    static void from_base(soci::values const &v, soci::indicator, DdlTester<T>& t)
+    {
+        map_ddl_tester(t, v);
```
vadz, line 357: Is there any reason not to just inline this function here, if it's only called from here?

If nothing else, this would make it more clear what it does.

tests/common-tests.h
```diff
@@ -288,8 +348,43 @@ template<> struct type_conversion<PhonebookEntry3>
...
+    }
+    static void to_base(DdlTester<T> const &t, soci::values &v, soci::indicator&)
+    {
+        v.set("VAL", t.inVal);
+    }
+};
 } // namespace soci
 
+template<typename T>
+void test_limit(soci::session &sql, DdlTester<T>&& tester)
```
vadz, line 367: The name of this function is confusing, why does it say "limit" when there are no limits in play here? Something like `test_roundtrip()` would be more clear.

tests/common-tests.h
```diff
@@ -172,6 +172,66 @@ std::ostream& operator<<(std::ostream& ostr, const std::vector<MyOptionalString>
...
+template<>
+void check(DdlTester<double> const &val)
+{
+    CHECK(val.inType == val.outType);
+    CHECK(val.expectNull == val.isNull);
+    if (val.expectNull)
+        return;
+    CHECK(std::fpclassify(val.inVal) == std::fpclassify(val.outVal));
+    if (std::isnormal(val.inVal) && std::isnormal(val.outVal))
+        CHECK_THAT(val.inVal, Catch::Matchers::WithinRel(val.outVal));
```
vadz, line 209: I suggest extracting this and the `CHECK(val.inVal == val.outVal);` line from above into a separate `DdlTester::check_roundtrip()` function and then removing this `check()` because the first 4 lines are identical for all types and should be done directly in `test_limit()`.

tests/common-tests.h
```diff
@@ -288,8 +348,43 @@ template<> struct type_conversion<PhonebookEntry3>
...
+        soci::statement stmt = (sql.prepare << "SELECT * FROM " << table);
+        stmt.exchange(soci::into(tester));
+        stmt.define_and_bind();
+        stmt.execute();
+        stmt.fetch();
+        check(tester);
+    }
+    catch (const std::exception& e)
+    {
+        CHECK_FALSE(e.what());
```
vadz, line 383:
```suggestion
        FAIL(e.what());
```

tests/sqlite3/test-sqlite3.cpp
```diff
@@ -641,6 +641,21 @@ TEST_CASE("SQLite DDL with metadata", "[sqlite][ddl]")
     CHECK(ddl_t3_found == false);
 }
 
+TEST_CASE("SQLite DDL symmetry", "[sqlite][ddl][symmetry]")
```
vadz, line 644: "symmetry" should really be replaced with "round trip"

tests/sqlite3/test-sqlite3.cpp
```diff
@@ -641,6 +641,21 @@ TEST_CASE("SQLite DDL with metadata", "[sqlite][ddl]")
...
+{
+    soci::session sql(backEnd, connectString);
+    test_limit(sql, DdlTester<double>       (soci::db_double, std::numeric_limits<double>::max(),        false));
+    test_limit(sql, DdlTester<double>       (soci::db_double, std::numeric_limits<double>::quiet_NaN(),  true));
```
vadz, line 648: So this is the _only_ case when we pass `true` as `expectNull`? I'd just remove it and simplify things by removing all the related checks, it seems like a huge overkill to add all this code just to check that NaN is getting returned as null (and I'm not even sure if it's the desired behaviour!).

include/soci/sqlite3/soci-sqlite3.h
```diff
@@ -345,15 +345,23 @@ struct sqlite3_session_backend : details::session_backend
             case db_double:
                 return "real";
             case db_date:
+                return "date";
```
vadz, line 348: This actually changes the affinity of the created column: it used to be `INTEGER` but is now `NUMERIC`. It's not really clear whether this is the right thing to do and if we do decide to do it, it needs to be documented as an incompatible change.

But first of all, could you please explain why do we want/need to do this?

include/soci/sqlite3/soci-sqlite3.h
```diff
@@ -345,15 +345,23 @@ struct sqlite3_session_backend : details::session_backend
...
             case db_int8:
+                return "int1";
```
vadz, line 350: Why do we need these weird names (I've never seen `int1` used anywhere, have you?) instead of just using `int8`? AFAICS SQLite doesn't care anyhow, as long as the type contains `int`.

Or, even better, why not use the names already recognized by `get_data_type_map()`, i.e. `tinyint` here etc?

vadz (commented):
Sorry but the test code is very difficult to understand as written: it has no comments and uses unclear or even misleading names. Could you please simplify it to make it actually understandable?

As for the changes themselves, I'm not sure at all about the `date` type, but otherwise I _think_ they should do no harm. I'd appreciate opinion of the other SQLite users however.

tests/common-tests.h
Sildra: val is still pushed to the database before getting the result.
This is to test the handling of the value NaN which is interprested by sqlite as null. Others databases does not handle NaN the same way.

tests/sqlite3/test-sqlite3.cpp
Sildra: This is a SQLite specific behavior

tests/common-tests.h
vadz: I think it would make more sense to add a separate test checking for NaN to null conversion if we really want it (as I said, I'm still not sure if we actually do) rather than reusing this -- apparently unrelated -- test for this, at the cost of a lot of extra complications.

include/soci/sqlite3/soci-sqlite3.h
Sildra: Will change to the previous names

tests/sqlite3/test-sqlite3.cpp
Sildra: Will replace everything with Roundtrip

include/soci/sqlite3/soci-sqlite3.h
Sildra: date is mapped to db_date but db_date is mapped to integer.
At this point, I am pretty sure that the function `create_column` described in https://soci.sourceforge.net/doc/master/utilities/ was only tested with int.

vadz: Sorry, I don't understand: this is not about mapping but about the actual SQL sent to SQLite. AFAICS after this change it will use `alter table $table add column $column date` while previously it used `add column integer`. Am I wrong about it?

tests/common-tests.h
Sildra: Actually, `FAIL` is forcing a failure for the test case while `CHECK_FALSE` will still allow the test to continue, in this case, it makes more sense to keep `CHECK_FALSE`.

vadz: Sorry, should have been `FAIL_CHECK()`.

But using `FAIL` is important, it makes _sense_, unlike `CHECK_FALSE(e.what())` which does not -- we're not checking for `e.what() == nullptr` at all here.

include/soci/sqlite3/soci-sqlite3.h
Sildra: Let's handle this issue separately for now.

tests/common-tests.h
Sildra: If we want to add non-arithmetic types, we will need explicit templates specialization (otherwise static_casts will not compile) and those cannot be declared static inline.

vadz: I'd prefer to keep things simple for now and refactor later if necessary rather than making them more complicated than needed already.

And if you really want to keep this function, please find a better name for it and add a comment explaining what it does and why it can't be inlined here.

TIA!

tests/common-tests.h
Sildra: II was looking at the macro in catch2/assertions.md instead of catch2/logging.md that's why I had trouble finding the macro.

tests/common-tests.h
Sildra: templates are implicitely inlined and free functions templates requires to be explicitely not static:
`warning C4499: 'static': an explicit specialization cannot have a storage class (ignored)`

vadz: Sorry, not sure what you're doing, but are you trying to specialize `from_base()`?

What I meant was not to specialize anything at all, just use the primary template.

Sildra: Updated

Sildra:
Hello, any news about improving the table creation with the DDL interface ?

Regards,

tests/common-tests.h
```diff
@@ -287,9 +287,89 @@ template<> struct type_conversion<PhonebookEntry3>
...
+void check(Roundtrip<double> const &val)
+{
+    CHECK(val.inType == val.outType);
+    CHECK(std::fpclassify(val.inVal) == std::fpclassify(val.outVal));
+    if (std::isnormal(val.inVal) && std::isnormal(val.outVal))
+        CHECK_THAT(val.inVal, Catch::Matchers::WithinRel(val.outVal));
+}
+
+template<typename T>
+void test_roundtrip(soci::session &sql, Roundtrip<T>&& tester)
```
vadz, line 352: Why is `tester` passed as r-value reference, it doesn't seem to be consumed by this function? Or is it? A comment would be nice...

FWIW I'd probably just pass `db_type` (unless it can be deduced from `T`?) and the value directly here.

tests/common-tests.h
```diff
@@ -287,9 +287,89 @@ template<> struct type_conversion<PhonebookEntry3>
         ind = i_ok;
     }
 };
+} // namespace soci
+
+template<typename T>
```
vadz, line 292: A comment explaining what is this type used for would be nice to have.

vadz (commented):
Thanks, I think the changes are good, with the exception of not recognizing the previously recognized `int2` and `int8`, but it still looks like the testing code could be simplified — or, if not, at least commented more...

src/backends/sqlite3/statement.cpp
```diff
@@ -470,18 +470,22 @@ static sqlite3_data_type_map get_data_type_map()
...
     m["boolean"]            = db_int32;
     m["int"]                = db_int32;
     m["integer"]            = db_int32;
-    m["int2"]               = db_int32;
     m["mediumint"]          = db_int32;
 
+    m["unsignedint"]        = db_uint32;
+
     m["bigint"]             = db_int64;
-    m["int8"]               = db_int64;
```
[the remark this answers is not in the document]
Sildra: I was wondering about int2 and int8 as they are not proper SQL types. If you want to keep them I would suggest to use int2/4/8 as 16/32/64 bits for consistency.

tests/common-tests.h
Sildra: db_type could be deduced from T but it would require defining templates for this and it would add a lot of code.

src/backends/sqlite3/statement.cpp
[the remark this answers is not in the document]
Sildra: Technically, the deduction of the type would have been delegated to sqlite so the code still worked.

include/soci/sqlite3/soci-sqlite3.h
Sildra: btw, this is mentioned in #969.

vadz:
Thanks, merged now with minor stylistic corrections.

Issue #1088, opened 2023-10-11
SOCI doesn't support automatic type conversion


Sildra:
Currently, SOCI doesn't support any kind of automatic conversion between integrals.

This issue forces developpers to perform database-specific conversions notably when using sqlite and another db simultaneously.

Providing a feature performing this automatic conversion between int and long long will enable sqlite to change it's default type from 32bit to 64bit and use dates without breaking existing current user code.

```c++
// Required in the user code to perform type safe conversions when using multiple backends
// Note: the name lookup is performed twice
template<typename T>
T get_safe_integer(const soci::values& v, const std::string& name) {
    switch (v.get_properties(name).get_data_type()) {
        case soci::dt_integer: return (T)v.get<int>(name);
        case soci::dt_unsigned_long_long: return (T)v.get<unsigned long long>(name);
        case soci::dt_long_long: return (T)v.get<long long>(name);
        default: return T();
    }
}
```

I kind of attempted to add this feature by performing a dynamic cast + virtual call and it is working as intended.

```text
std::string to std::string: OK
std::string to int: Got a bad cast
int to int: OK
int to float: Got a bad cast
double to float: OK
float to double: Got a bad cast
int to long long: Got a bad cast
long long to int: OK
```

```c++
//
// Copyright (C) 2004-2008 Maciej Sobczak, Stephen Hutton
// Distributed under the Boost Software License, Version 1.0.
// (See accompanying file LICENSE_1_0.txt or copy at
// http://www.boost.org/LICENSE_1_0.txt)
//

#ifndef SOCI_TYPE_HOLDER_H_INCLUDED
#define SOCI_TYPE_HOLDER_H_INCLUDED

// std
#include <typeinfo>
#include <string>
#include <memory>
#include <iostream>


// Base class holder + derived class type_holder for storing type data
// instances in a container of holder objects
template <typename T>
class type_holder;

template <typename T>
class type_getter
{
public:
    virtual T value(T*) const = 0;
};


class holder
{
public:
    holder() {}
    virtual ~holder() {}

    template<typename T>
    T get()
    {
        if (auto* p = dynamic_cast<type_getter<T>*>(this))
            return p->value((T*)nullptr);
        throw std::bad_cast();
    }
};

template <typename T>
class type_holder : public holder, public type_getter<T>
{
public:
    type_holder(T * t) : t_(t) {}
    ~type_holder() override { delete t_; }

    T value(T*) const override { return *t_; }

private:
    T * t_;
};

[...]

PR #1127, opened 2024-02-27
Add automatic type conversion to row::get()

zann1x:
As discussed in https://github.com/SOCI/soci/issues/1088, it is possible that the merge of https://github.com/SOCI/soci/pull/1116 might require users to use a different data type when calling `row::get<T>(...)`.

This PR tries to add implicit conversions between data types in the most conservative way. I.e., conversions are only defined between integer types.
Unit tests that previously had to be changed to conform to the stricter type mapping are also reverted back to their old state.

The work here is based on https://github.com/SOCI/soci/pull/1097.

### DIFF
--- a/include/soci/type-holder.h
+++ b/include/soci/type-holder.h
@@ -8,9 +8,14 @@
 #ifndef SOCI_TYPE_HOLDER_H_INCLUDED
 #define SOCI_TYPE_HOLDER_H_INCLUDED
 
-#include "soci/soci-platform.h"
-// std
-#include <cstring>
+#include "soci/soci-backend.h"
+#include "soci/soci-types.h"
+
+#include <cmath>
+#include <cstdint>
+#include <ctime>
+#include <limits>
+#include <type_traits>
 #include <typeinfo>
 
 namespace soci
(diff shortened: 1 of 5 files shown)


zann1x:
I honestly don't understand why the MSVC build is failing with `warning C4702: unreachable code` because AFAICS these parts are actually used in the tests. Should I just disable C4702 for this file?

include/soci/soci-platform.h
```diff
@@ -37,6 +37,9 @@
 //base class must have dll interface
 #pragma warning(disable:4251 4275)
 
+// As long as we don't require C++17, we must disable the warning
+// "conditional expression is constant"
```
vadz, line 41: Could you please add the context here, i.e. where would this warning be given?

tests/common-tests.h
```diff
@@ -3317,6 +3317,122 @@ TEST_CASE_METHOD(common_tests, "Dynamic binding with type conversions", "[core][
...
+        sql << "insert into soci_test(id, d, str, tm)"
+            << " values(10, 20.0, 'foobar',"
+            << tc_.to_date_time("2005-12-19 22:14:17")
+            << ")";
+
+        {
+            row r;
+            sql << "select id from soci_test", into(r);
+
+            static_assert(std::numeric_limits<int32_t>::digits10 == 9, "");
```
vadz, line 3338: Sorry, could you please add a comment (or maybe message in the assert?) explaining why is this needed, because it doesn't seem obvious to me (i.e. I don't understand it)?

include/soci/type-holder.h
```diff
@@ -44,49 +50,310 @@ T* checked_ptr_cast(U* ptr)
     return static_cast<T*>(ptr);
 }
 
-// Base class holder + derived class type_holder for storing type data
-// instances in a container of holder objects
-template <typename T>
-class type_holder;
-
-class holder
+// Type safe conversion that fails at compilation if instantiated
```
vadz, line 53: Is this really correct? I don't see why would it fail at compile-time, it seems to be failing at run-time instead.

include/soci/type-holder.h
```diff
@@ -44,49 +50,310 @@ T* checked_ptr_cast(U* ptr)
...
 };
 
 template <typename T>
-class type_holder : public holder
+struct type_holder_trait
+{
+    static_assert(std::is_same<T, void>::value, "Unmatched raw type");
+    // dummy value to satisfy the template engine, never used
+    static const db_type type = (db_type)0;
+};
```
vadz, line 122: Do we actually have to define it? I.e. could we just declare it without defining it?

include/soci/type-holder.h
```diff
@@ -44,49 +50,310 @@ T* checked_ptr_cast(U* ptr)
...
+    static const db_type type = db_double;
+};
+
+template <>
+struct type_holder_trait<std::tm>
+{
+    static const db_type type = db_date;
+};
+
+// Base class for storing type data instances in a container of holder objects
```
vadz, line 222: Is it actually "base"? It doesn't look like we inherit from it any longer.

vadz (commented):
Thanks, the functionality provided by this PR definitely looks very useful.

It's a bit of a pity to replace a type-safe `type_holder` template with a union, I'm not sure to understand why did it have to be done, I guess it's because template `get()` can't be virtual? But, in any case, if it has to be done like this, so be it.

vadz:
> I honestly don't understand why the MSVC build is failing with `warning C4702: unreachable code` because AFAICS these parts are actually used in the tests. Should I just disable C4702 for this file?

If possible (i.e. if it works), I'd disable it for this function with `#pragma` push/pop, as usual.

include/soci/type-holder.h
```diff
@@ -44,49 +50,310 @@ T* checked_ptr_cast(U* ptr)
...
+        case db_date:
+            val_.t = static_cast<std::tm*>(val);
+            break;
+        case db_blob:
+        case db_xml:
+        case db_string:
+            val_.s = static_cast<std::string*>(val);
+            break;
+        default:
+            break;
```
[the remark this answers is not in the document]
zann1x: I don't see why this should throw as the constructor is private and it is only called during preparation for statement execution.

tests/common-tests.h
zann1x: Whoops, that's a leftover that I've missed to remove after testing.

include/soci/type-holder.h
zann1x: That's indeed not true. It's a leftover from the PR I've based this work on.

include/soci/soci-platform.h
zann1x: TBH I don't understand this error either. But as it didn't seem to be local to `type-holder.h`, I suppressed it here.
The failing build was this one: https://ci.appveyor.com/project/SOCI/soci/builds/49285927/job/fikvgmaiqq11a9np

zann1x:
> It's a bit of a pity to replace a type-safe type_holder template with a union, I'm not sure to understand why did it have to be done, I guess it's because template get() can't be virtual? But, in any case, if it has to be done like this, so be it.

In an eariler attempt to fixing this issue, I've had problems getting it to work with the old class hierarchy setup that isn't explicit about its `db_type`. I think it was a problem of having a templated `get()` which I wasn't able to combine with the templated `type_holder::value()`. As #1097 followed that same route, I honestly didn't think about an alternative way.

include/soci/soci-platform.h
vadz: I guess the line refers to the `val > static_cast<U>(t_max)` test? If so, I can understand why the compiler would warn about it, as it could be always false for some types, but I'm not sure how to avoid it. Using pragma push/pop around this test might be better than suppressing the warning globally, however, it's a potentially useful one.

include/soci/type-holder.h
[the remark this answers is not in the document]
vadz: This should be unreachable, right? I.e. it might be better to use `std::unreachable()` here but as long as we can't do this, throwing seems better than doing nothing.

vadz:
> I've had problems getting it to work with the old class hierarchy setup that isn't explicit about its db_type. I think it was a problem of having a templated get() which I wasn't able to combine with the templated type_holder::value()

I think it should be possible to do it by having template function in the base (non-template) class using virtual functions overridden in the derived template class, but I really don't have time to try it, so I'm just leaving it here as an idea, but will merge it as is if you decide not to follow it.

BTW, forgot to say: @Sildra if you have any comments about this, please let us know if you have any comments about this.

zann1x:
> I think it should be possible to do it by having template function in the base (non-template) class using virtual functions overridden in the derived template class, but I really don't have time to try it, so I'm just leaving it here as an idea, but will merge it as is if you decide not to follow it.

I'm sorry but I really don't know/understand how to achieve that, so I would just leave it as is.

include/soci/soci-platform.h
zann1x: Ah, I see. I'll move the pragma push/pop to this section.

vadz:
Sorry, it looks like this got conflicts due to the changes of #992, i.e. it needs to be adjusted to support movable types, such as blobs. Could you please check if this can be fixed? TIA!

zann1x:
I rebased the branch onto main and added a commit to support movable types.
The quickest solution I came up with was a separate `get()` method that returns a reference to the underlying value if and only if the types are matching. That was the easiest way to ensure that moving from that value is okay.

Unforunately, I don't have the time to create a more sophisticated solution at the moment. If there is more to do, it would be great if someone is willing to pick it up from here.

vadz:
Thanks again! I'll merge this soon.

vadz:
Squash-merged now in the commit above.

PR #1134, opened 2024-03-27
Add back support for PostgreSQL 9.2

vadz:
This was broken by the recent blob-related changes (see #992) but is still needed.

cc @Krzmbrzl

### DIFF
--- a/src/backends/postgresql/blob.cpp
+++ b/src/backends/postgresql/blob.cpp
@@ -8,6 +8,7 @@
 #define SOCI_POSTGRESQL_SOURCE
 #include "soci/postgresql/soci-postgresql.h"
 #include <libpq/libpq-fs.h> // libpq
+#include <pg_config.h>      // for PG_VERSION_NUM
 #include <cctype>
 #include <cstdio>
 #include <cstring>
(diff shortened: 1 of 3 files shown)


src/backends/postgresql/blob.cpp
```diff
@@ -22,6 +23,20 @@
 using namespace soci;
 using namespace soci::details;
 
+// We need this helper function when using ancient PostgreSQL versions without
+// support for 64-bit offets.
+#if PG_VERSION_NUM < 90300
+static int pg_check_fits_32_bits(std::size_t len)
+{
+    if (len > 0x7fffffff)
```
Krzmbrzl, line 31: This would be a tad cleaner by using `std::numeric_limits<std::uint32_t>::max()` instead of `0x7fffffff`

Krzmbrzl (commented):
I have no issue with supporting older Postgres versions - I only removed support because I personally think it makes no sense to support unsupported Postgres versions and it simplified the implementation not to :shrug:

For reference: The oldest still supported  version is currently 12 (source: https://www.postgresql.org/support/versioning/)

src/backends/postgresql/blob.cpp
vadz: I admit I was lazy, but we'd need to include `<cstdint>` then and then remember to remove it when we finally drop support for Postgres 9.2, so I thought it would be simpler to just do it like this.

vadz:
> I have no issue with supporting older Postgres versions - I only removed support because I personally think it makes no sense to support unsupported Postgres versions and it simplified the implementation not to 🤷

We have customers still using CentOS 7 which uses 9.2, so we have to do it for at least a bit longer.

Issue #1135, opened 2024-03-27
Dates before year 1900 stored in Oracle as one hundred years earlier.


JaroslawCendrowski:
Observed on table with TIMESTAMP column type.
tm_year component of std::tm structure is negative then. The solution is to use Euclidean modulo instead of % operator when converting from std::tm to Oracle type..

vadz:
Sorry, not on a Oracle expert by any means, so I don't really know what's going on here. If you can propose a PR fixing this, please do.

PR #1136, opened 2024-03-27
Fix handling dates earlier than 1900 in Oracle backend

JaroslawCendrowski:
Solves https://github.com/SOCI/soci/issues/1135

### DIFF
--- a/tests/oracle/test-oracle.cpp
+++ b/tests/oracle/test-oracle.cpp
@@ -112,6 +112,7 @@ TEST_CASE("Oracle datetime", "[oracle][datetime]")
             CHECK(t1.tm_mday == t2.tm_mday);
             CHECK(t1.tm_mon == t2.tm_mon);
             CHECK(t1.tm_year == t2.tm_year);    
+            CHECK(1900 + t1.tm_year == i);
         }
     }    
 }
--- a/tests/oracle/test-oracle.cpp
+++ b/tests/oracle/test-oracle.cpp
@@ -85,6 +85,35 @@ TEST_CASE("Oracle datetime", "[oracle][datetime]")
 
         CHECK(t_out == std::string(buf));
     }
+
+    {
+        // date and time - between years 1 - 2201
+        for(int i = 1; i <= 2201; i = i + 50)
+        {
+            char t[10];
+            sprintf(t, "%04d", i);
+
+            std::string date = std::to_string(i) + "-03-28 14:06:13";
+            std::tm t1 {}, t2 {};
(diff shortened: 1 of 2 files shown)


vadz:
I think this solves the problem for the years between 1800 and 1900, but not those earlier than 1800, right? This is probably better than nothing, but it looks like we could easily solve it for any negative `tm_year` in the same way.

But beyond this I also wonder if this actually works in the other backends. It would be great to add a unit test checking for this to the "std::tm" section of the corresponding test in `tests/common-tests.h`.

JaroslawCendrowski:
It fixes years in starting from 1. Didn't test for years BC yet.

> I think this solves the problem for the years between 1800 and 1900, but not those earlier than 1800, right? This is probably better than nothing, but it looks like we could easily solve it for any negative `tm_year` in the same way.
> 
> But beyond this I also wonder if this actually works in the other backends. It would be great to add a unit test checking for this to the "std::tm" section of the corresponding test in `tests/common-tests.h`.

JaroslawCendrowski:
Test added. It fails with current implementation, works with my changes in standard-use-type.cpp.

> I think this solves the problem for the years between 1800 and 1900, but not those earlier than 1800, right? This is probably better than nothing, but it looks like we could easily solve it for any negative `tm_year` in the same way.
> 
> But beyond this I also wonder if this actually works in the other backends. It would be great to add a unit test checking for this to the "std::tm" section of the corresponding test in `tests/common-tests.h`.

tests/oracle/test-oracle.cpp
```diff
@@ -85,6 +93,51 @@ TEST_CASE("Oracle datetime", "[oracle][datetime]")
...
+        table_creator_for_timestamp tableCreator(sql);
+
+        for(int i = 1; i <= 2201; i = i + 50)
+        {
+            char t[10];
+            sprintf(t, "%04d", i);
+
+            std::string date = std::to_string(i) + "-03-28 14:06:13";
+            std::tm t1 {}, t2 {}, t4 {};
+            strptime(date.c_str(), "%Y-%m-%d %H:%M:%S", &t2);
```
vadz, line 110: I don't think we can use `strptime()` under Windows, unfortunately.

JaroslawCendrowski: Changed to std::get_time, unfortunately it doesn't work well with years < 100, so I changed the tested years range also.
Tested on Windows and Ubuntu.

vadz:
Thanks, I'm (squash) merging this now!

P.S. Oracle "excess-100 notation" is really weird, I wonder who came up with something like this and why.

Issue #1141, opened 2024-04-15
BLOB does not work in Oracle 


avpalienko:
In the last version we have a problem: 

        sql << "insert into soci_test (id, b) values(5, '0')";
        sql.commit ();

        soci::blob      output_blob ( sql );
        soci::indicator ind;

        sql << "select b from soci_test where id = 5", soci::into ( output_blob, ind );
        CHECK ( ind == soci::i_ok );
        CHECK ( output_blob.get_len () == 1 );

ORA-22292: Cannot open a LOB in read-write mode without a transaction
   for the parameter number 1 while executing "select b from soci_test where id
  = 5".

vadz:
AFAIU (@Krzmbrzl please correct me if I'm wrong), this never worked before, as you need a transaction in order to use LOBs with Oracle, but now this is more explicit.

IOW the fix is to have a transaction around the code working with LOBs.

avpalienko:
It worked at 047b74915bb9e8a259536f6d3596bf4bfefe36d2

vadz:
Could you please use git-bisect which exact commit broke it? I see 8b5b312d (Fix Oracle bug that prevented selecting into initialized BLOB, 2023-11-01) which might be related...

Krzmbrzl:
Tbh I don't quite remember the exact semantics of Oracle. Iirc there were/are situations in which you didn't need an explicit transaction... it was a bit of an odd situation, that much I know 👀

avpalienko:
I don't know - in my fork it worked.  Fork is based on 8ddddcac3073a7f33dbfdea318a6c9ad45f639db
I try rebase to the last commit of master and found the problem

avpalienko:
> Could you do a bisect to find the commit that causes the issue?

I'll try, but I have no experience with bisect. So it will take a while

avpalienko:
I found one more problem. 
If rowset contains more than one row move_as() does access violation ( use after move )

```
diff --git a/tests/common-tests.h b/tests/common-tests.h
index d2b8fb42..ffc80fc6 100644
--- a/tests/common-tests.h
+++ b/tests/common-tests.h
@@ -6779,7 +6779,7 @@ TEST_CASE_METHOD(common_tests, "BLOB", "[core][blob]")
         }
         SECTION("move_as")
         {
-            soci::rowset< soci::row > rowSet = (sql.prepare << "select b from soci_test where id=:id", soci::use(id));
[...]

PR #1142, opened 2024-04-15
Fix blob errors in Oracle backend  issue #1141 

avpalienko:
The LOB opening and closing mechanism in Oracle has some restrictions which makes it implicitly usage is dangerous This commit remove the call OCILobOpen from fetching and add some oracle blob tests

### DIFF
--- a/tests/oracle/test-oracle.cpp
+++ b/tests/oracle/test-oracle.cpp
@@ -179,7 +179,7 @@ struct basic_table_creator : public table_creator_base
     }
 };
 
-TEST_CASE("Oracle nested statement", "[oracle][blob]")
+TEST_CASE("Oracle nested statement", "[oracle][nested]")
 {
     soci::session sql(backEnd, connectString);
     basic_table_creator tableCreator(sql);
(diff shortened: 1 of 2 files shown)


vadz:
Thanks for the fixes!

I don't know much about LOBs in Oracle, but we definitely shouldn't be keeping this code commented out, if we prefer/have to rely on implicit LOB opening, it should be just removed.

@Krzmbrzl Any comments?

vadz:
> > Thanks for the fixes!
> > I don't know much about LOBs in Oracle, but we definitely shouldn't be keeping this code commented out, if we prefer/have to rely on implicit LOB opening, it should be just removed.
> 
> It's commented as TODO. It will be good to have an explicit Open/Close for oracle blob in future Oracle says "These functions allow an application to mark the beginning and end of a series of LOB operations so that specific processing (e.g., updating indexes, etc.) can be performed when a LOB is closed."

I think we need to either do it now, if it's possible, or remove this completely. Keeping a chunk of code commented out is just never a good idea.
 
> Could you restart pipelines, please?

I had already done it and just did it again but they just keep failing, I have no idea what's going on here but will try to look tomorrow if it still doesn't work then.

Krzmbrzl:
Do the regular Blob test cases (the Backend independent ones) still work with these changes? I thought I needed explicit opening/closing for consistent-ish behavior across?backends but I don't recall the exact details 👀

avpalienko:
I've deleted commented code

avpalienko:
> Do the regular Blob test cases (the Backend independent ones) still work with these changes? I thought I needed explicit opening/closing for consistent-ish behavior across?backends but I don't recall the exact details 👀

All existing tests are passed

vadz:
@Krzmbrzl Do you have any objections to merging this?

tests/oracle/test-oracle.cpp
```diff
@@ -1387,6 +1387,69 @@ TEST_CASE("Bulk iterators", "[oracle][bulkiters]")
...
+        // b.append(buf, sizeof(buf));
+        // assert(b.get_len() == sizeof(buf) + 10);
+        
+    }
+    sql.commit ();
+
+    {
+        blob b ( sql );
+        sql << "select img from soci_test where id = 7", into ( b );
+        // assert(b.get_len() == sizeof(buf) + 10);
```
Krzmbrzl, line 1430:
```suggestion
```

tests/oracle/test-oracle.cpp
```diff
@@ -1387,6 +1387,69 @@ TEST_CASE("Bulk iterators", "[oracle][bulkiters]")
...
+    blob_table_creator ( soci::session &sql ) : table_creator_base ( sql )
+    {
+        sql << "create table soci_test ("
+               "    id number(10) not null,"
+               "    img blob"
+               ")";
+    }
+};
+
+TEST_CASE ( "Oracle blob", "[oracle][blob]" )
```
Krzmbrzl, line 1401: What exactly is this test case testing that is not covered in the test cases shared across backends? The reasoning for having a Oracle-specific Blob test case should probably be written down here as a comment

Krzmbrzl:
> Do you have any objections to merging this?

No - as long as the common Blob interface remains functional (which seems to be the case), I'm fine with the changes.

tests/oracle/test-oracle.cpp
avpalienko: I have restored the code from previous commits. We can remove it and use common test, but we need add test cases using blob before start transaction and after commit/rollback

Krzmbrzl: > we need add test cases using blob before start transaction and after commit/rollback

This would be an Oracle-specific test then as this is undefined behavior in a cross-DB setting (as specified in SOCI docs). So for this specific scenario, a backend-specific test case makes sense

vadz:
Sorry, just to confirm: is this still needed even with/after #1145?

avpalienko:
> Sorry, just to confirm: is this still needed even with/after #1145?

I think no

PR #1145, opened 2024-04-21
Fix `row::move_as` behavior breaking reusability

### LINKED ISSUE #1144: `rowset::move_as` can lead to segmentation faults
Originally reported by @avpalienko in https://github.com/SOCI/soci/issues/1141#issuecomment-2057755837

When applying the following patch to the `common-tests.h` file, the `move_as` test case will crash due to a segmentation fault:

```patch
diff --git a/tests/common-tests.h b/tests/common-tests.h
index d2b8fb42..ffc80fc6 100644
--- a/tests/common-tests.h
+++ b/tests/common-tests.h
@@ -6779,7 +6779,7 @@ TEST_CASE_METHOD(common_tests, "BLOB", "[core][blob]")
         }
         SECTION("move_as")
         {
-            soci::rowset< soci::row > rowSet = (sql.prepare << "select b from soci_test where id=:id", soci::use(id));
+            soci::rowset< soci::row > rowSet = (sql.prepare << "select b from soci_test where id=:id union all select b from soci_test where id=:id", soci::use(id, "id"));
             bool containedData = false;
             for (auto it = rowSet.begin(); it != rowSet.end(); ++it)
             {
```

The issue was originally reported to happen with Oracle, but I was also able to reproduce it with SQLite3, indicating that the issue is not backend-specific.

Krzmbrzl:
For the time being, this is merely a sketch of how I think the fix has to work. The current implementation fails at the unavailability of a `session` object at the `row` scope...

Once this is finished, this PR fixes #1144

### DIFF
--- a/src/core/row.cpp
+++ b/src/core/row.cpp
@@ -121,7 +121,7 @@ blob row::move_as<blob>(std::size_t pos) const
     type_conversion<blob>::move_from_base(baseVal, *indicators_.at(pos), ret);
 
     // Re-initialize blob object so it can be used in further queries
-    baseVal.initialize(session);
+    baseVal.initialize(ret.get_backend()->get_session_backend().make_blob_backend());
 
     return ret;
 }
--- a/src/core/row.cpp
+++ b/src/core/row.cpp
@@ -119,5 +119,9 @@ blob row::move_as<blob>(std::size_t pos) const
 
     blob ret;
     type_conversion<blob>::move_from_base(baseVal, *indicators_.at(pos), ret);
+
+    // Re-initialize blob object so it can be used in further queries
+    baseVal.initialize(session);
+
     return ret;
 }
--- a/include/private/soci-trivial-blob-backend.h
+++ b/include/private/soci-trivial-blob-backend.h
@@ -2,6 +2,7 @@
 #define SOCI_PRIVATE_SOCI_TRIVIAL_BLOB_BACKEND_H_INCLUDED
 
 #include "soci/soci-backend.h"
+#include "soci/session.h"
 
 #include <vector>
 #include <cstring>
@@ -22,6 +23,8 @@ namespace details
 class trivial_blob_backend : public details::blob_backend
 {
 public:
+    trivial_blob_backend(details::session_backend &backend) : session_(backend) {}
+
     std::size_t get_len() override { return buffer_.size(); }
 
     std::size_t read_from_start(void* buf, std::size_t toRead,
@@ -71,7 +74,10 @@ public:
 
     const std::uint8_t *get_buffer() const { return buffer_.data(); }
 
+    details::session_backend &get_session_backend() override { return session_; }
+
 protected:
+    details::session_backend &session_;
     std::vector< std::uint8_t > buffer_;
 };
 
--- a/include/soci/blob.h
+++ b/include/soci/blob.h
@@ -40,6 +40,7 @@ public:
 
     // (Re)initializes this blob
     void initialize(session &s);
+    void initialize(details::blob_backend *backend);
 
     std::size_t get_len();
 
--- a/include/soci/db2/soci-db2.h
+++ b/include/soci/db2/soci-db2.h
@@ -233,6 +233,7 @@ struct db2_blob_backend : details::blob_backend
     std::size_t write_from_start(const void* buf, std::size_t toWrite, std::size_t offset = 0) override;
     std::size_t append(const void* buf, std::size_t toWrite) override;
     void trim(std::size_t newLen) override;
+    details::session_backend &get_session_backend() override;
 
     db2_session_backend& session_;
 };
--- a/include/soci/empty/soci-empty.h
+++ b/include/soci/empty/soci-empty.h
@@ -146,6 +146,8 @@ struct empty_blob_backend : details::blob_backend
     std::size_t append(const void* buf, std::size_t toWrite) override;
     void trim(std::size_t newLen) override;
 
+    details::session_backend &get_session_backend() override;
+
     empty_session_backend& session_;
 };
 
--- a/include/soci/firebird/soci-firebird.h
+++ b/include/soci/firebird/soci-firebird.h
@@ -266,6 +266,8 @@ struct firebird_blob_backend : details::blob_backend
     std::size_t append(const void *buf, std::size_t toWrite) override;
     void trim(std::size_t newLen) override;
 
+    details::session_backend &get_session_backend() override;
+
     // Writes the current data into the database by allocating a new BLOB
     // object for it.
     //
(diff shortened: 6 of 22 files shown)


src/core/row.cpp
```diff
@@ -119,5 +119,9 @@ blob row::move_as<blob>(std::size_t pos) const
 
     blob ret;
     type_conversion<blob>::move_from_base(baseVal, *indicators_.at(pos), ret);
+
+    // Re-initialize blob object so it can be used in further queries
+    baseVal.initialize(session);
```
Krzmbrzl, line 124: @vadz any idea how I could access a `session` object in this place? I need one as it is the only object that can create blob backend objects, which is required to get a fully functioning `blob` again.

An alternative I have considered is to write a `clone()` or `duplicate()` function for the `blob_backend` interface that can be used here (via some interface from the `blob` object) but that seems rather inconsistent with how things work for all other backend objects. Besides, that would essentially require duplicating the functionality of `session::make_blob_backend()` and code duplication is never a good thing...

Yet another alternative I was thinking of was to keep track of moved-from objects at the `row` level in order to re-initialize them only when the object is actually used again, but I don't think there currently exists a (nice) way of executing code before new data is bound to the row's holders (as they are simply referenced as `into` holders in the respective statement.

I'd appreciate thoughts from your side on this.

vadz: This would require storing a back-pointer (or reference) to the session in each row object. It should probably be possible, but I wonder if we can avoid moving out of the blob here to avoid having to do this. Could you please remind me why was it necessary to add `move_as()` in the first place?

Krzmbrzl: Because blob objects are not copyable, which the standard `get` API does

vadz: Yes, but couldn't we just provide access to their data instead?

Krzmbrzl: This would require copying data into some sort of buffer (plus getting the type conversion magic in SOCI's internals right), which might not be what the user wants. Just imagine the scenario in which I want to fetch a blob just to write it into a different table. That'd then require copying the data into my temp buffer just to yet again copy the data into a new `blob` object that I can then `use` in my insertion query.

vadz: I don't know how common is this scenario and I think you could still do it using base API (and not rowset-based one)? So I'm really not sure if we want to complicate things just to support that...

Another possibility would be to implement copying for blobs, by using copy-on-write for them. But this also will add a lot of complexity. Maybe it's worth it, but I'm not sure.

Krzmbrzl: There's also the issue of having an easily understandable interface for users and given that SOCI has a dedicated `blob` object type, the expected thing (in my experience) is that one can fetch a `blob` object from a row (see also https://github.com/SOCI/soci/issues/922).
If you try that with the current code using `get<blob>` there will be a message telling you that you'll have to use `move_as<blob>`, which I suppose is still good enough in terms of users figuring out what to do with blobs.

So I kinda like the idea of having this functionality available. :eyes: 

A hack that I could think of would be adding a `session` object to the call signature of `move_as`:
```cpp
blob b = row.move_as<blob>(session);
```

I suppose that this wouldn't limit the usability for the user as they (presumably) have a `session` object at hand anyway when operating on `row`s :thinking:

vadz: If you expect blobs to have value-like semantics, you have to implement support for copying them (which means using COW).

Passing `session` to `move_as()` is not great because it would be possible to pass a _different_ session here, which is something a well-designed API shouldn't allow.

avpalienko: In my fork I use 
typedef std::shared_ptr<blob> blob_sptr; 

and 

        auto b = r.get<blob_sptr> ( 0 );
        CHECK ( b->get_len () == 10 );

Krzmbrzl: Pointers to blobs don't fix the issue, I think. You can still end up moving from the underlying blob elsewhere thereby breaking the `row` object again.

> If you expect blobs to have value-like semantics, you have to implement support for copying them (which means using COW).

Value-like as a move-only object. That's already fulfilled. And COW has the same underlying issue that we are trying to fix here, no? We need a way to create another blob backend, which can only be done via a `session` instance.

> Passing session to move_as() is not great because it would be possible to pass a different session here, which is something a well-designed API shouldn't allow.

True :thinking: 

How bad would adding a `session` member (as pointer) to `row` be? :thinking:

avpalienko: > Pointers to blobs don't fix the issue, I think. You can still end up moving from the underlying blob elsewhere thereby breaking the `row` object again.
> 

This requires intentional user action. User could break a string, e.g.
string s;
sql.exchange(into(s));
string t = move(s);

Krzmbrzl: Yes but for `row` objects it is not at all obvious that the object the user receives is part of the internal state.

vadz: I think the simplest would be to add `blob_impl::clone()`.

avpalienko: Is it possible for the writable blob received from the select after insert ?

Krzmbrzl: > Is it possible for the writable blob received from the select after insert ?

Sorry, I can't quite follow - there seems to be a verb missing in your sentence (that's important for your meaning) :eyes:

Krzmbrzl:
@avpalienko could you debug where the SegFault for Oracle is coming from? You seem to have good access to an Oracle installation, whereas for me it is always super tedious to get something to work.

src/core/row.cpp
avpalienko: > > Is it possible for the writable blob received from the select after insert ?
> 
> Sorry, I can't quite follow - there seems to be a verb missing in your sentence (that's important for your meaning) 👀

Sorry, I mean blob_impl::clone. Is it possible to clone the blob? Who will own the blob locator ?

Krzmbrzl: Cloning is not a good idea, I think. We don't want to duplicate the stored data. But we can create a new, empty blob from an existing one.

avpalienko:
> @avpalienko could you debug where the SegFault for Oracle is coming from? You seem to have good access to an Oracle installation, whereas for me it is always super tedious to get something to work.

I've debugged on windows and have no segfault. But test fails:

-------------------------------------------------------------------------------
BLOB
  Blob binding
  move_as
-------------------------------------------------------------------------------
D:\CMA\Dev\SOCI\soci_githab\tests\common-tests.h(6784)
...............................................................................

D:\CMA\Dev\SOCI\soci_githab\tests\common-tests.h(6784): FAILED:
  {Unknown expression after the reported line}
  soci error: Invalid handle while fetching data from "select b from soci_test
  where id=:id" with :1=42.

===============================================================================
test cases:   2 |   1 passed | 1 failed
assertions: 162 | 161 passed | 1 failed



I think it's a result of the move_as. Lob locator binds to the statement in define_by_pos(): 

bbe->reset();
ociData_ = bbe->get_lob_locator();

and then it initialized by OCIDefineByPos(). But in  row::move_as<blob>() we call dtor ~oracle_blob_backend() with OCIDescriptorFree inside.  On the next fetch in the post_fetch we call bbe->set_lob_locator(). Inside it we set locator in the old released value and call OCILobIsOpen which returns OCI_INVALID_HANDLE

Krzmbrzl:
Finally! I think with my latest commit I have found a way to fix the underlying issue :rocket: 

@avpalienko sorry that I took so long to look into this again. Your debug insights were still very valuable though, so thanks for having taken the time :pray: 

@vadz with all CI now passing, I think this is ready to be merged or at least reviewed.

PR #1154, opened 2024-07-04
Add support for building on SunOS with gcc

snaiffer:
I've tested it on SunOS 5.11.3 with gcc (GCC) 13.1.0

### DIFF
--- a/include/soci/exchange-traits.h
+++ b/include/soci/exchange-traits.h
@@ -44,12 +44,15 @@ struct exchange_traits
     };
 };
 
+#if SOCI_OS != SOCI_OS_SUN
+// As in SunOS: typedef char int8_t;
 template <>
 struct exchange_traits<int8_t>
 {
     typedef basic_type_tag type_family;
     enum { x_type = x_int8 };
 };
+#endif
 
 template <>
 struct exchange_traits<uint8_t>
--- a/include/soci/soci-types.h
+++ b/include/soci/soci-types.h
@@ -6,7 +6,7 @@
 #if defined(__GNUC__) || defined(__clang__)
     #if defined(__LP64__)
         #define SOCI_LONG_IS_64_BIT 1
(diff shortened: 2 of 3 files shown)


include/soci/exchange-traits.h
```diff
@@ -44,12 +44,15 @@ struct exchange_traits
     };
 };
 
+#if SOCI_OS != SOCI_OS_SUN
+// As in SunOS: typedef char int8_t;
```
vadz, line 48: I don't think it's a property of Solaris per se, we should almost certainly check whether `char` and `int8_t` are the same in CMake and define some `SOCI_HAVE_INT8_T_TYPEDEF` or something like this and test for it here.

include/soci/soci-types.h
```diff
@@ -6,7 +6,7 @@
 #if defined(__GNUC__) || defined(__clang__)
     #if defined(__LP64__)
         #define SOCI_LONG_IS_64_BIT 1
-        #if SOCI_OS == SOCI_OS_LINUX || SOCI_OS == SOCI_OS_FREE_BSD
+        #if SOCI_OS == SOCI_OS_LINUX || SOCI_OS == SOCI_OS_FREE_BSD || SOCI_OS == SOCI_OS_SUN
```
vadz, line 9: Here too we should just check for the condition directly instead of testing for the platform... In fact, I really don't know why did we do it like this in the first place, @zann1x do you remember if there was some good reason for it?

vadz (commented):
Thanks for the fix but I'd really prefer to have generic tests for the types instead of testing for the individual platforms. If you can update CMake to do it instead, this would be very welcome. TIA!

include/soci/soci-types.h
zann1x: I don't think there was any other reason for it than to just make it work both locally and in the CI pipeline. In the end, it's only important that we know which data type definitions the underlying OS uses.

vadz:
I've finally implemented my own proposal in #1170, please check if it works for you and let me know if it doesn't. Thanks!

PR #1156, opened 2024-07-23
Using connection string schema when working with metadata.

cstiborg:
Adding possiblity for database backends to register the schema to use when working with metadata.

Currently this is hardcoded to public.

PostgreSQL and MySQL have been implemented.

### DIFF
--- a/include/private/soci-compiler.h
+++ b/include/private/soci-compiler.h
@@ -36,4 +36,22 @@
 #   define SOCI_GCC_WARNING_RESTORE(x)
 #endif
 
+// SOCI_FALLTHROUGH macro can be used on both g++, mingw, and msvc to generate the
+// correct behavior for expected fallthrough in switch statements.
+#if defined(__GNUC__) || defined(__MINGW32__) || defined(__MINGW64__)
+#   if __cplusplus >= 201402L
+#       define SOCI_FALLTHROUGH [[fallthrough]]
+#   else
+#       define SOCI_FALLTHROUGH ((void)0)
+#   endif
+#elif defined(_MSC_VER)
+#   if _MSC_VER >= 1910 && __cplusplus >= 201703L
+#       define SOCI_FALLTHROUGH [[fallthrough]]
+#   else
+#       define SOCI_FALLTHROUGH ((void)0)
+#   endif
+#else
+#   define SOCI_FALLTHROUGH ((void)0)
+#endif
+
 #endif // SOCI_PRIVATE_SOCI_COMPILER_H_INCLUDED
--- a/include/private/soci-compiler.h
+++ b/include/private/soci-compiler.h
@@ -36,22 +36,36 @@
 #   define SOCI_GCC_WARNING_RESTORE(x)
 #endif
 
-// SOCI_FALLTHROUGH macro can be used on both g++, mingw, and msvc to generate the
-// correct behavior for expected fallthrough in switch statements.
-#if defined(__GNUC__) || defined(__MINGW32__) || defined(__MINGW64__)
-#   if __cplusplus >= 201402L
-#       define SOCI_FALLTHROUGH [[fallthrough]]
-#   else
-#       define SOCI_FALLTHROUGH ((void)0)
-#   endif
-#elif defined(_MSC_VER)
-#   if _MSC_VER >= 1910 && __cplusplus >= 201703L
-#       define SOCI_FALLTHROUGH [[fallthrough]]
-#   else
-#       define SOCI_FALLTHROUGH ((void)0)
-#   endif
+// CHECK_CXX_STD(version) evaluates to 1 if the C++ version is at least the
+// version specified.
+#if defined(_MSVC_LANG)
+#    define SOCI_CHECK_CXX_STD(ver) (_MSVC_LANG >= (ver))
+#elif defined(__cplusplus)
+#    define SOCI_CHECK_CXX_STD(ver) (__cplusplus >= (ver))
 #else
-#   define SOCI_FALLTHROUGH ((void)0)
+#    define SOCI_CHECK_CXX_STD(ver) 0
+#endif
+
+// HAS_CLANG_FEATURE(feature) evaluates to 1 if clang is available and the
+// provided feature is available.
+#ifndef __has_feature
+#   define SOCI_HAS_CLANG_FEATURE(x) 0
+#else
+#   define SOCI_HAS_CLANG_FEATURE(x) __has_feature(x)
+#endif
+
+// FALLTHROUGH macro defines a cross-platform/version to mark fallthrough
+// behavior in switch statement.
+#if SOCI_CHECK_CXX_STD(201703L)
+#    define SOCI_FALLTHROUGH [[fallthrough]]
+#elif defined(__has_warning) && SOCI_HAS_CLANG_FEATURE(cxx_attributes)
+#    define SOCI_FALLTHROUGH [[clang::fallthrough]]
+#elif defined(__GNUC__) && __has_cpp_attribute(fallthrough)
+#    define SOCI_FALLTHROUGH [[fallthrough]]
+#endif
+
+#ifndef SOCI_FALLTHROUGH
+    #define SOCI_FALLTHROUGH ((void)0)
 #endif
 
 #endif // SOCI_PRIVATE_SOCI_COMPILER_H_INCLUDED
--- a/include/private/soci-compiler.h
+++ b/include/private/soci-compiler.h
@@ -58,7 +58,7 @@
 // behavior in switch statement.
 #if SOCI_CHECK_CXX_STD(201703L)
 #    define SOCI_FALLTHROUGH [[fallthrough]]
-#elif defined(__has_warning) && SOCI_HAS_CLANG_FEATURE(cxx_attributes)
+#elif SOCI_HAS_CLANG_FEATURE(cxx_attributes)
 #    define SOCI_FALLTHROUGH [[clang::fallthrough]]
 #elif defined(__GNUC__)
 #    if defined(__has_cpp_attribute)
--- a/include/private/soci-compiler.h
+++ b/include/private/soci-compiler.h
@@ -60,8 +60,12 @@
 #    define SOCI_FALLTHROUGH [[fallthrough]]
 #elif defined(__has_warning) && SOCI_HAS_CLANG_FEATURE(cxx_attributes)
 #    define SOCI_FALLTHROUGH [[clang::fallthrough]]
-#elif defined(__GNUC__) && __has_cpp_attribute(fallthrough)
-#    define SOCI_FALLTHROUGH [[fallthrough]]
+#elif defined(__GNUC__)
+#    if defined(__has_cpp_attribute)
+#        if __has_cpp_attribute(fallthrough)
+#            define SOCI_FALLTHROUGH [[fallthrough]]
+#        endif
+#    endif
 #endif
 
 #ifndef SOCI_FALLTHROUGH
--- a/include/soci/column-info.h
+++ b/include/soci/column-info.h
@@ -11,7 +11,6 @@
 #include "soci/soci-backend.h"
 #include "soci/type-conversion.h"
 #include "soci/values.h"
-
 #include <cstdint>
 
 namespace soci
@@ -90,10 +89,17 @@ struct type_conversion<column_info>
             ci.dataType = db_string;
         }
         else if (type_name == "integer" || type_name == "INTEGER" ||
-		 type_name == "int")
+		 type_name == "int" || type_name == "INT" ||
+		 type_name == "bigint" || type_name == "BIGINT")
         {
             ci.type = dt_integer;
             ci.dataType = db_int32;
+        }
+	else if (type_name.find("double") != std::string::npos ||
+            type_name.find("DOUBLE") != std::string::npos)
+	{
+            ci.type = dt_double;
+            ci.dataType = db_double;
         }
         else if (type_name.find("number") != std::string::npos ||
             type_name.find("NUMBER") != std::string::npos ||
--- a/include/soci/column-info.h
+++ b/include/soci/column-info.h
@@ -89,15 +89,19 @@ struct type_conversion<column_info>
             ci.dataType = db_string;
         }
         else if (type_name == "integer" || type_name == "INTEGER" ||
-		 type_name == "int" || type_name == "INT" ||
-		 type_name == "bigint" || type_name == "BIGINT")
+                 type_name == "int" || type_name == "INT")
         {
             ci.type = dt_integer;
             ci.dataType = db_int32;
         }
-	else if (type_name.find("double") != std::string::npos ||
+        else if (type_name == "bigint" || type_name == "BIGINT")
+        {
+            ci.type = dt_long_long;
+            ci.dataType = db_int64;
+        }
+        else if (type_name.find("double") != std::string::npos ||
             type_name.find("DOUBLE") != std::string::npos)
-	{
+        {
             ci.type = dt_double;
             ci.dataType = db_double;
         }
--- a/include/soci/column-info.h
+++ b/include/soci/column-info.h
@@ -89,7 +89,8 @@ struct type_conversion<column_info>
             ci.type = dt_string;
             ci.dataType = db_string;
         }
-        else if (type_name == "integer" || type_name == "INTEGER")
+        else if (type_name == "integer" || type_name == "INTEGER" ||
+		 type_name == "int")
         {
             ci.type = dt_integer;
             ci.dataType = db_int32;
--- a/include/soci/mysql/soci-mysql.h
+++ b/include/soci/mysql/soci-mysql.h
@@ -284,24 +284,24 @@ struct mysql_session_backend : details::session_backend
 
     std::string get_table_names_query() const override
     {
-        return "SELECT table_name AS 'TABLE_NAME' FROM information_schema.tables WHERE table_schema = DATABASE()";
+        return R"delim(SELECT table_name AS 'TABLE_NAME' FROM information_schema.tables WHERE table_schema = DATABASE())delim";
     }
 
     std::string get_column_descriptions_query() const override
     {
-        return "SELECT column_name as \"COLUMN_NAME\","
-            " data_type as \"DATA_TYPE\","
-            " character_maximum_length as \"CHARACTER_MAXIMUM_LENGTH\","
-            " numeric_precision as \"NUMERIC_PRECISION\","
-            " numeric_scale as \"NUMERIC_SCALE\","
-            " is_nullable as \"IS_NULLABLE\""
-            " from information_schema.columns"
-            " where"
-            " case"
-            " when :s is not NULL THEN table_schema = :s"
-            " else table_schema = DATABASE()"
-            " end"
-            " and table_name = :t";
+        return R"delim(SELECT column_name as "COLUMN_NAME",
+            data_type as "DATA_TYPE",
+            character_maximum_length as "CHARACTER_MAXIMUM_LENGTH",
+            numeric_precision as "NUMERIC_PRECISION",
+            numeric_scale as "NUMERIC_SCALE",
+            is_nullable as "IS_NULLABLE"
+            from information_schema.columns
+            where
+            case
+            when :s is not NULL THEN table_schema = :s
+            else table_schema = DATABASE()
+            end
+            and table_name = :t)delim";
     }
 
     MYSQL *conn_;
--- a/include/soci/mysql/soci-mysql.h
+++ b/include/soci/mysql/soci-mysql.h
@@ -284,7 +284,7 @@ struct mysql_session_backend : details::session_backend
 
     std::string get_table_names_query() const override
     {
-        return R"delim(SELECT table_name AS 'TABLE_NAME' FROM information_schema.tables WHERE table_schema = DATABASE())delim";
+        return R"delim(SELECT LOWER(table_name) AS 'TABLE_NAME' FROM information_schema.tables WHERE table_schema = DATABASE())delim";
     }
 
     std::string get_column_descriptions_query() const override
@@ -301,7 +301,7 @@ struct mysql_session_backend : details::session_backend
             when :s is not NULL THEN table_schema = :s
             else table_schema = DATABASE()
             end
-            and table_name = :t)delim";
+            and UPPER(table_name) = UPPER(:t))delim";
     }
 
     MYSQL *conn_;
--- a/include/soci/mysql/soci-mysql.h
+++ b/include/soci/mysql/soci-mysql.h
@@ -287,6 +287,23 @@ struct mysql_session_backend : details::session_backend
         return "SELECT table_name AS 'TABLE_NAME' FROM information_schema.tables WHERE table_schema = DATABASE()";
     }
 
+    std::string get_column_descriptions_query() const override
+    {
+        return "SELECT column_name as \"COLUMN_NAME\","
+            " data_type as \"DATA_TYPE\","
+            " character_maximum_length as \"CHARACTER_MAXIMUM_LENGTH\","
+            " numeric_precision as \"NUMERIC_PRECISION\","
+            " numeric_scale as \"NUMERIC_SCALE\","
+            " is_nullable as \"IS_NULLABLE\""
+            " from information_schema.columns"
+            " where"
+            " case"
+            " when :s is not NULL THEN table_schema = :s"
+            " else table_schema = DATABASE()"
+            " end"
+            " and table_name = :t";
+    }
+
     MYSQL *conn_;
 };
 
--- a/include/soci/session.h
+++ b/include/soci/session.h
@@ -25,6 +25,7 @@ namespace soci
 {
 class values;
 class backend_factory;
+struct schema_table_name;
 
 namespace details
 {
@@ -146,7 +147,8 @@ public:
     // Since this is intended for use with statement objects, where results are obtained one row after another,
     // it makes sense to bind either std::string for each output field or soci::column_info for the whole row.
     // Note: table_name is a non-const reference to prevent temporary objects,
-    // this argument is bound as a regular "use" element.
+    // this argument is bound as a regular "use" element. The table_name can consist of both a schema name and
+    // a table_name separated by a dot.
     details::prepare_temp_type prepare_column_descriptions(std::string & table_name);
 
     // Functions for basic portable DDL statements.
@@ -215,6 +217,8 @@ private:
     SOCI_NOT_COPYABLE(session)
 
     void reset_after_move();
+    struct schema_table_name * alloc_schema_table_name(std::string & tableName);
+    void clean_schema_table_names();
 
     std::ostringstream query_stream_;
     std::unique_ptr<details::query_transformation_function> query_transformation_;
@@ -232,6 +236,8 @@ private:
     bool isFromPool_;
     std::size_t poolPosition_;
     connection_pool * pool_;
+
+    struct schema_table_name * schema_table_name_;
 };
 
 } // namespace soci
--- a/include/soci/session.h
+++ b/include/soci/session.h
@@ -218,7 +218,7 @@ private:
     SOCI_NOT_COPYABLE(session)
 
     void reset_after_move();
-    schema_table_name * alloc_schema_table_name(std::string & tableName);
+    schema_table_name * alloc_schema_table_name(const std::string& tableName);
 
     std::ostringstream query_stream_;
     std::unique_ptr<details::query_transformation_function> query_transformation_;
--- a/include/soci/session.h
+++ b/include/soci/session.h
@@ -20,6 +20,7 @@
 #include <ostream>
 #include <sstream>
 #include <string>
+#include <forward_list>
 
 namespace soci
 {
@@ -217,8 +218,7 @@ private:
     SOCI_NOT_COPYABLE(session)
 
     void reset_after_move();
-    struct schema_table_name * alloc_schema_table_name(std::string & tableName);
-    void clean_schema_table_names();
+    schema_table_name * alloc_schema_table_name(std::string & tableName);
 
     std::ostringstream query_stream_;
     std::unique_ptr<details::query_transformation_function> query_transformation_;
@@ -237,7 +237,7 @@ private:
     std::size_t poolPosition_;
     connection_pool * pool_;
 
-    struct schema_table_name * schema_table_name_;
+    std::forward_list<schema_table_name> schema_table_name_;
 };
 
 } // namespace soci
--- a/include/soci/session.h
+++ b/include/soci/session.h
@@ -218,7 +218,7 @@ private:
     SOCI_NOT_COPYABLE(session)
 
     void reset_after_move();
-    schema_table_name * alloc_schema_table_name(const std::string& tableName);
+    schema_table_name& alloc_schema_table_name(const std::string& tableName);
 
     std::ostringstream query_stream_;
     std::unique_ptr<details::query_transformation_function> query_transformation_;
@@ -237,6 +237,10 @@ private:
     std::size_t poolPosition_;
     connection_pool * pool_;
 
+    // Storing schema_table_names in a forward list as these are required
+    // as persistent input to prepare_temp_type object during their life-
+    // span. The prepare_temp_type uses the addresses of the content of the
+    // schema_table_name_ thus, a container which doesn't move data is used.
     std::forward_list<schema_table_name> schema_table_name_;
 };
 
--- a/include/soci/soci-backend.h
+++ b/include/soci/soci-backend.h
@@ -332,7 +332,7 @@ private:
 class session_backend
 {
 public:
-    session_backend() : failoverCallback_(NULL), session_(NULL), schema_name_("public") {}
+    session_backend() : failoverCallback_(NULL), session_(NULL) {}
     virtual ~session_backend() {}
 
     virtual bool is_connected() = 0;
@@ -366,7 +366,7 @@ public:
     {
         return "select table_name as \"TABLE_NAME\""
             " from information_schema.tables"
-            " where table_schema = '" + schema_name_ + "'";
+            " where table_schema = 'public'";
     }
 
     // Returns a query with a single parameter (table name) for the list
@@ -380,7 +380,7 @@ public:
             " numeric_scale as \"NUMERIC_SCALE\","
             " is_nullable as \"IS_NULLABLE\""
             " from information_schema.columns"
-            " where table_schema = '" + schema_name_ + "' and table_name = :t";
+            " where table_schema = 'public' and table_name = :t";
     }
 
     virtual std::string create_table(const std::string & tableName)
@@ -565,7 +565,6 @@ public:
 
     failover_callback * failoverCallback_;
     session * session_;
-    std::string schema_name_;
 
 private:
     SOCI_NOT_COPYABLE(session_backend)
--- a/include/soci/soci-backend.h
+++ b/include/soci/soci-backend.h
@@ -332,7 +332,7 @@ private:
 class session_backend
 {
 public:
-    session_backend() : failoverCallback_(NULL), session_(NULL) {}
+    session_backend() : failoverCallback_(NULL), session_(NULL), schema_name_("public") {}
     virtual ~session_backend() {}
 
     virtual bool is_connected() = 0;
@@ -366,7 +366,7 @@ public:
     {
         return "select table_name as \"TABLE_NAME\""
             " from information_schema.tables"
-            " where table_schema = 'public'";
+            " where table_schema = '" + schema_name_ + "'";
     }
 
     // Returns a query with a single parameter (table name) for the list
@@ -380,7 +380,7 @@ public:
             " numeric_scale as \"NUMERIC_SCALE\","
             " is_nullable as \"IS_NULLABLE\""
             " from information_schema.columns"
-            " where table_schema = 'public' and table_name = :t";
+            " where table_schema = '" + schema_name_ + "' and table_name = :t";
     }
 
     virtual std::string create_table(const std::string & tableName)
@@ -565,6 +565,7 @@ public:
 
     failover_callback * failoverCallback_;
     session * session_;
+    std::string schema_name_;
 
 private:
     SOCI_NOT_COPYABLE(session_backend)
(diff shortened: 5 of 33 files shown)


cstiborg:
I'm trying to figure out what is going on with the PostgreSQL check - I can reproduce the error in a clean SOCI master build - but not in my cstiborg master.

include/soci/soci-backend.h
```diff
@@ -366,7 +366,7 @@ class session_backend
     {
         return "select table_name as \"TABLE_NAME\""
             " from information_schema.tables"
-            " where table_schema = 'public'";
+            " where table_schema = '" + schema_name_ + "'";
```
vadz, line 369: Shouldn't we use a parameter for this? I'm not sure if the schema name will always be under program control and if it isn't, then this is a flagrant SQL injection problem.

src/backends/postgresql/session.cpp
```diff
@@ -67,6 +67,16 @@ void postgresql_session_backend::connect(
                          : "SET extra_float_digits = 2",
         "Cannot set extra_float_digits parameter");
 
+    PGresult* res = PQexec(conn, "SHOW search_path");
+    if (PQresultStatus(res) == PGRES_TUPLES_OK)
+    {
+        if (PQntuples(res) > 0)
+        {
+            schema_name_ = PQgetvalue(res, 0, 0);
```
vadz, line 75: How does this work? Search path is something like `"$user", public` by default for me, which can't be used as the schema name...

include/soci/soci-backend.h
cstiborg: The variable _should_ be managed by backends - but I totally agree that it should be a parameter, I was just a bit in doubt how to achieve this while not changing the API and contracts.

src/backends/postgresql/session.cpp
cstiborg: You are right, I lost track of how this works. Search path in PostgreSQL can hold 0 or more comma separated schema names, if it holds 0, public will be the default, if it holds more than one, I am actually not sure how it works. I assume (this must indeed be verified) that for DDL creates it will use the first schema to create into, and for any other operation it'll keep searching until the first table it finds.

No matter how it works it of course changes how PR should alter the current behavior that I still believe is not great for anybody using a schema which is not public.

vadz: Yes, it uses the first existing schema for creating the new tables and any existing schema for lookup, see [the docs](https://www.postgresql.org/docs/current/ddl-schemas.html#DDL-SCHEMAS-PATH).

cstiborg:
While MySQL and Sqlite seems to be no brainers, PostgreSQL or any of the other supported databases which in fact do support schemas requires some decisions to be made.

I did start out with an attempt of adding an extra parameter to the metadata methods - however, I got discouraged when I realized that:
1. It's not necessary for MySQL and SQLite
2. For consistency it would extent to the DDL code as well, despite it not having the same issue of public being hardcoded.

I then thought, maybe foolishly, that when you connect to the database using a search_path, PostgreSQL will (I know PostgreSQL the best of all the backends supported and I lack the infrastructure and experience to test any changes on other platforms) actually do what you want it to do in terms of DDL. 

While I can see the purpose of being able to override a schema in the API, I don't believe such a solution should stand alone as everything else seems to be working out of the box, while the metadata currently always reads from public.

So, what I set out to do is mimicking the functionality that I see in the DDL - which of course is governed by the underlying library or RDBMS itself, as I believe this will lead to the least confusion in terms of consistency between the different parts of the SOCI API.

src/backends/postgresql/session.cpp
cstiborg: So, the PR is currently broken - obviously - and I'm going to fix it.

However, how I should fix it is the big question.

Using the search path is what is used everywhere else in SOCI not because it's explicitly implemented like that, but it's implicitly like that due to the use of the underlying RDBMS.
So for the sake of consistency I would use that, and that will change how the metadata API works in SOCI. (Hopefully to the better)

The question is, how should it actually work? I want the current state to remain, ie. if you don't have a search_path I want to use the public schema - and maybe $user too - this is not how it has been working historically, but maybe the original implementation was not meant for anything but public? - I've been trying to figure that one out.
If $user should be used as well, then I need to implement handling of multiple schemas which is not currently implemented. Currently, looking up tables just returns a table of table names. But maybe the schema name should be there as well? However, that will break backward compatibility...

Going the opposite direction, aiming at keeping backward compatibility only one schema can be used with the metadata. I.e. public without the search path and only the first schema on the search path.

Please let me hear your thought.

vadz: I don't think the existing query can work for Postgres, i.e. it really should be `where table_schema in list-of-schemas-in-the-search-path` instead of just simple equality. Perhaps Postgres has some nicer way of writing this condition, but the idea would be to use the same logic which is used for unqualified table names lookup.

Which brings me to another point: we should support qualified table names too and in this case there shouldn't be any condition involving `table_schema` at all, i.e. I suspect this is broken too currently.

In any case, this definitely needs new tests (even if they're Postgres-specific).

cstiborg: Maybe I'm trying to force a kube into a round hole.

As a user of SOCI I really would prefer if the <code>details::prepare_temp_type</code> returned by <code>session::prepare_table_names()</code> has the same layout for any RDBMS (i.e. not returning both schema and table name for one DB and then only table name for everything else). 

My preferred option would be to have both fields in the output - I think it is cleaner. However, in the hunt for backward compatibility I would suggest the following:

- <code>session::prepare_table_names()</code> returns one field, for PostgreSQL this is a qualified table name. I.e. "schema.table".
- <code>postgresql_session_backend::get_table_names_query()</code> must be implemented to select the "TABLE_NAME" column such that it is created from both the schema_name and table_name and the schema_name must be used in the where clause to select all schemas in the search path, as discussed above.
- <code>details::prepare_temp_type session::prepare_column_descriptions(std::string & table_name)</code> would then always look for a dot and if it exists parse the schema- and table name to give it to the prepared statement as <code>:s</code> and <code>:t</code> respectively. The default statement in <code>session_backend::get_column_description_query()</code> could then make use of this <code>:s</code> in the query, avoiding SQL injection. For <code>table_name</code> input without a schema the default value would simply be "public". Other implementations of <code>get_column_description_query()</code> doesn't have to use the <code>:s</code>. Maintaining backward compatibility.

This solution would allow [schema.]table syntax input to <code>details::prepare_temp_type session::prepare_column_descriptions(std::string & table_name)</code>. However, the tables returned by <code>session::prepare_table_names()</code> will still depend on what is in the search path.

vadz: Sorry, maybe we need to rewind and actually define what problem exactly are we trying to solve here. I thought it was just about getting metadata of the tables in non-default schema but it actually seems something different? If so, could you please define what exactly it is, i.e. what do you need to do that doesn't work right now?

cstiborg: The problem is that the current implementation hardcodes the public schema. This is hardcoded in both get_table_names_query() and get_column_description_query().

This doesn't make sense in PostgreSQL and even more so in MySQL. These are the DBs I'm familiar with - but it could be a problem for other DBs too.

For PostgreSQL I propose to use the search path to get the schemas that should be used for the search of table names.

However, as the same table names may exist in multiple schemas so want to return these in their qualified form and I want to also update the prepare_column_descriptions to be able to get retrieve the metadata for these tables.

cstiborg: > Are you really sure about it? As long as unqualified names are always interpreted in the same way, I don't see any need to qualify them. Although it could be more clear, admittedly.

When having multiple schemas you can end up in this situation (obviously it will only happen ever so often):

```
postgres=# select table_schema || '.' || table_name as "TABLE_NAME" from information_schema.tables where table_schema in ('a', 'b',  'c');
 TABLE_NAME
-------------
 a.soci_test
 b.soci_test
 c.soci_test
(3 rows)
```

Which is the reason why, for the PostgreSQL implementation, I would return the qualified names as in the query suggested above.

cstiborg: In a select statement in PostgreSQL it will work like this, sure.

This write-up mostly serves to convince myself that it can be done. 
To mimic the PostgreSQL behavior in the metadata API, which is indeed sensible, I believe <code>details::prepare_temp_type session::prepare_column_descriptions(std::string & table_name) const</code> and the underlying PostgreSQL can be changed as follows:

Currently <code>details::prepare_temp_type session::prepare_column_descriptions(std::string & table_name) const</code> looks like this:

```
details::prepare_temp_type session::prepare_column_descriptions(std::string & table_name)
{
    ensureConnected(backEnd_);

    return prepare << backEnd_->get_column_descriptions_query(), use(table_name, "t");
}
```

I already had the idea that I would change it to also have a sanitizing function returning a struct of schema and table name like:

```
details::prepare_temp_type session::prepare_column_descriptions(std::string & table_name)
{
    ensureConnected(backEnd_);

    struct table_data = getTableData(table_name);

    return prepare << backEnd_->get_column_descriptions_query(), use(table_data.schema, "s"), use(table_data.name, "t");
}
```

When no schema is present the value of <code>table_data.schema</code> would be <code>NULL</code>. The query for PostgreSQL would then be (Please notice I'm using an example in which I assume the search path to hold $user - being 'soci_user' - and 'public'):

```
select 
    column_name as \"COLUMN_NAME\",
    data_type as \"DATA_TYPE\",
    character_maximum_length as \"CHARACTER_MAXIMUM_LENGTH\",
    numeric_precision as \"NUMERIC_PRECISION\",
    numeric_scale as \"NUMERIC_SCALE\",
    is_nullable as \"IS_NULLABLE\"
from
    information_schema.columns
where
    table_name = :t
    and
    table_schema in ('soci-user', 'public')
    and 
    case
        when :s is not NULL THEN table_schema = :s
        else true
    end
order by
    case table_schema
        when 'soci-user' then 1
        when 'public' then 2
        else 3
    end
limit 1;
```

The interesting blocks which hold values that I can't figure out how to store as :-notation are the in-clause which holds the tables of the search path (I keep them in order but it's irrelevant) which must have the same length as the length of the search path and the order by-clause which also holds the values of the search path and once again must have that length to maintain the order of the result - i.e. the soci-user schema must come before the public schema when no schema is present in :s.

As I wrote in the beginning I'm writing this to convince myself it can be done, however, please let me know if you find anything that seems to be overly complicated.

cstiborg:
I've added all requirements set up in the comments:
- MySQL is now not looking in the public schema for table configuration, instead it looks in the schema for the database.
- PostgreSQL will return table names in <schema>.<table_name> format.
- PostgreSQL will parse the search path coming from the database and use it to look up tables if a table name using the <schema>.<table_name> format isn't used.
- The PostgreSQL test is updated accordingly.

Apparently there is an effect on Oracle and SQLite3 so I will work on that.

cstiborg:
It seems that only macOS tests and Windows tests are failing now (macOS due to lack of funds to run the test and windows due to postgresql missing in the test image)

I believe the rest is good.

include/private/soci-compiler.h
```diff
@@ -36,4 +36,22 @@
 #   define SOCI_GCC_WARNING_RESTORE(x)
 #endif
 
+// SOCI_FALLTHROUGH macro can be used on both g++, mingw, and msvc to generate the
```
vadz, line 39: Sorry, the conditions here don't seem right in several places. Compare with

https://github.com/wxWidgets/wxWidgets/blob/89b38ded388302fd6842f36405dfab27a6c34e9c/include/wx/defs.h#L258-L268

which, I believe, is correct.

include/soci/column-info.h
```diff
@@ -89,11 +88,19 @@ struct type_conversion<column_info>
             ci.type = dt_string;
             ci.dataType = db_string;
         }
-        else if (type_name == "integer" || type_name == "INTEGER")
+        else if (type_name == "integer" || type_name == "INTEGER" ||
+		 type_name == "int" || type_name == "INT" ||
+		 type_name == "bigint" || type_name == "BIGINT")
```
vadz, line 93: Shouldn't this map to `db_int64`?

include/soci/mysql/soci-mysql.h
```diff
@@ -287,6 +287,23 @@ struct mysql_session_backend : details::session_backend
         return "SELECT table_name AS 'TABLE_NAME' FROM information_schema.tables WHERE table_schema = DATABASE()";
     }
 
+    std::string get_column_descriptions_query() const override
+    {
+        return "SELECT column_name as \"COLUMN_NAME\","
```
vadz, line 292: Minor, but I think using raw string literals would make this much more readable, if only by removing the need to escape double quotes inside the string.

include/soci/session.h
```diff
@@ -215,6 +217,8 @@ class SOCI_DECL session
     SOCI_NOT_COPYABLE(session)
 
     void reset_after_move();
+    struct schema_table_name * alloc_schema_table_name(std::string & tableName);
```
vadz, line 220: 2 comments here, one minor but the other one important:

1. `struct` is redundant and should be removed.
2. Is the returned pointer owned by the caller? If so, it should be returned as `std::unique_ptr<schema_table_name>`.

include/soci/session.h
```diff
@@ -232,6 +236,8 @@ class SOCI_DECL session
     bool isFromPool_;
     std::size_t poolPosition_;
     connection_pool * pool_;
+
+    struct schema_table_name * schema_table_name_;
```
vadz, line 240: We should initialize members in their declaration:

```suggestion
    schema_table_name * schema_table_name_ = nullptr;
```

src/backends/postgresql/session.cpp
```diff
@@ -29,6 +30,130 @@ void hard_exec(postgresql_session_backend & session_backend,
...
+    std::vector<std::string> schema_names;
+    PGresult* search_path = PQexec(conn, "SHOW search_path");
+    if (PQresultStatus(search_path) == PGRES_TUPLES_OK)
+    {
+        if (PQntuples(search_path) > 0)
+        {
+            std::string search_path_content = PQgetvalue(search_path, 0, 0);
+
+            bool quoted = false;
+	    std::string schema;
```
vadz, line 62: Weird indent, check for hard TABs?

src/backends/postgresql/session.cpp
```diff
@@ -29,6 +30,130 @@ void hard_exec(postgresql_session_backend & session_backend,
     postgresql_result(session_backend, PQexec(conn, query)).check_for_errors(errMsg);
 }
 
+// helper function to quote a string before sinding to PostgreSQL
```
vadz, line 33: There is a typo here:
```suggestion
// helper function to quote a string before sending to PostgreSQL
```

src/backends/postgresql/session.cpp
```diff
@@ -29,6 +30,130 @@ void hard_exec(postgresql_session_backend & session_backend,
...
+char * quote(PGconn * conn, const char *s, size_t len)
```
vadz, line 34: Why deal with `char*` instead of just returning `std::string`? Using raw pointers is needlessly error-prone.

src/backends/postgresql/session.cpp
```diff
@@ -29,6 +30,130 @@ void hard_exec(postgresql_session_backend & session_backend,
...
+    retv[len_esc + 2] = '\0';
+
+    return retv;
+}
+
+// helper function to collect schemas from search_path
+std::vector<std::string> get_schema_names(PGconn * conn)
+{
+    std::vector<std::string> schema_names;
+    PGresult* search_path = PQexec(conn, "SHOW search_path");
```
vadz, line 54: Please use `details::postgresql_result` to avoid having explicit `PQclear()` calls in the code.

src/core/session.cpp
```diff
@@ -15,6 +15,17 @@
...
 using namespace soci::details;
 
+namespace soci
+{
+    struct schema_table_name
+    {
+        std::string schema;
+        std::string table;
+        soci::indicator ind;
+        struct schema_table_name * next;
```
vadz, line 25:
```suggestion
        schema_table_name * next = nullptr;
```

vadz (changes requested):
Thanks for for the updates!

I've hopefully fixed the CI in #1161 (definitely for macOS, still waiting for AppVeyor), so the next push should run all CI jobs successfully.

But there are some changes needed, notably to make code more obviously safe and correct, i.e. avoid `delete[]` (by using `std::string`), `PGClear()` (by using `postgresql_result` RAII helper) and, I think, also avoid the linked list of schema table name objects: why do we bother with this instead of just using a `std::vector<>` of them?

include/private/soci-compiler.h
cstiborg: I tried to bring over the wxWidgets macros. Just SOCI'fying the macro names like this:

```
#if defined(_MSVC_LANG)
/*
   We want to always use the really supported C++ standard when using MSVC
   recent enough to define _MSVC_LANG, even if /Zc:__cplusplus option is not
   used, but unfortunately we can't just redefine __cplusplus as _MSVC_LANG
   because this is not allowed by the standard and, worse, doesn't work in
   practice (it results in a warning and nothing else).

   So, instead, we define a macro for testing __cplusplus which also works in
   this case.
*/
    #define SOCI_CHECK_CXX_STD(ver) (_MSVC_LANG >= (ver))
#elif defined(__cplusplus)
    #define SOCI_CHECK_CXX_STD(ver) (__cplusplus >= (ver))
#else
    #define SOCI_CHECK_CXX_STD(ver) 0
#endif


#ifndef SOCI_HAS_CLANG_FEATURE
#   ifndef __has_feature
#       define SOCI_HAS_CLANG_FEATURE(x) 0
#   else
#       define SOCI_HAS_CLANG_FEATURE(x) __has_feature(x)
#   endif
#endif

#pragma message(__cplusplus)
#if SOCI_CHECK_CXX_STD(201703L)
#    define SOCI_FALLTHROUGH [[fallthrough]]
#elif defined(__has_warning) && SOCI_HAS_CLANG_FEATURE(cxx_attributes)
#    define SOCI_FALLTHROUGH [[clang::fallthrough]]
#endif

#ifndef SOCI_FALLTHROUGH
    #define SOCI_FALLTHROUGH ((void)0)
#endif
```

However, when I compile it on my platform I get...

```
[ 82%] Building CXX object src/backends/postgresql/CMakeFiles/soci_postgresql_static.dir/session.cpp.o
In file included from /root/soci/src/backends/postgresql/session.cpp:12:
/root/soci/include/private/soci-compiler.h:66:17: warning: expected a string after ‘#pragma message’ [-Wpragmas]
   66 | #pragma message(VAR_NAME_VALUE(__cplusplus))
      |                 ^~~~~~~~~~~~~~
/root/soci/src/backends/postgresql/session.cpp: In function ‘std::vector<std::__cxx11::basic_string<char> > {anonymous}::get_schema_names(PGconn*)’:
/root/soci/src/backends/postgresql/session.cpp:72:21: warning: this statement may fall through [-Wimplicit-fallthrough=]
   72 |                     if (!quoted)
      |                     ^~
/root/soci/src/backends/postgresql/session.cpp:82:17: note: here
   82 |                 default:
      |                 ^~~~~~~
[ 83%] Linking CXX static library ../../../lib/libsoci_postgresql.a
```

This means that I'm not getting `[[fallthrough]]` declared.
Previously I tried to use the BOOST_FALLTHROUGH macro which gave me the same result. So, I'm guessing that there's something funny going on with my platform. Not sure what though.

Are we aiming at using these macro definitions (with a change of the comments) - or should I try to dig further into this?

cstiborg: I just tried building from scratch in a ubuntu docker container. It does the same thing - so not just me.
It must be something with G++. Any ideas?

vadz: What is your compiler and which compilation options (notable `-std=xxx`) do you use?

cstiborg: I'm using g++ on ubuntu. 
`g++ (Ubuntu 13.2.0-23ubuntu4) 13.2.0`

Using default soci install:
```
mkdir soci/build && \
cd soci/build && \
cmake -G "Unix Makefiles" ../ --install-prefix=/usr && \
make
```

So nothing out of the ordinary...

vadz: g++13 uses C++17 by default, so `SOCI_FALLTHROUGH` should be defined as `[[fallthrough]]` for it, please check — by inserting `#error` in the branches that should be taken — if all else fails, that this does happen in the header.

vadz: g++ may support `[[fallthrough]]` in C++14 mode as an extension, I guess, but the real issue is that C++14 shouldn't be used in the first place.

There is

```cmake
if(NOT DEFINED CMAKE_CXX_STANDARD OR CMAKE_CXX_STANDARD LESS 14)
  set(CMAKE_CXX_STANDARD 14)
endif()
```

in SOCI CMakefile, which means that C++14 is used if `CMAKE_CXX_STANDARD` is not set at all. What we need to do is to only set the standard to 14 if the compiler doesn't support it in its default configuration. Unfortunately I don't think CMake allows to do this in any sane way. Here is how [wx does it](https://github.com/wxWidgets/wxWidgets/blob/f4d00c10828c378134ba67de9c473341fa94e174/build/cmake/init.cmake#L11-L52) for comparison.

@Krzmbrzl I wonder if your work on new CMakefiles addresses this already by chance?

Krzmbrzl: > which means that C++14 is used if CMAKE_CXX_STANDARD is not set at all.

or it is set to a value < 14.

> What we need to do is to only set the standard to 14 if the compiler doesn't support it in its default configuration.

Why would we want to do such a thing? Projects should always set the C++ standard they use explicitly. Everything else becomes messy pretty quickly.
Given that you are looking at a feature provided as an extension (enabled when using `std=gnu++14` compared to `std=c++14`) the variable you are looking for is `CMAKE_CXX_EXTENSIONS`.

In my cmake PR I have set this variable to `OFF` in order to force the use of portable code that should work on all compilers.

vadz: > > which means that C++14 is used if CMAKE_CXX_STANDARD is not set at all.
> 
> or it is set to a value < 14.

Yes, but this is as it should be, i.e. not a problem.
 
> > What we need to do is to only set the standard to 14 if the compiler doesn't support it in its default configuration.
> 
> Why would we want to do such a thing? Projects should always set the C++ standard they use explicitly. 

I don't see why would you forbid people from using a later standard, especially if it's the one used by their compiler by default. It also means you can never have any "progressive enhancements", i.e. we're not going to require C++17 just for `[[fallthrough]]` but it's nice to get it when using C++17 compiler anyhow.

I strongly believe that we should never lower the default language version, if anybody really wants to do it, let them set `CMAKE_CXX_STANDARD` to 14 explicitly. This makes more sense by default and is more flexible than the current behaviour. To be honest, I didn't even think that anybody would want to have the current logic, to me it's obviously buggy...

> Everything else becomes messy pretty quickly. Given that you are looking at a feature provided as an extension (enabled when using `std=gnu++14` compared to `std=c++14`) the variable you are looking for is `CMAKE_CXX_EXTENSIONS`.
> 
> In my cmake PR I have set this variable to `OFF` in order to force the use of portable code that should work on all compilers.

The question of `-std=gnuXX` vs `-std=c++XX` is something else and isn't really related to the question of `XX` to use or whether to use any language dialect options at all.

Krzmbrzl: If you use the compiler default then you essentially encourage introduction of non-portable code in the sense that the project specifies that it is compatible with a given C++ standard but since someone uses a compiler that defaults to a more recent version they can code up something that won't compile for someone else who in fact has to use the standard the project specifies. 
By being explicit about the used standard version you prevent this as everyone is using the same standard version. 

And with the construct you see in the current cmake files, you ensure that when SOCI is embedded as a subproject, the parent project can use a more recent C++ standard without being held back by the embedded project.

So as you can see we have completely opposite views on this. For me, not having a fixed standard version is an absolute no go.

vadz: The crucial point is that you can specify the fixed C++ version easily on CMake command line. You can't specify "use my compiler default" easily or, maybe, even at all.

And using the default is much friendlier for non-experts who are much less likely to know how to change this.

Krzmbrzl: I disagree. Non-experts should not fiddle with the C++ standard of a given project. 
And I remain confident with the statement that "use whatever standard the compiler happens to default to" is not really good.

Also, I think if you simply don't specify the standard variable at all, you probably end up with the compiler's default.

vadz: > Also, I think if you simply don't specify the standard variable at all, you probably end up with the compiler's default.

This is what this thread started with: if you don't set anything, SOCI sets the standard to C++14. And I think it shouldn't, if the compiler supports something later by default.

Krzmbrzl: I was talking about cmake.

And I think SOCI does the right thing by setting a default standard version. I have not yet encountered any halfway serious project that doesn't specify the used standard version.

src/backends/postgresql/session.cpp
cstiborg: I used the same approach (function declaration) as for MySQL. I will change it.

include/soci/mysql/soci-mysql.h
cstiborg: I assume this is a general comment as they appear in a couple of places

include/soci/column-info.h
cstiborg: I've mapped this correctly on my side - I'll commit it in the next few days.

include/soci/session.h
cstiborg: The schema_table_name is only on loan. It is stored in the linked list commented on elsewhere as it must be stored for the lifetime of the prepare_temp_type object - so deleting it when exiting the scope of the method will break the intended functionality.

include/private/soci-compiler.h
cstiborg: I'm going to extend the fallthrough definition to include a check for GNU C++ extensions. (The last elif in the following)

```
#if SOCI_CHECK_CXX_STD(201703L)
#    define SOCI_FALLTHROUGH [[fallthrough]]
#elif defined(__has_warning) && SOCI_HAS_CLANG_FEATURE(cxx_attributes)
#    define SOCI_FALLTHROUGH [[clang::fallthrough]]
#elif defined(__GNUC__) && __has_cpp_attribute(fallthrough)
#    define SOCI_FALLTHROUGH [[fallthrough]]
#endif
```

Then it'll also work (as in not complain) in the current scenario.

include/soci/session.h
vadz: Please add a comment saying that this pointer is owned by the class and shouldn't be deleted then, as this is rather non-obvious.

Alternatively, and even better, if the returned pointer is always non-null, we could return a reference here instead.

include/private/soci-compiler.h
```diff
@@ -36,4 +36,40 @@
...
+#   define SOCI_HAS_CLANG_FEATURE(x) 0
+#else
+#   define SOCI_HAS_CLANG_FEATURE(x) __has_feature(x)
+#endif
+
+// FALLTHROUGH macro defines a cross-platform/version to mark fallthrough
+// behavior in switch statement.
+#if SOCI_CHECK_CXX_STD(201703L)
+#    define SOCI_FALLTHROUGH [[fallthrough]]
+#elif defined(__has_warning) && SOCI_HAS_CLANG_FEATURE(cxx_attributes)
```
vadz, line 61: Is `__has_warning` check really needed? I.e. can't we just do

```suggestion
#elif SOCI_HAS_CLANG_FEATURE(cxx_attributes)
```

include/soci/column-info.h
```diff
@@ -11,7 +11,6 @@
 #include "soci/soci-backend.h"
 #include "soci/type-conversion.h"
 #include "soci/values.h"
-
```
vadz, line 14: Minor, but it's better to try to avoid making whitespace-only changes in the commits with significant changes.

src/backends/postgresql/session.cpp
```diff
@@ -29,6 +30,126 @@ void hard_exec(postgresql_session_backend & session_backend,
     postgresql_result(session_backend, PQexec(conn, query)).check_for_errors(errMsg);
 }
 
+// helper function to quote a string before sending to PostgreSQL
+std::string quote(PGconn * conn, std::string& s)
+{
+    int error_code;
+    char *retv = new char[2 * s.length() + 1];
+    size_t len_esc = PQescapeStringConn(conn, retv, s.c_str(), s.length(), &error_code);
```
vadz, line 38: Sorry, when I recommended using `std::string` I meant to do it _instead_ of manually allocated buffer, not in addition to it. I.e. could we please replace this with

```suggestion
    std::string retv;
    retv.resize(2 * s.length() + 3);
    retv[0] = '\'';
    size_t len_esc = PQescapeStringConn(conn, retv.data() + 1, s.c_str(), s.length(), &error_code);
```

and remove `delete []` below?

cstiborg:
> ... and, I think, also avoid the linked list of schema table name objects: why do we bother with this instead of just using a `std::vector<>` of them?

The point of not using std::vector, is that as it fills up it may move memory around on the heap. The reason for having the data allocated on the heap in the first place is that the prepare_temp_type created to fetch the column data and table data uses the memory addresses of the parameters of the prepared statement, thus these cannot change for the lifetime of the prepare_temp_type object.
I've changed it to a std::forward_list.

include/soci/session.h
```diff
@@ -215,6 +218,7 @@ class SOCI_DECL session
     SOCI_NOT_COPYABLE(session)
 
     void reset_after_move();
+    schema_table_name * alloc_schema_table_name(std::string & tableName);
```
vadz, line 221: It doesn't look like the table name is modified by this function, so I think this should be

```suggestion
    schema_table_name * alloc_schema_table_name(const std::string& tableName);
```

include/soci/session.h
cstiborg: References are so much better - it's coming in a few minutes - when I've tested it locally.

src/core/session.cpp
```diff
@@ -207,6 +217,33 @@ void session::reset_after_move()
     pool_ = nullptr;
 }
 
+schema_table_name* session::alloc_schema_table_name(std::string & tableName)
+{
+    std::size_t dot_pos = tableName.find( '.' );
```
vadz, line 222: Out of curiosity, what happens if there is more than one period in the string?

Also, what if it's in the first or last position?

Should we validate it here somehow, maybe?

tests/postgresql/test-postgresql.cpp
```diff
@@ -689,6 +689,23 @@ TEST_CASE("PostgreSQL ORM cast", "[postgresql][orm]")
     sql << "select :a::int", use(v); // Must not throw an exception!
 }
 
+bool table_name_equals_ignore_schema(const std::string& table_name_with_schema, const std::string& table_name)
```
vadz, line 692: It's a bit surprising to have such a function rather than a simple(r) `get_table_name_without_schema()` and then continuing to use simple string comparison below.

src/backends/postgresql/session.cpp
```diff
@@ -29,6 +30,126 @@ void hard_exec(postgresql_session_backend & session_backend,
...
+    if (search_path_result.check_for_data("search_path doesn't exist"))
+    {
+        if (PQntuples(search_path_result) > 0)
+        {
+            std::string search_path_content = PQgetvalue(search_path_result, 0, 0);
+
+            bool quoted = false;
+            std::string schema;
+            while (!search_path_content.empty())
+           {
```
vadz, line 67: Minor, but the indent is still wrong here.

src/backends/postgresql/session.cpp
```diff
@@ -29,6 +30,126 @@ void hard_exec(postgresql_session_backend & session_backend,
...
+                    if (current_user_result.check_for_data("current_user is not defined"))
+                    {
+                        if (PQntuples(current_user_result) > 0)
+                        {
+                            schema_name = PQgetvalue(current_user_result, 0, 0);
+                        }
+                    }
+                }
+
+                // Assure no bad characters
```
vadz, line 106:
```suggestion
                // Ensure no bad characters
```

src/backends/postgresql/session.cpp
```diff
@@ -29,6 +30,126 @@ void hard_exec(postgresql_session_backend & session_backend,
...
+            }
+        }
+        schema_names.push_back("public");
+    }
+
+    return schema_names;
+}
+
+// helper function to create a comma separated list of strings
+std::string create_list_of_strings(const std::vector<std::string>& list)
```
vadz, line 131: Also minor, but it's rather confusing to call a `vector` a "list", especially when there is a (forward) list in this code too.

src/backends/postgresql/session.cpp
```diff
@@ -29,6 +30,126 @@ void hard_exec(postgresql_session_backend & session_backend,
...
+{
+    std::ostringstream oss;
+    for (size_t i = 0; i < list.size(); ++i) {
+        if (i != 0) {
+            oss << ", ";
+        }
+        oss << list[i];
+    }
```
vadz, line 139: I'd still use a range for here:

```suggestion
    bool first = true;
    for ( const auto& s: list )
    {
        if ( first )
            first = false;
        else
            oss << ", ";

        oss << s;
    }
```

vadz (commented):
Thanks for the updates, changes look broadly good but could we please have some new tests showing how can this actually be used?

And updating the docs would be useful too.

TIA!

vadz:
> The point of not using std::vector, is that as it fills up it may move memory around on the heap.

Thanks, I've realized this too now, while rereading the code and using `std::forward_list` is fine, but it would be worth adding a comment saying that we use the list because we need the pointers/references to remain stable as this might not be obvious (as it wasn't do me).

src/core/session.cpp
cstiborg: The table name will be treated as if it contains a ".". I was thinking about it, and came to the conclusion that - well - if you try to look up a table name with a dot - which typically isn't allowed in SQL table names. (I don't know of anything allowing it) - It simply won't find anything... So, garbage in, garbage out. It surely shouldn't break anything.

src/backends/postgresql/session.cpp
cstiborg: Would `strings` be okay? (It's always the pressure of a reasonable name)

vadz: Yes, this, or `names`, or `tables` (if this is correct?).

And yes, naming is hard, as has been long recognized by the authorities in the field:

> There are 2 hard problems in computer science: cache invalidation, naming things, and off-by-1 errors.

src/backends/postgresql/session.cpp
cstiborg: Thank you - I was too much caught up in bit keeping... (I had to const_cast retv.data() but it seems good otherwise)

tests/postgresql/test-postgresql.cpp
cstiborg: I've changed this

src/backends/postgresql/session.cpp
```diff
@@ -29,6 +30,128 @@ void hard_exec(postgresql_session_backend & session_backend,
...
+// helper function to collect schemas from search_path
+std::vector<std::string> get_schema_names(postgresql_session_backend & session, PGconn * conn)
+{
+    std::vector<std::string> schema_names;
+    postgresql_result search_path_result(session, PQexec(conn, "SHOW search_path"));
+    if (search_path_result.check_for_data("search_path doesn't exist"))
+    {
+        if (PQntuples(search_path_result) > 0)
+        {
+            std::string search_path_content = PQgetvalue(search_path_result, 0, 0);
```
vadz, line 60: Sorry, I hadn't realized it on first reading, but I think this code could be simplified like this:

```suggestion
        std::string search_path_content;
        if (PQntuples(search_path_result) > 0)
            search_path_content = PQgetvalue(search_path_result, 0, 0);
        if (search_path_content.empty())
            search_path_content = R"("$user", public)"; // fall back to default value
```

as then we could get rid of the `if` below.

src/backends/postgresql/session.cpp
```diff
@@ -29,6 +30,128 @@ void hard_exec(postgresql_session_backend & session_backend,
...
+            if (PQntuples(current_user_result) > 0)
+            {
+                std::string user = PQgetvalue(current_user_result, 0, 0);
+
+                // Ensure no bad characters
+                schema_names.push_back(quote(conn, user));
+            }
+        }
+        schema_names.push_back("public");
+    }
```
vadz, line 123: I.e. this duplicated code wouldn't be needed if we just set `search_path_content` to the default value if we couldn't get it from the database.

vadz (commented):
Thanks, looks good now, except for one simple refactoring that I'd like to do just to avoid duplicating the code dealing with the current user name.

To answer your previous comment: I think nobody cares about DB/2 (or, if somebody still does, I haven't encountered them yet) and Firebird and SQLite don't support schemas at all (they use multiple databases instead, but this is not the same thing), so I am not sure how to interpret "highly likely not doing anything constructive" — do you mean that this doesn't improve anything for, e.g., FB? If so, this is true, of course, but I don't see what else could be done for it, as it doesn't support schemas anyhow.

ODBC _could_ be improved because we can know which database we're connected to, see `odbc_session_backend::get_database_product()` and it could indeed be nice to have schema support for SQL Server (which can only be used via this backend) and PostgreSQL (which sometimes happens to be used via ODBC rather than natively due to whatever reasons).

Please let me know if you'd like to add support for this to ODBC backend too (this would require some refactoring) or if we should merge this as is.

Thanks again!

cstiborg:
> "highly likely not doing anything constructive"

For Firebird this means it is the default code - as it has been since the original implementation. I know nothing about Firebird, except that it doesn't use schemas, as you just taught me. However, without schemas, and the default implementation looking in `information_schema.tables` and `information_schema.columns` I am almost certain that it will simply fail when using the soci metadata api. I haven't tested it, so I don't know for sure, but it doesn't add up for me. That's what I meant.

There is a specialisation for SQLite3 - still hasn't tested it so I'm just talking - which I assume will work as it collects table information from the `sqlite_master` table and some SQLite3 voodoo (`pragma_table_info(<table_name>)`) for the columns. Again, I don't know anything about Firebird, but possibly something like that could be done to make the metadata API work across the board. If there's no way to get the data, then I guess the error produced by FB is as good as any error.

> Please let me know if you'd like to add support for this to ODBC backend too (this would require some refactoring) or if we should merge this as is.

Tricky question. A couple of things: I am working on SOCI because I use it for another project. That project currently supports MySQL and PostgreSQL. Which means I have a setup to test that. I have a requirement to support MSSQL as well, however, I'm not quite there yet. So, I would like to make ODBC work, particularly for MSSQL. But I don't have a test setup yet.

All of this makes me think, that I personally, would appreciate to have the current PR merged, and I will then work on ODBC (MSSQL and PostgreSQL) when I'm there on my other project.

Issue #1151, opened 2024-06-04
Query not working through freeTDS works in Microsoft ODBC Driver 


kowsikbabu:
Before posting here, I have verified this is not a driver side issue, because the query works in both FreeTDS & Microsoft ODBC Driver using native SQL code. 

When we use Microsoft ODBC driver via SOCI, we get the results as expected. 3 rows with several column values
But when we use FreeTDS via SOCI, we get the same number of rows returned, but the rows are returned empty.

We have tried to debug the issue for more than a week, but we are stuck. Kindly help us resolve this issue. 

We have tried with both 4.0.3 & master & both (64 & 32) versions as well. SQL server version is 2019.

Query for reference: 

 `SET NOCOUNT ON;DECLARE db_cursor CURSOR FOR SELECT name FROM sys.databases WHERE state_desc = 'ONLINE' AND name IN ('TestDB');DECLARE @DatabaseName NVARCHAR(128);DECLARE @outset TABLE(INSTANCENAME varchar(50),DATABASENAME varchar(100),TABLENAME varchar(100),NUMBEROFRECORDS_I bigint);OPEN db_cursor;FETCH NEXT FROM db_cursor INTO @DatabaseName;WHILE @@FETCH_STATUS = 0 BEGIN DECLARE @command nvarchar(1000) = 'USE '+ QUOTENAME(@DatabaseName) +'; SELECT @@SERVERNAME,DB_NAME(),T.NAME,P.[ROWS] FROM sys.tables T INNER JOIN sys.indexes I ON T.OBJECT_ID = I.OBJECT_ID INNER JOIN sys.partitions P ON I.OBJECT_ID = P.OBJECT_ID AND I.INDEX_ID = P.INDEX_ID INNER JOIN sys.allocation_units A ON P.PARTITION_ID = A.CONTAINER_ID WHERE T.NAME NOT LIKE ''DT%'' AND I.OBJECT_ID > 255 AND I.INDEX_ID <= 1 GROUP BY T.NAME, I.OBJECT_ID, I.INDEX_ID, I.NAME, P.[ROWS] ORDER BY OBJECT_NAME(I.OBJECT_ID)' INSERT INTO @outset EXEC (@command) FETCH NEXT FROM db_cursor INTO @DatabaseName END CLOSE db_cursor DEALLOCATE db_cursor select INSTANCENAME,DATABASENAME,TABLENAME,NUMBEROFRECORDS_I from @outset;`
 
[...]

PR #1157, opened 2024-07-23
core/statement.cpp : Modified to support complex queries

JoshuaVThomas:
Added a part where describe is called again if numcols doesn't get populated before execution of query.

Ref : https://github.com/SOCI/soci/issues/1151

### DIFF
--- a/src/core/statement.cpp
+++ b/src/core/statement.cpp
@@ -318,6 +318,12 @@ bool statement_impl::execute(bool withDataExchange)
 
         statement_backend::exec_fetch_result res = backEnd_->execute(num);
 
+        // calling describe again in-case it is a complex query and numcols is not populated before execution
+        if (row_ != NULL && alreadyDescribed_ == false)
+        {
+            describe();
+        }
+
         bool gotData = false;
 
         if (res == statement_backend::ef_success)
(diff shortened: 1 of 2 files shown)


vadz:
Thanks, but it would be possible to add a test case that fails now but would work with this change? Otherwise this risks getting broken again in the future.

vadz:
Sorry, I really can't merge without understanding what exactly this fixes.

JoshuaVThomas:
Hi @vadz I've added the test case where this failed but succeeds now.

vadz:
Thank you, but I'll need some help with understanding what exactly happens with this query... Any extra explanations would be welcome!

JoshuaVThomas:
This is a sample query which has a CURSOR declared. It's aim is to execute a complex query at ones instead of set by set operation. Since it's a complex query it takes some time to execute so that the numcols doesn't get set. So the code has been modified to address this.

vadz:
Sorry for the delay, I've finally looked at it and I couldn't find any better way to fix this than to apply _almost_ this change, please see #1182 which I've created to replace this one.

The main difference is that I don't understand how did it work for you, and even in the CI build, without calling `define_for_row()` after the 2nd `describe()`. In my local testing this just crashed because `intosForRow_` elements were never initialized and, again, I just don't see how could it work without it.

Anyhow, if you don't have any objections to calling `define_for_row()` too, I'll merge the other PR soon.

Thanks again for the fix!

PR #1164, opened 2024-08-25
Implement session::get_last_query_context()

### LINKED ISSUE #678: How to get actual query of statement with bound input values
```cpp
string first_name = "Steve";
string last_name = "Jobs";
sql << "insert into test(first_name, last_name)"
    " values(:first_name, :last_name)",
    use(first_name), use(last_name);
```

when I use `sql.get_query()`， the result is `insert into test(first_name, last_name) values(:first_name, :last_name)`

but what I expected is `insert into test(first_name, last_name) values('Steve', 'Jobs')`

### LINKED ISSUE #1027: Create `session::get_last_query_with_context`
We already have the ability of obtaining the last executed query from a given `session` object. This is great, but in certain cases it would be beneficial to also be able to get a hold of the bound parameters for that query - essentially exactly is is reported if SOCI throws an exception itself.

As I see it, this should be possible by leveraging
https://github.com/SOCI/soci/blob/80d4122cf05c8c9fc8fe3f3356c19e7c3b418338/src/core/statement.cpp#L786-L821
This could be factored out into its own function and then the `session` class can access this function to expose the query with the bound parameters to the user via a public API.

My use case for this is that I perform some post-checking of whether the given query resulted in the expected data and if not, I want to throw an exception myself. In order to give it some context, I include the last executed query, but it would be a lot more useful if I could also include the parameters for that query.

Related:
- #678
- https://github.com/SOCI/soci/commit/8bb66107a2132d424f441305d04ee71a39798f2b

Krzmbrzl:
Fixes #678
Fixes #1027

### DIFF
--- a/include/soci/logger.h
+++ b/include/soci/logger.h
@@ -11,9 +11,20 @@
 #include "soci/soci-platform.h"
 
 #include <ostream>
+#include <vector>
 
 namespace soci
 {
+
+
+struct SOCI_DECL query_parameter
+{
+    explicit query_parameter(std::string name = {}, std::string value = {})
+        : name(std::move(name)), value(std::move(value)) {}
+
+    std::string name;
+    std::string value;
+};
 
 // Allows to customize the logging of database operations performed by SOCI.
 //
@@ -28,12 +39,13 @@ public:
     virtual ~logger_impl();
 
     // Called to indicate that a new query is about to be executed.
-    virtual void start_query(std::string const & query) = 0;
+    virtual void start_query(std::string const & query);
 
     // Called to log a parameter that is bound to the currently active query
-    virtual void add_query_parameter(std::string name, std::string value) = 0;
+    virtual void add_query_parameter(std::string name, std::string value);
 
-    virtual void reset_query_parameter() = 0;
+    // Clears all currently logged query parameters
+    virtual void clear_query_parameters();
 
     logger_impl * clone() const;
 
@@ -43,7 +55,10 @@ public:
     virtual void set_stream(std::ostream * s);
     virtual std::ostream * get_stream() const;
     virtual std::string get_last_query() const;
-    virtual std::string get_last_query_with_context() const;
+    virtual std::string get_last_query_context() const;
+
+protected:
+    std::vector<query_parameter> queryParams_;
 
 private:
     // Override to return a new heap-allocated copy of this object.
@@ -79,13 +94,13 @@ public:
         m_impl->add_query_parameter(std::move(name), std::move(value));
     }
 
-    virtual void reset_query_parameter() { m_impl->reset_query_parameter(); }
+    virtual void clear_query_parameters() { m_impl->clear_query_parameters(); }
 
     // Methods used for the implementation of session basic logging support.
     void set_stream(std::ostream * s) { m_impl->set_stream(s); }
     std::ostream * get_stream() const { return m_impl->get_stream(); }
     std::string get_last_query() const { return m_impl->get_last_query(); }
-    std::string get_last_query_with_context() const { return m_impl->get_last_query_with_context(); }
+    std::string get_last_query_context() const { return m_impl->get_last_query_context(); }
 
 private:
     logger_impl * m_impl;
--- a/include/soci/logger.h
+++ b/include/soci/logger.h
@@ -30,6 +30,11 @@ public:
     // Called to indicate that a new query is about to be executed.
     virtual void start_query(std::string const & query) = 0;
 
+    // Called to log a parameter that is bound to the currently active query
+    virtual void add_query_parameter(std::string name, std::string value) = 0;
+
+    virtual void reset_query_parameter() = 0;
+
     logger_impl * clone() const;
 
     // These methods are for compatibility only as they're used to implement
@@ -38,6 +43,7 @@ public:
     virtual void set_stream(std::ostream * s);
     virtual std::ostream * get_stream() const;
     virtual std::string get_last_query() const;
+    virtual std::string get_last_query_with_context() const;
 
 private:
     // Override to return a new heap-allocated copy of this object.
@@ -68,10 +74,18 @@ public:
 
     void start_query(std::string const & query) { m_impl->start_query(query); }
 
+    virtual void add_query_parameter(std::string name, std::string value)
+    {
+        m_impl->add_query_parameter(std::move(name), std::move(value));
+    }
+
+    virtual void reset_query_parameter() { m_impl->reset_query_parameter(); }
+
     // Methods used for the implementation of session basic logging support.
     void set_stream(std::ostream * s) { m_impl->set_stream(s); }
     std::ostream * get_stream() const { return m_impl->get_stream(); }
     std::string get_last_query() const { return m_impl->get_last_query(); }
+    std::string get_last_query_with_context() const { return m_impl->get_last_query_with_context(); }
 
 private:
     logger_impl * m_impl;
--- a/src/core/session.cpp
+++ b/src/core/session.cpp
@@ -11,6 +11,7 @@
 #include "soci/connection-pool.h"
 #include "soci/soci-backend.h"
 #include "soci/query_transformation.h"
+#include "soci/log-context.h"
 
 using namespace soci;
 using namespace soci::details;
@@ -465,6 +466,16 @@ std::ostream * session::get_log_stream() const
     {
         return logger_.get_stream();
     }
+}
+
+void session::set_query_context_logging_mode(log_context ctx)
+{
+    query_ctx_logging_mode_ = ctx;
+}
+
+log_context session::get_query_context_logging_mode() const
+{
+    return query_ctx_logging_mode_;
 }
 
 void session::log_query(std::string const & query)
--- a/src/core/session.cpp
+++ b/src/core/session.cpp
@@ -36,6 +36,15 @@ void ensureConnected(session_backend * backEnd)
     }
 }
 
+struct query_parameter
+{
+    query_parameter(std::string name = {}, std::string value = {})
+        : name(std::move(name)), value(std::move(value)) {}
+
+    std::string name;
+    std::string value;
+};
+
 // Standard logger class used by default.
 class standard_logger_impl : public logger_impl
 {
@@ -53,7 +62,15 @@ public:
         }
 
         lastQuery_ = query;
+        reset_query_parameter();
     }
+
+    virtual void add_query_parameter(std::string name, std::string value)
+    {
+        queryParams_.emplace_back(std::move(name), std::move(value));
+    }
+
+    virtual void reset_query_parameter() { queryParams_.clear(); }
 
     virtual void set_stream(std::ostream * s)
     {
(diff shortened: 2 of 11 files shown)


Krzmbrzl:
@vadz do you think we need a way to disable parameter logging? Compared to the overhead of the actual DB query the logging of parameters is likely insignificant but maybe you have a scenario in mind in which this might become important?

vadz:
> @vadz do you think we need a way to disable parameter logging? Compared to the overhead of the actual DB query the logging of parameters is likely insignificant but maybe you have a scenario in mind in which this might become important?

The only 2 potential problems I can think about are:

1. Logging sensitive information, e.g. if you store passwords or credit card numbers in the database.
2. Logging too much data and flooding the logs with useless information.

But it looks like both could be addressed by overriding `add_query_parameter()`?

Krzmbrzl:
> But it looks like both could be addressed by overriding add_query_parameter()?

Yes.

Also, parameters are not logged persistently. They only live to the next query execution. That is still not good for sensitive information of course. Just wanted to clarify.

Krzmbrzl:
Just to be explicit about it: from my POV this PR is done

src/core/session.cpp
```diff
@@ -26,6 +26,15 @@ void ensureConnected(session_backend * backEnd)
     }
 }
 
+struct query_parameter
+{
+    query_parameter(std::string name = {}, std::string value = {})
```
vadz, line 31: Minor, but

```suggestion
    explicit query_parameter(std::string name = {}, std::string value = {})
```

src/core/session.cpp
```diff
@@ -26,6 +26,15 @@ void ensureConnected(session_backend * backEnd)
...
+        : name(std::move(name)), value(std::move(value)) {}
+
+    std::string name;
+    std::string value;
```
vadz, line 35: Not that important, but I'd make this change

```suggestion
    const std::string name;
    const std::string value;
```

to be explicit about the fact that they never change.

include/soci/logger.h
```diff
@@ -30,6 +30,11 @@ class SOCI_DECL logger_impl
     // Called to indicate that a new query is about to be executed.
     virtual void start_query(std::string const & query) = 0;
 
+    // Called to log a parameter that is bound to the currently active query
+    virtual void add_query_parameter(std::string name, std::string value) = 0;
+
+    virtual void reset_query_parameter() = 0;
```
vadz, line 36: As this resets _all_ parameters and not just one of them, I'd like to rename it to

```suggestion
    virtual void reset_query_parameters() = 0;
```

src/core/session.cpp
```diff
@@ -60,6 +77,31 @@ class standard_logger_impl : public logger_impl
         return lastQuery_;
     }
 
+    virtual std::string get_last_query_with_context() const
+    {
+        if (queryParams_.empty()) {
```
vadz, line 82: Minor but this doesn't conform to SOCI coding style:

```suggestion
        if (queryParams_.empty())
        {
```

src/core/session.cpp
```diff
@@ -60,6 +77,31 @@ class standard_logger_impl : public logger_impl
...
+            return get_last_query();
+        }
+
+        std::string query = get_last_query();
+
+        query += " with ";
+
+        for (std::size_t i = 0; i < queryParams_.size(); ++i)
```
vadz, line 90: I'd suggest still using the ranged for loop, it's just much more clear:

```suggestion
        bool first = true;
        for (const auto& param: queryParams_)
```

and then in the loop body

```cpp
if (first)
    first = false;
else
    query += ", ";
```

tests/common-tests.h
```diff
@@ -3640,16 +3640,45 @@ TEST_CASE_METHOD(common_tests, "Basic logging support", "[core][logging]")
...
-    catch (...) {}
+        auto_table_creator tableCreator(tc_.table_creator_1(sql));
 
-    CHECK(sql.get_last_query() == "drop table soci_test2");
+        int id = 1;
+        std::string name = "b";
+        sql << "insert into soci_test (name,id) values (:name,:id)", use(name, "name"), use(id, "id");
+
+        CHECK(sql.get_last_query() == "insert into soci_test (name,id) values (:name,:id)");
+        CHECK(sql.get_last_query_with_context() == "insert into soci_test (name,id) values (:name,:id) with :name=\"b\", :id=1");
```
vadz, line 3655: Pretty minor too, but I think using raw string literals is more readable (and allows copying and pasting easily, if necessary) than escaping quotes, i.e.
```suggestion
        CHECK(sql.get_last_query_with_context() == R"(insert into soci_test (name,id) values (:name,:id) with :name="b", :id=1)");
```

include/soci/logger.h
```diff
@@ -30,6 +30,11 @@ class SOCI_DECL logger_impl
     // Called to indicate that a new query is about to be executed.
     virtual void start_query(std::string const & query) = 0;
 
+    // Called to log a parameter that is bound to the currently active query
+    virtual void add_query_parameter(std::string name, std::string value) = 0;
```
vadz, line 34: I wonder if this really should be pure virtual. Both not implementing it at all (i.e. leaving it empty by default) and just storing the values (as standard logger does by default) look reasonable default implementations.

It's also somewhat inconsistent with `get_last_query_with_context()` not being pure virtual.

include/soci/logger.h
```diff
@@ -38,6 +43,7 @@ class SOCI_DECL logger_impl
     virtual void set_stream(std::ostream * s);
     virtual std::ostream * get_stream() const;
     virtual std::string get_last_query() const;
+    virtual std::string get_last_query_with_context() const;
```
vadz, line 46: I'd call this `get_last_query_context()` and return the string with the parameter values from it (and empty string by default) and append it to the query in `get_last_query()` if it's non-empty.

This doesn't really change much, of course, but it just looks more logical to have a function doing A and a function doing B, with A using B, instead of having a function doing A and a function doing A+B.

vadz (commented):
Thanks for your work on this, but after rereading it carefully, I've realized that, in addition to various small remarks below, I have a more global questions: with this change we now impose some extra overhead for all queries, even if it's not needed at all. Granted, this overhead is not huge, but I'm not sure it's completely negligible neither, considering that there may be a lot of parameters in a query (I'm sure I have some with a few dozens in my own code). Have you tried to measure it by chance?

I don't know if we can avoid this by only returning the query context on demand, i.e. only when/if `logger::get_last_query_context()` is called, but it looks like it should be possible, or am I missing some reason it isn't?

In any case, after we decide on their final shape, the new APIs need to be mentioned in at least `docs/api/client.md` and `docs/logging.md` (we probably need to have a new section about logging parameter values in the latter).

Thanks again!

Krzmbrzl:
> I don't know if we can avoid this by only returning the query context on demand, i.e. only when/if logger::get_last_query_context() is called, but it looks like it should be possible, or am I missing some reason it isn't?

I don't think we have access to the last executed statement, do we? And without access to that, how do you want to obtain the last used parameters?
Always saving them on use seemed like the only way to pull this off, but if I missed something, I'm open to suggestions. 

> Granted, this overhead is not huge, but I'm not sure it's completely negligible neither, considering that there may be a lot of parameters in a query (I'm sure I have some with a few dozens in my own code). Have you tried to measure it by chance?

I didn't measure it. My gut tells me that any DB call will take significantly longer than logging a tens, even hundreds of parameters but that's just guessing. 

I way thinking of whether there might be a semantic such as
```
sql.enable_query_context_logging(true);
```
to have a way to explicitly toggle the behavior to opt in or out of this feature. 
If we have that, we only have to decide on the default being on or off.

src/core/session.cpp
Krzmbrzl: I think here is the wrong place to do this. This should be encoded in the place that actually uses an instance of this struct by making that specific instance `const`.

Additionally, having const members can have unintended side-effects like a disabled (by default) assignment operator. That's why I usually don't like having `const` members :thinking:

include/soci/logger.h
Krzmbrzl: I'll make storing the parameters the default :+1:

include/soci/logger.h
Krzmbrzl: Sounds good

Krzmbrzl:
@vadz what do you think of the above suggestion?

vadz:
> @vadz what do you think of the above suggestion?

Sorry for the delay, I still didn't have time to look at this properly, but maybe you can answer my question without spending any time on it:

Could we store a pointer to the statement being executed in the session? Or do we need to get the query context after the statement has been already destroyed?

Krzmbrzl:
> do we need to get the query context after the statement has been already destroyed?

I think we do need this functionality. Imo the context is most useful when young some sort of error handling (for the error message) and at this point the executed statement is usually destroyed already (as to my understanding the statement only lives until it has been executed).

vadz:
But couldn't we fill the context in `rethrow_current_exception_with_context()`, just as we do now?

Krzmbrzl:
I don't think so. When we use the rethrow, we are still within the statement (which is therefore still alive). However, we need to be able to access the context even when the statement has already been destroyed (as we want to be able to access the context whenever we access the last query, which is independent of the statement object).

And also, we don't want to have the context available only when there is an error while executing the statement.

vadz:
I thought we could fill the context stored in the session, but do it only when an exception happens. What prevents us from doing it?

This doesn't help if we want to have the context even without the exception, of course, but IME this is mostly useful when an error occurs, logging successful executions doesn't seem very interesting...

But, in principle, we could also pass some

```cpp
enum class LogContext
{
  Never,
  Always,
  OnError
};
```

to `enable_query_context_logging()` to make this maximally configurable, with `OnError` only doing it in `rethrow_current_exception_with_context()`.

Krzmbrzl:
>  successful executions doesn't seem very interesting...

I disagree. There are soft errors that arise from the semantics which happen even though the query was successful. In fact, this is the primary use case for me as the exception message in case of a DB error already contains the parameters. 
In my case the example is the execution of a query which I expect to produce data. So if the query produces the empty set, that is an error that I want to report even though the query itself ran fine and didn't throw an exception. 

> But, in principle, we could also pass some
>
> enum class LogContext { Never, Always, OnError };
>
> to enable_query_context_logging() to make this maximally configurable, with OnError only doing it in rethrow_current_exception_with_context()

Works for me :+1:

vadz:
Would you be willing to implement this `LogContext` proposal? TIA!

Krzmbrzl:
Yes absolutely! I just didn't find the time to do so yet.

vadz:
> Yes absolutely! I just didn't find the time to do so yet.

No problem at all, I just wanted to check if you were waiting for something from me.

Krzmbrzl:
Note that when using `sql.set_query_context_logging_mode(log_context::none)`, not only are parameters not cached with the new mechanism. They are then also excluded from exception messages. My line of thought has been that if the user believes that their query parameters shouldn't be cached even in case of a DB query error, then they probably don't want these parameters as part of an exception message either (as most likely the parameters contain sensitive information that should not leak).

vadz:
Thanks, this looks good to me, I've added a few minor changes and if you don't have any objections to them, I'll (squash) merge this soon.

One unfortunate thing is that even this doesn't help when using vector parameters, as their data still doesn't appear in the output and even if it did appear, it wouldn't be helpful to just dump a 1000 element vector without knowing which of its elements resulted in the problem and I don't know how to do it for all backends (notably ODBC). If you, or anybody else, have any ideas about how to improve this, they would be very welcome!

Krzmbrzl:
> I've added a few minor changes and if you don't have any objections to them, I'll (squash) merge this soon.

Sure, LGTM.

> One unfortunate thing is that even this doesn't help when using vector parameters

Ah, right. I completely forgot about those. Given that they (I think?) work by effectively sending all data at once to the DB in order to perform a mass query, I kinda tend to think that it is impossible to know which parameter has caused an issue. Unless there is a Backend specific API to query that kind of thing 🤔

PR #1165, opened 2024-08-31
Move trivial blob backend to public includes

### LINKED ISSUE #1153: File "private/soci-trivial-blob-backend.h" missing
@Krzmbrzl, the commit #992 includes `#include <private/soci-trivial-blob-backend.h>` which is not present when installed with CMake.

Krzmbrzl:
Since private headers are not (and also should not be) installed, this makes it impossible to compile SOCI as an installed package.

Fixes #1153


vadz:
Thanks! Will merge soon.

PR #1171, opened 2024-10-10
Run macOS CI jobs under macOS 14

vadz:
The currently used macOS 12 images are deprecated and will cease to be available on GitHub soon.


vadz:
Merged in master after rebase.

PR #1174, opened 2024-10-22
Split common tests into a common library

vadz:
Avoid recompiling `common-tests.h` many, many times.


vadz:
@Krzmbrzl Sorry for the extra conflicts with your PR but I'd like to merge this. I hope the conflicts won't be that difficult to resolve, and in the meanwhile this will make iterating over new tests much faster.

PR #1183, opened 2024-11-18
fix bug in the reading from oracle clob

avpalienko:
OCILobGetLength and OCILobRead for character LOBs (CLOBs, NCLOBs) uses the number of characters, not bytes. So, there is a bug with buffer array sizing

### DIFF
--- a/src/backends/oracle/standard-into-type.cpp
+++ b/src/backends/oracle/standard-into-type.cpp
@@ -232,44 +232,56 @@ void oracle_standard_into_type_backend::pre_fetch()
     }
 }
 
+/* ----------------------------------------------------------------- 
+/* callback function to write buffer to the string.     
+ * works in bytes, not in characters                 
+/* ----------------------------------------------------------------- */
+sb4 callback_to_read_in_bytes
+(
+    dvoid                             *pResultString,  // pointer to std::string to accumulate result
+    CONST dvoid                       *bufxp,          // pointer to buffer from read
+    ub4                                lenp,           // length data in the buffer
+    ub1                                piecep          // OCI_FIRST_PIECE, OCI_NEXT_PIECE or OCI_LAST_PIECE
+)
+{
+    static_cast<std::string*> ( pResultString )->append ( static_cast<const char *> ( bufxp ), lenp );
+    return OCI_CONTINUE;
+}
+
+
 void oracle::read_from_lob(oracle_session_backend& session,
     OCILobLocator * lobp, std::string & value)
 {
     ub4 len;
 
-    sword res = OCILobGetLength(session.svchp_, session.errhp_,
-        lobp, &len);
+    // result is not a bytes count!!! For character LOBs (CLOBs, NCLOBs) it is the number of characters
+    sword res = OCILobGetLength ( session.svchp_, session.errhp_, lobp, &len );
     if (res != OCI_SUCCESS)
     {
         throw_oracle_soci_error(res, session.errhp_);
     }
 
-    std::vector<char> buf(len);
-
-    if (len != 0)
+    ub4 amtp = 0;
+    ub4 offset = 1;
+    char buf[10240] = {};
+    std::string accumTo;
+    res = OCILobRead
+    (
+        session.svchp_, session.errhp_, lobp,   // service context, error handle, LOB locator
+        &amtp,                                  // amount (in characters) to read. 0 means until the end of the LOB in streaming mode
+        offset,                                 // offset == 1 means from the start of the LOB    
+        &buf,
+        sizeof (buf),
+        &accumTo,                               // ctx for the callback - pointer to the result string        
+        callback_to_read_in_bytes,              // callback function works in bytes!!!
+        0,
+        0
+    );
+    if ( res != OCI_SUCCESS )
     {
-        ub4 lenChunk = len;
-        ub4 offset = 1;
-        do
-        {
-            res = OCILobRead(session.svchp_, session.errhp_,
-                lobp, &lenChunk,
-                offset,
-                reinterpret_cast<dvoid*>(&buf[offset - 1]),
-                len, 0, 0, 0, 0);
-            if (res == OCI_NEED_DATA)
-            {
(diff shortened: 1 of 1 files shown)


vadz (commented):
Is the switch to using a callback really needed? If the problem is really just a wrong byte count, couldn't we multiply (divide?) the length by whatever to fix it?

Also, it would be tremendously helpful if you could please add a unit test checking for this problem, i.e. failing now and passing with this fix. Could you please do this? TIA!

avpalienko:
> Is the switch to using a callback really needed? If the problem is really just a wrong byte count, couldn't we multiply (divide?) the length by whatever to fix it?

If we should avoid the callback we can try  OCILobRead2 
How to know the multiplier ?

> 
> Also, it would be tremendously helpful if you could please add a unit test checking for this problem, i.e. failing now and passing with this fix. Could you please do this? TIA!

Ok

src/backends/oracle/standard-into-type.cpp
```diff
@@ -232,44 +232,56 @@ void oracle_standard_into_type_backend::pre_fetch()
...
-            {
-                throw_oracle_soci_error(res, session.errhp_);
-            }
-        }
-        while (res == OCI_NEED_DATA);
+        throw_oracle_soci_error ( res, session.errhp_ );
     }
-
-    value.assign(buf.begin(), buf.end());
+    value = std::move ( accumTo );
```
vadz, line 284: BTW, this makes no sense, we should just pass `value` to the callback instead of `accumTo` directly.

src/backends/oracle/standard-into-type.cpp
```diff
@@ -232,44 +232,56 @@ void oracle_standard_into_type_backend::pre_fetch()
...
     {
         throw_oracle_soci_error(res, session.errhp_);
     }
 
-    std::vector<char> buf(len);
-
-    if (len != 0)
+    ub4 amtp = 0;
+    ub4 offset = 1;
+    char buf[10240] = {};
```
vadz, line 266: Why do we use both a buffer and a callback? Shouldn't we not need the buffer any longer if we use the callback?

tests/oracle/test-oracle.cpp
```diff
@@ -1407,6 +1407,62 @@ TEST_CASE("Bulk iterators", "[oracle][bulkiters]")
...
+        sql <<
+             "create or replace function soci_test(input in varchar2) return clob as "
+             "begin return input; end;";
+    }
+};
+
+TEST_CASE ( "Oracle clob", "[oracle][clob]" )
+{
+    soci::session sql ( backEnd, connectString );
+    std::string testPhrase{ "SystГЁme est ouvert aux requГЄtes d'information" };
```
vadz, line 1448: Is this intentionally misencoded?

vadz (commented):
Sorry, this is changing more often than I can review it, so I'll stop for now, but let me publish some comments I had so far.

src/backends/oracle/standard-into-type.cpp
avpalienko: oci read into the buffer and callback after

tests/oracle/test-oracle.cpp
avpalienko: yes

vadz: Well, first of all, this doesn't really work even with this PR apparently due to this.

Second, any I really don't want to be mean, but just I don't understand how can a so obviously questionable line be added without any comment explaining it. You should know that something so clearly unusual and apparently wrong would raise questions from whoever is reading it, and I can't fathom why you wouldn't want to explain that the string must be wrongly encoded to trigger the bug (if this is how it really is).

vadz:
In fact, the same problem happens with a plain `VARCHAR2()` too, not just `CLOB`, so even if there is a problem with count mismatch for CLOBs, it's clearly not the only problem here.

I don't know anything about charsets in Oracle, but it looks like we need to specify that we're using UTF-8 somehow. It's pretty weird that this doesn't work out of the box, even though `NLS_CHARACTERSET` database parameter is set to `AL32UTF8` and it would be nice to understand what's going on here, but I'm pretty sure that this PR is not the right solution for the problem, whatever it is.

avpalienko:
> In fact, the same problem happens with a plain `VARCHAR2()` too, not just `CLOB`, so even if there is a problem with count mismatch for CLOBs, it's clearly not the only problem here.

There are no problems with varchar2, export NLS_LANG is required on client session

> I don't know anything about charsets in Oracle, but it looks like we need to specify that we're using UTF-8 somehow. It's pretty weird that this doesn't work out of the box, even though `NLS_CHARACTERSET` database parameter is set to `AL32UTF8` and it would be nice to understand what's going on here, but I'm pretty sure that this PR is not the right solution for the problem, whatever it is.

I have changed oracle.sh to use utf-8 encoding

src/backends/oracle/standard-into-type.cpp
vadz: It's not very nice to mark a comment as resolved when it wasn't addressed at all.

vadz:
Thanks, I forgot about `NLS_LANG` weirdness. However I still don't think this has anything to do with using callback — does the (simpler) solution in #1184 work for you or do you still see any problems with it and, if so, how can they be reproduced?

avpalienko:
> Thanks, I forgot about `NLS_LANG` weirdness. However I still don't think this has anything to do with using callback — does the (simpler) solution in #1184 work for you or do you still see any problems with it and, if so, how can they be reproduced?

What is the advantage of using a self-made loop over an oci internal loop with the callback? 
However, I won't insist on my solution. What matters to me is resolving the bug

vadz:
> What is the advantage of using a self-made loop over an oci internal loop with the callback?

I think using both the buffer and the callback is confusing, if the OCI really doesn't allow to do without the buffer in this case, I'd rather avoid defining the callback.

I don't know if the behaviour is the same with `OCILobRead2()`, maybe we could make things simpler using it, but I didn't have time to try it.

PR #1179, opened 2024-11-14
Add std::wstring support for ODBC backend

vadz:
This is for testing of squash-merged #1133 with some changes.

PR #1184, opened 2024-11-19
Fix support for non-ASCII characters in Oracle CLOBs

vadz:
Replaces #1183.

### DIFF
--- a/src/backends/oracle/standard-into-type.cpp
+++ b/src/backends/oracle/standard-into-type.cpp
@@ -235,41 +235,48 @@ void oracle_standard_into_type_backend::pre_fetch()
 void oracle::read_from_lob(oracle_session_backend& session,
     OCILobLocator * lobp, std::string & value)
 {
-    ub4 len;
-
-    sword res = OCILobGetLength(session.svchp_, session.errhp_,
-        lobp, &len);
+    // We can't get the CLOB size in bytes directly, only in characters, which
+    // is useless for UTF-8 as it doesn't tell us how much memory do we
+    // actually need for storing it, so we'd have to allocate 4 bytes for every
+    // character which could be a huge overkill. So instead read the CLOB in
+    // chunks of its natural size until we get everything.
+    ub4 len = 0;
+    sword res = OCILobGetChunkSize(session.svchp_, session.errhp_, lobp, &len);
     if (res != OCI_SUCCESS)
     {
         throw_oracle_soci_error(res, session.errhp_);
     }
 
+    value.clear();
+
+    if (!len)
+        return;
+
+    // Read the LOB in chunks into the buffer while anything remains to be read.
     std::vector<char> buf(len);
-
-    if (len != 0)
+    ub4 offset = 1;
+    do
     {
-        ub4 lenChunk = len;
-        ub4 offset = 1;
-        do
-        {
-            res = OCILobRead(session.svchp_, session.errhp_,
-                lobp, &lenChunk,
-                offset,
-                reinterpret_cast<dvoid*>(&buf[offset - 1]),
-                len, 0, 0, 0, 0);
-            if (res == OCI_NEED_DATA)
-            {
-                offset += lenChunk;
-            }
-            else if (res != OCI_SUCCESS)
-            {
-                throw_oracle_soci_error(res, session.errhp_);
-            }
-        }
-        while (res == OCI_NEED_DATA);
-    }
+        // By setting the input length to 0, we tell Oracle to read as many
+        // bytes as possible (so called "streaming" mode).
+        ub4 lenChunk = 0;
+        res = OCILobRead(session.svchp_, session.errhp_, lobp,
+                         &lenChunk, offset,
+                         &buf[0], len,
+                         0, 0, 0, 0);
 
-    value.assign(buf.begin(), buf.end());
+        if (res == OCI_NEED_DATA)
+        {
+            offset += lenChunk;
+        }
+        else if (res != OCI_SUCCESS)
+        {
+            throw_oracle_soci_error(res, session.errhp_);
+        }
+
+        value.append(buf.begin(), buf.begin() + lenChunk);
+    }
+    while (res == OCI_NEED_DATA);
 }
 
 void oracle_standard_into_type_backend::post_fetch(
--- a/src/backends/oracle/standard-into-type.cpp
+++ b/src/backends/oracle/standard-into-type.cpp
@@ -254,29 +254,33 @@ void oracle::read_from_lob(oracle_session_backend& session,
 
     // Read the LOB in chunks into the buffer while anything remains to be read.
     std::vector<char> buf(len);
-    ub4 offset = 1;
-    do
+    for (bool done = false; !done; )
     {
         // By setting the input length to 0, we tell Oracle to read as many
         // bytes as possible (so called "streaming" mode).
         ub4 lenChunk = 0;
         res = OCILobRead(session.svchp_, session.errhp_, lobp,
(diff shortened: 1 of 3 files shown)


src/backends/oracle/standard-into-type.cpp
```diff
@@ -235,41 +235,48 @@ void oracle_standard_into_type_backend::pre_fetch()
...
+    // chunks of its natural size until we get everything.
+    ub4 len = 0;
+    sword res = OCILobGetChunkSize(session.svchp_, session.errhp_, lobp, &len);
     if (res != OCI_SUCCESS)
     {
         throw_oracle_soci_error(res, session.errhp_);
     }
 
-    std::vector<char> buf(len);
+    value.clear();
```
avpalienko, line 250: Could this be an issue if an exception is thrown later?

vadz: I don't think we make any guarantees about what the contents of `value` is if an exception is thrown, so I don't think there is any issue here, but please let me know if I'm overlooking something.

src/backends/oracle/standard-into-type.cpp
```diff
@@ -235,41 +235,48 @@ void oracle_standard_into_type_backend::pre_fetch()
...
+
+    if (!len)
+        return;
 
-    if (len != 0)
+    // Read the LOB in chunks into the buffer while anything remains to be read.
+    std::vector<char> buf(len);
+    ub4 offset = 1;
```
avpalienko, line 257: offset is not needed at all
"If you use streaming (by polling or a callback), specify the offset in the first call; in subsequent polling calls, the offset parameter is ignored. "

avpalienko:
There might be an issue with the Fixed-Width Client-Side Character Set. In this case, Oracle states that the output value of amtp is in characters. However, I haven't been able to create a test case for it.

src/backends/oracle/standard-into-type.cpp
vadz: Thanks, I've removed `offset` variable now.

vadz:
> There might be an issue with the Fixed-Width Client-Side Character Set. In this case, Oracle states that the output value of amtp is in characters.

Oracle documentation says

> The output amount indicates how many bytes were read into the buffer bufp.

so I don't think it's ever in characters (except on input — using different units for an in/out parameter deserves some kind of a prize for the worst API design ever).

avpalienko:
> > There might be an issue with the Fixed-Width Client-Side Character Set. In this case, Oracle states that the output value of amtp is in characters.
> 
> Oracle documentation says
> 
> > The output amount indicates how many bytes were read into the buffer bufp.
> 
> so I don't think it's ever in characters (except on input — using different units for an in/out parameter deserves some kind of a prize for the worst API design ever).

I am basing this on the following
image

vadz:
Hmm, yes. I wonder if they really include UTF-16 and UTF-32 in "fixed width encodings", I have a feeling that they might be speaking about pre-Unicode fixed width encodings only (e.g. CP1252 etc). In any case, I can't test this neither: setting `NLS_LANG` to `.AL32UTF16` or anything like this results in a failure to create the Oracle environment.

src/backends/oracle/standard-into-type.cpp
```diff
@@ -235,41 +235,58 @@ void oracle_standard_into_type_backend::pre_fetch()
...
 
-    if (len != 0)
+    // Read the LOB in chunks into the buffer while anything remains to be read.
+    for (bool done = false; !done; )
     {
-        ub4 lenChunk = len;
-        ub4 offset = 1;
-        do
+        auto const prevSize = value.size();
+        value.resize(prevSize + len);
```
avpalienko, line 259: Could it be that "reserve" should be used here?

avpalienko:
> Hmm, yes. I wonder if they really include UTF-16 and UTF-32 in "fixed width encodings", I have a feeling that they might be speaking about pre-Unicode fixed width encodings only (e.g. CP1252 etc). In any case, I can't test this neither: setting `NLS_LANG` to `.AL32UTF16` or anything like this results in a failure to create the Oracle environment.

Currently, only AL16UTF16 is relevant. But SOCI is not ready to use it

src/backends/oracle/standard-into-type.cpp
vadz: No, accessing the buffer wouldn't be valid without `resize()`.

vadz:
> Currently, only AL16UTF16 is relevant. But SOCI is not ready to use it

I get an error (without any details) from `OCISessionBegin()` with this charset too. Maybe we could support UTF-16 with #1179 but I'm not really interested in this, to be honest.

avpalienko:
> > Currently, only AL16UTF16 is relevant. But SOCI is not ready to use it
> 
> I get an error (without any details) from `OCISessionBegin()` with this charset too. Maybe we could support UTF-16 with #1179 but I'm not really interested in this, to be honest.

The username and password must be in Unicode in this case, and SQL texts must also be in Unicode. I worked on it today but didn't achieve any results.
SOCI connection parameters include charset and ncharset, but I'm not sure if they are equivalent to the NLS_LANG setting.

src/backends/oracle/standard-into-type.cpp
avpalienko: Then we have resize being called twice in a row on the second and subsequent iterations. See line 284

vadz: Yes, but this is correct and intentional: we first resize to get a big enough buffer and then resize to the actual length. The second resize won't allocate any memory (and may, in fact, do nothing at all, if `lenChunk == len`).

What do you think is a problem?

vadz: Yes, but I see nothing wrong here. If you do, please explain what.

I have no idea why does Oracle return us only 2000 bytes when we told it to get as much as possible that fits in 4000, but I don't suppose we can change how it behaves.

avpalienko: May be try so:
``

    // Read the LOB in chunks into the buffer while anything remains to be read.
    auto theUsedSize = value.size();
    do 
    {
        // resize to max possible read size
        value.resize( theUsedSize + len );

        // By setting the input length to 0, we tell Oracle to read as many
        // bytes as possible (so called "streaming" mode).
        ub4 lenChunk = 0;
        res = OCILobRead(session.svchp_, session.errhp_, lobp,
                         &lenChunk,
                         1, // Only used for the first chunk, ignored later.
                         const_cast<char*>(value.data()) + theUsedSize, len,
                         0, 0, 0, 0);
        theUsedSize += lenChunk;

        if ( res != OCI_SUCCESS && res != OCI_NEED_DATA )
        {
            throw_oracle_soci_error ( res, session.errhp_ );
        }

    } while ( res == OCI_NEED_DATA );

    // resize to the actual size
    value.resize( theUsedSize );
    // We may have over-allocated the string, especially when a big chunk size
    // is used, so release the unused memory.
    value.shrink_to_fit();

``

vadz: This would work too AFAICS but seems a bit more difficult to understand to me.

It also results in exactly the same allocation pattern because resizing the string to shrink doesn't allocate (or free) any memory and is a very cheap operation (it basically adjusts one pointer or size, i.e. about the same as `theUsedSize += lenChunk`).

IOW I still have no idea what problem does this solve?

avpalienko: No problem at all, but my approach seems a bit more readable to me and potentially more efficient. However, performance is not crucial in this case. I’m perfectly fine with your solution as well

PR #1189, opened 2025-01-02
Allow to row::get blob into suitable containers

### LINKED ISSUE #1173: Reading BLOB from SQLite3 no longer working
Hi,

I've updated an application using SOCI from `9bcc5f8a9181886f4c73ea5b4671b35d8722cb3a` to `a55c80984a447bf8c788a7d6a23ff93c022069fa`.

The only problem I encountered so far is with regards of reading a `BLOB` from an SQLite3 database.
Previously, the application did this (inside of a custom/user type conversion):
```
std::string str = v.get<std::string>("my_column");
```
Where:
- `my_column` is an SQLite3 column of type `BLOB`
- `v` is an lvalue reference to `soci::base_type`

This used to work well. However, after the update the code is throwing an `std::bad_cast` in `soci::details::holder::get()` (called by `soci::row::get()`).

According to the git log there have been some activities around blob support.

What is the new, correct way of reading (and writing!) a blob to/from SQLite3?

Krzmbrzl:
"Suitable" means that the `soci::is_contiguous_resizable_container` is specialized to have its `value` member yield `true`. This specialization is provided for `std::string` and `std::vector<T>` where `sizeof(T) == sizeof(char)`.

Fixes #1173

### DIFF
--- a/docs/types.md
+++ b/docs/types.md
@@ -138,6 +138,53 @@ Users are encouraged to use the latter as it supports a wider range of numerical
 The mapping of underlying database column types to SOCI datatypes is database specific.
 See the [backend documentation](backends/index.md) for details.
 
+While `row::get<T>()` (semantically) copies the internally stored data (that was fetched from the database), there is also `row::move_as<T>`, which
+instead makes use of C++11's move semantics by moving the data out of the `row` instance. In case the used `T` does not support move semantics,
+`move_as` is functionally equivalent to `get`.
+
+### Dealing with Blobs
+
+If the fetched data is of type `db_blob` it is strongly recommended to use `row::move_as<blob>()` in order to obtain the data as a `blob` object. This
+allows for the most flexible and efficient access of the underlying data.
+
+Note that it is not possible to `row::get<blob>()` as `blob` objects are not copyable. However, you can instead get the blob's data into a
+`std::string` or a `std::vector<T>` as long as `T` has a size of exactly one byte (i.e. is a byte-like type such as `char` or `std::byte`). Depending
+on the used backend, this will either directly fetch the data from the database directly into the provided container or copy the already fetched data
+into the provided container.
+
+If you want to use a different container than the ones mentioned above, you need to specialize the `soci::is_contiguous_resizable_container` trait.
+E.g.
+```cpp
+template<>
+struct ::soci::is_contiguous_resizable_container<MyAwesomeContainer> : std::true_type {};
+
+template<typename ValueType>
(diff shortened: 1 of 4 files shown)


Krzmbrzl:
> I'd say that string and vector are the only containers that you may ever want to read BLOBs into but if we want to be more generic, we could just specialize is_contiguous_byte_container for them by default but document that it be specialized for custom types too: it seems like it would be simpler and safer.

It would be safer, yes, but also much more hassle (compared to "it just works") to get working. What I have in mind is someone using e.g. Boost containers.

The current implementation relies on any container type being passed in to be defined _sensibly_, that is in analogy with the standard containers. I think as long as this is true, then the assumptions in this PR hold (including the implicitly assumed availability member functions `size` and `resize`). That's why I thought that it is probably safe to go the plug-and-play route I chose.

If you know of even a single commonly-ish used container that would pass my heuristic without having contiguous storage in memory, then we obviously can't use this. However, I was unable to come up with one :eyes:

vadz:
I think any legacy custom container wouldn't have `resize()` (the ones I worked with definitely didn't) and I do think that we need to test for its presence using SFINAE if we keep this version. This could be done, of course, but IMO it's not worth it.

I'd really like to keep the code as simple as possible while still avoiding regressions (such as #1173) and just allowing reading a BLOB into a string would be good enough for me. Allowing reading it into a vector is even more enough, I'm not even sure we need a trait at all — but I'm ready to have it if you think it's worth it, just let's maximally simplify it.

OTOH I thought of something else: why don't we specialize `soci_cast` for `blob` itself? It looks like we should be able to copy blobs using it, and I think this may have been allowed before as well, wasn't it?

Krzmbrzl:
I'll add a trait and a default implementation for an accessor. This way, the code can be more or less fully customized if needed, but the logic on SOCI's side is minimal.
I'll specialize the trait for `std::string` and `std::vector<T>` where `sizeof(T) == sizeof(char)`.

> OTOH I thought of something else: why don't we specialize soci_cast for blob itself? It looks like we should be able to copy blobs using it, and I think this may have been allowed before as well, wasn't it?

BLOBs are inherently non-copyable and I believe they always have been. Making Blobs copyable might be possible but I think this would involve a significant amount of work as different backends have different semantics of what a Blob actually represents and how they have to be accessed. At the very least the Blob handle (where present) would have to have shared_ptr like semantic to ensure that we don't have double-frees. However, copying Blobs would then only copy the SOCI Blob instance. Yet both instances would still point to the same "physical" Blob in the DBs. That's because for most backends, SOCI Blobs have more of a pointer-like semantic. Yet, this is not true for all backends (e.g. MySQL).

So all in all I believe that the Blob backends are just too different to support this with a consistent API and with consistent semantics.

Issue #560, opened 2017-07-20
Use ELF visibility to export only public symbols from the library


vadz:
We should really define `SOCI_DLL` appropriately not only in Win32 builds, but also when ELF visibility is supported for all the usual reasons (more efficient shared library loading; consistency between platforms).

As usual, I don't know how to handle this at CMake level, but if anyone can take care of this, I could do the required (trivial) changes to the sources.

phetdam:
Very late but this is something I've had to personally work around on Windows when using the SOCI DLLs.

Take for example use of the ODBC backend DLL. The [ODBC backend docs](https://github.com/SOCI/soci/blob/c5678e172bff57c61c9a3a60aa253cebf98837ea/docs/backends/odbc.md) show use of `soci::odbc`, declared as:

https://github.com/SOCI/soci/blob/c5678e172bff57c61c9a3a60aa253cebf98837ea/include/soci/odbc/soci-odbc.h#L612

However, without `SOCI_DLL` defined during compilation, trying to access the `soci::odbc` symbol while using `SOCI::soci_odbc` in a `target_link_libraries` call will just result in a linker error with MSVC since `SOCI_DECL_IMPORT`, and thus `SOCI_ODBC_DECL`, is just defined to nothing. To work around this, I made my interface targets for my own project, e.g.

```cmake
# SOCI core library, propagating SOCI_DLL when using the shared target
add_library(myproj::soci_core INTERFACE IMPORTED)
target_link_libraries(myproj::soci_core INTERFACE SOCI::soci_core)
target_compile_definitions(myproj::soci_core INTERFACE SOCI_DLL)
# assume something similar for a myproj::soci_odbc target
```

Linking against these targets now works seamlessly because the `SOCI_DLL` macro definition is propagated. Something similar can be done for SOCI's own targets, e.g. `target_compile_options(${SOCI_CORE_TARGET} PUBLIC SOCI_DLL)`.

phetdam:
[...]

PR #1192, opened 2025-01-16
enable consistent DSO symbol export across POSIX/Windows + SOCI_DLL usage

phetdam:
### Summary

This PR addresses the very old #560 by ensuring that the `SOCI_DECL` macro and related symbol import/export macros in `soci-platform.h` are consistently defined with the `SOCI_DLL` macro that indicates that SOCI is being used as a shared library. Furthermore, `SOCI_DLL` is set as a `PUBLIC` CMake compile definition for the shared SOCI library target so that downstream consumers of `SOCI::soci_core` can directly use `target_link_libraries` with their own CMake project targets.

Furthermore, the `soci_backend` CMake macro used to add a new SOCI backend library is enhanced to use [`DEFINE_SYMBOL`](https://cmake.org/cmake/help/latest/prop_tgt/DEFINE_SYMBOL.html) appropriately, e.g. defining `SOCI_<backend>_SOURCE` when compiling a SOCI backend library, obviating the need for the manual `#define SOCI_<backend>_SOURCE` in each backend's source files. Similarly, the SOCI core target uses `DEFINE_SYMBOL` to define `SOCI_SOURCE` so `#define SOCI_SOURCE` can be removed from the SOCI core target sources.

The changes in this PR have been made with the [GCC visibility wiki](https://gcc.gnu.org/wiki/Visibility) as an additional reference. Some miscellaneous fixes are also included in the PR, namely ensuring `failover_callback` in `callbacks.h` does not have a `SOCI_DECL` annotation (as the class is wholly defined inline; `SOCI_DECL` just confuses the linker), and building a separate `soci_tests_common_static` static library since the shared SOCI library tests linking against `soci_tests_common` require `SOCI_DLL` to be defined.


phetdam:
Note that despite the huge (100+) number of changed files, the vast majority of the files are sources that have simply had their `#define SOCI_<backend>_SOURCE` statements removed. CMake's `DEFINE_SYMBOL` takes care of this macro definition.

vadz:
Thanks, but I wonder if it's not going to create more conflicts with #1118.

@Krzmbrzl Any comments?

Krzmbrzl:
I would very much appreciate if no further CMake changes are merged into master. This PR could probably be based on top of #1118 rather easily though. Either we can include it in that PR or have the rebased PR ready for when the CMake PR gets merged.

phetdam:
Hey, sorry for the late response everyone.

As I mentioned [above](https://github.com/SOCI/soci/pull/1192#issuecomment-2594432862) the majority of the files touched are `.cpp` files that have had their manual `#define SOCI_<backend>_SOURCE` lines removed. This also includes the core `.cpp` sources that have `#define SOCI_SOURCE` lines in them.

Material changes are fortunately pretty limited in scope and diff size, namely the following:

1. Updating `soci_backend` and SOCI core target to use `DEFINE_SYMBOL` appropriately
2. Define `SOCI_DLL` for all shared target definitions
3. Adding a separate `soci_tests_common_static` since `soci_tests_common` has `SOCI_DLL` as a compile definition
4. Require consistent `SOCI_DLL` usage by ensuring decl macros are nonempty for import/export when `SOCI_DLL` defined
5. Removing the `SOCI_DECL` annotation from `failover_callback` since the class is fully defined inline
6. Minor tweak to `FindMySQL.cmake` because `MYSQL_INCLUDE_DIR` wasn't always defined/nonempty

Since #1118 is PR of massive scope I can see that merging against `master` to ingest some of these changes would certainly result in a large set of merge conflicts (just the nature of the large refactoring). In general, I would prefer having these changes merged to `master` (just personal philosophy of merge improvements ASAP when possible) rather than wait. But in either case, I'd be happy to work with @Krzmbrzl to get these in #1118 as the changes are quite targeted.

Krzmbrzl:
Another thing that might be worth considering: By moving the `SOCI_SOURCE` definitions from being in the actual source files to letting CMake define them as appropriate would (could) mean that building SOCI with non-CMake buildsystems. Personally, I don't need that but I think vadz is typically using something else.

phetdam:
Ah, it's pretty straightforward to work around this. When building the SOCI core or backend libraries, the `SOCI_SOURCE` or `SOCI_<backend>_SOURCE` definition just needs to be added to the command line, e.g. with `-DSOCI_SOURCE`.

This is the non-CMake way to do it when using Make/MSBuild to build projects.

Krzmbrzl:
Yep. It's just something to keep in mind and I wanted to point out the need to do it :)

phetdam:
Hm, all of a sudden the ODBC CI run is failing during installation of dependencies:

```
Running /__w/soci/soci/scripts/ci/install_odbc.sh
Reading package lists...
Building dependency tree...
Reading state information...
Package 'libodbc1' is not installed, so not removed
Package 'odbcinst1debian2' is not installed, so not removed
Package 'unixodbc-dev' is not installed, so not removed
Package 'unixodbc' is not installed, so not removed
0 upgraded, 0 newly installed, 0 to remove and 1 not upgraded.
Reading package lists...
Building dependency tree...
Reading state information...
E: Unable to locate package unixodbc-dbgsym
E: Unable to locate package odbc-postgresql-dbgsym
Error: Process completed with exit code 100.
```

I haven't touched any CI-related stuff in this PR at all so entirely sure what's happening here.

vadz:
@phetdam The CMake PR is merged now, please rebase this one on it. TIA!

vadz:
I don't know if we want to include this in 4.1.0 release, but if we do, there is very little time left until it.

vadz:
I've just made the RC3 which will, hopefully, become the final release. I.e. AFAIK the release could be done any day, including today — but we can definitely wait a bit if you think that

(a) This is important enough to have in 4.1.0.
(b) Plan to have some time to work on this in the near future.

phetdam:
I don't want to be the reason that the big 4.1.0 release gets delayed too much; am totally fine to get this in as part of a patch release instead. But if there is still within a week's time, then I can get started on rebasing :)

phetdam:
Oh drat, I am going to need to upgrade my CMake version on Ubuntu to 3.23 from my current 3.22.1.

For Windows I'm ok as I have CMake 3.26.0 provided by Visual Studio.

vadz:
After looking at the related stuff due to #1248, I have to say that it looks like most parts of what this PR is supposed to do seems to be either already done or could be done very easily:

1. `SOCI_DLL` is added using `target_compile_definitions(soci_core PUBLIC)`, i.e. it's already inherited by everybody using SOCI.
2. `SOCI_${BACKEND}_SOURCE` could be defined in `soci_define_backend_target()` and this would be a pretty simple change.
3. `SOCI_DECL` has already been removed from `failover_callback` in 85985e3a (Remove SOCI_DECL for callbacks (as class is fully inline), 2025-02-07).

So I think it could be better to just close this and implement (2) in a later PR if really necessary (personally I don't really mind doing it the current way).

phetdam:
Hey, sorry for not attending to this sooner.

> either already done or could be done very easily:

Yes, originally this PR's scope was intentionally limited; although it touched a decent number of files, the nature of their change was really the same (removing the manual `#define` since the CMake backend registration macro would add it during compile). Given that there have been a lot of CMake-related changes, I also think it might be easier to start clean off `master`.

Thanks and I'll try to get back onto this using a fresh PR when I can :)

PR #1200, opened 2025-01-30
Show vector parameters values in error messages

vadz:
This addresses the problem of not knowing which vector element resulted in the error when using bulk operations. For now this is implemented for some backends only, but it's already better than nothing.


vadz:
This is ready now, I've implemented this for all backends except DB/2 (which I can't test) and MySQL (which I could but don't use, so I'm leaving this for somebody who cares about it).

If there are no comments/objections, I'll merge this soon.

Issue #1210, opened 2025-03-04
PostgreSQL connection parameter parsing regression


Krzmbrzl:
@vadz just to make you aware, updating SOCI to latest master caused my program to no longer be able to connect to a PostgreSQL database. From a first glance it seems like the culprit is the changed parameter parsing.

I'll investigate some more but given that backwards compatibility always seemed important to you, I figured I'll let you know ASAP (before you finished drafting the 4.1 release)

vadz:
Yes, it would be certainly great to fix this, thanks for reporting. What is your connection string exactly?

Krzmbrzl:
The connection string I am using is
```
postgresql://dbname='mumble_test_db' user='mumble_test_user' password='MumbleTestPassword' host='127.0.0.1'
```
and the error I get is
```
could not translate host name "'127.0.0.1'" to address: Name or service not known
```
so it seems like it doesn't like the single quotes (looking at the code, it should handle double quotes fine but doesn't do anything to single quotes. though the old code seems to have done the same).

Interestingly, the following works fine for MySQL:
```
mysql://dbname='mumble_test_db' user='mumble_test_user' password='MumbleTestPassword'
```
but here single quotes are only used for string values...

Krzmbrzl:
I believe the culprit is due to [this part of the docs](https://www.postgresql.org/docs/current/libpq-connect.html#LIBPQ-CONNSTRING-KEYWORD-VALUE):
> To write an empty value, or a value containing spaces, surround it with **single quotes**, for example keyword = 'a value'.

In my input string, all arguments are already quoted properly, but it seems that SOCI surrounds arguments by another layou of single quotes?
https://github.com/SOCI/soci/blob/5e7a97d41a5f7cc0a3b4c8172daa48fbb823effc/src/backends/postgresql/session.cpp#L182
https://github.com/SOCI/soci/blob/5e7a97d41a5f7cc0a3b4c8172daa48fbb823effc/src/core/connection-parameters.cpp#L340-L381

and presumably before https://github.com/SOCI/soci/commit/1a6607bc2fe31a909e7b7e00ec2a0c5a8afd7eae the already quoted parameters were simply forwarded to PostgreSQL?

vadz:
Previously we didn't do any parsing at all, so the string was passed to libpq as given.

Now we do it in order to be able to extract our own ones, but we expect double quotes in the input.

[...]

PR #1212, opened 2025-03-04
Support single-quotes for parsing connection strings

Krzmbrzl:
Fixes https://github.com/SOCI/soci/issues/1210

### DIFF
--- a/src/core/connection-parameters.cpp
+++ b/src/core/connection-parameters.cpp
@@ -240,23 +240,43 @@ getPossiblyQuotedWord(std::string const &s, std::string::const_iterator &i)
     std::string::const_iterator const end = s.end();
     skipWhiteSpace(i, end);
 
+    if (i == end)
+    {
+        return {};
+    }
+
+    char quote = '"';
+    if (*i == '\'')
+    {
+        quote = '\'';
+    }
+
     std::string word;
 
-    if (i != end && *i == '"')
+    if (i != end && *i == quote)
     {
         for (;;)
         {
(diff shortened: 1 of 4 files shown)


Krzmbrzl:
I am currently assuming that it is enough for the parsing to properly understand that single quotes are also quotes and that other parts of the code don't implicitly rely on the parsed connection string using a specific kind of quotes.

The accessor functions for the original connection string make me a bit nervous in that regard... but I guess CI will tell us whether this change broke something or not :eyes:

src/core/connection-parameters.cpp
```diff
@@ -240,23 +240,43 @@ getPossiblyQuotedWord(std::string const &s, std::string::const_iterator &i)
...
         for (;;)
         {
             if (++i == end)
             {
                 std::ostringstream os;
-                os << "Expected '\"' not found before the end of the string "
+                os << "Expected ";
+                if (quote == '\'')
+                {
+                    os << "'";
```
vadz, line 266: If you want to show it differently I'd do this for symmetry:
```suggestion
                    os << R"("'")";
```

but I think we could also just use the `else` version in all cases.

vadz (approved):
Looks good, thanks! I'll (squash) merge it soon.

vadz:
Merged as caaf5627a119bf733560a0826217155b006c72b0 with minor changes to make this a bit more symmetrical between single and double quotes, thanks!

Will also push to 4.1 branch soon.

PR #1217, opened 2025-03-18
Make SQLite overwrite exchange type for integers

### LINKED ISSUE #1190: `row::get<int64_t>()` overflows 32bit integer in internal representation (SQLite)
This is a continuation of #1185. The MWE remains the same:
```cpp
#include <soci/soci.h>
#include <soci/sqlite3/soci-sqlite3.h>
#include <iostream>

int main() {
    try {
        // Create an in-memory SQLite database session
        soci::session sql(soci::sqlite3, ":memory:");
        sql << "CREATE TABLE test_table (id INTEGER PRIMARY KEY, myInt INTEGER);";
        sql << "INSERT INTO test_table (myInt) VALUES (9223372036854775807);";

        // This works
        int64_t retrievedValue1;
        sql << "SELECT myInt FROM test_table WHERE id = 1", soci::into(retrievedValue1);
        std::cout << "Implicit retrieval: " << retrievedValue1 << std::endl;

        // This throws std::bad_cast (windows) or overflows int32 for a -1 (linux)
        soci::row r;
        sql << "SELECT myInt FROM test_table WHERE id = 1", soci::into(r);
        int64_t retrievedValue2 = r.get<int64_t>("myInt");
        std::cout << "Explict 64-bit retrieval: " << retrievedValue2 << std::endl;

    } catch (const std::exception& e) {
        std::cerr << "Error: " << e.what() << std::endl;
    }

    return 0;
}
```
On current `master` this compiles fine, but the `get` function of the `row` object will return `-1` (on Linux), which seems to suggest that internally SOCI is using a 32bit integer representation. However, this seems wrong as clearly SQLite can support 64bit integer values for `INTEGER` columns.
This is likely related to SQLite not really supporting data types. The [docs](https://sqlite.org/datatype3.html) describe the `INTEGER` type as
> INTEGER. The value is a signed integer, stored in 0, 1, 2, 3, 4, 6, or 8 bytes depending on the magnitude of the value.

Krzmbrzl:
Since SQLite doesn't have a type system, we have to always be prepared to select huge numbers, even if the "type" is e.g. int8. To do that, we overwrite the exchange type (e.g. in row based APIs) to always be (u)int64 for integers to avoid over-/underflows.

Fixes #1190

### DIFF
--- a/include/soci/soci-backend.h
+++ b/include/soci/soci-backend.h
@@ -294,6 +294,8 @@ public:
     virtual vector_into_type_backend* make_vector_into_type_backend() = 0;
     virtual vector_use_type_backend* make_vector_use_type_backend() = 0;
 
+    virtual db_type exchange_dbtype_for(db_type type) const { return type; }
+
     // These are set when the corresponding make_xxx_backend() above is called
     // by statement_impl. This goes against encapsulation but allows to avoid
     // having to set them in all backends implementations of these functions.
--- a/include/soci/soci-backend.h
+++ b/include/soci/soci-backend.h
@@ -294,6 +294,14 @@ public:
     virtual vector_into_type_backend* make_vector_into_type_backend() = 0;
     virtual vector_use_type_backend* make_vector_use_type_backend() = 0;
 
+    // Converts the given (deduced) DB type and converts it into the data type
+    // that should be used for fetching the corresponding data.
+    // Normally, the deduced DB type should be exactly what should be used but
+    // we need to give backends the possibility to overwrite this to e.g. account
+    // for known shortcomings in the DB type deduction (e.g. for SQLite).
+    // This function is only relevant in cases where we are not dealing with a
+    // user-provided exchange type but instead have to set one up ourselves. Most
+    // notably this is the case for dynamic (i.e. row-based) binding.
     virtual db_type exchange_dbtype_for(db_type type) const { return type; }
 
     // These are set when the corresponding make_xxx_backend() above is called
--- a/src/backends/sqlite3/statement.cpp
+++ b/src/backends/sqlite3/statement.cpp
@@ -614,7 +614,7 @@ sqlite3_statement_backend::make_vector_use_type_backend()
 db_type sqlite3_statement_backend::exchange_dbtype_for(db_type type) const
 {
     // Due to SQLite not really having a type system, any integer type may hold
-    // values that should be way outside of its range.
+    // values that could be way outside of its range.
     // Hence, we have to be prepared to get a huge number, even if the determined
     // db_type is e.g. db_int8.
     // In order to do that, we ensure that we'll always select into an (u)int64,
--- a/src/backends/sqlite3/statement.cpp
+++ b/src/backends/sqlite3/statement.cpp
@@ -610,3 +610,28 @@ sqlite3_statement_backend::make_vector_use_type_backend()
 {
     return new sqlite3_vector_use_type_backend(*this);
 }
+
+db_type sqlite3_statement_backend::exchange_dbtype_for(db_type type) const
+{
+    // Due to SQLite not really having a type system, any integer type may hold
+    // values that should be way outside of its range.
+    // Hence, we have to be prepared to get a huge number, even if the determined
+    // db_type is e.g. db_int8.
+    // In order to do that, we ensure that we'll always select into an (u)int64,
+    // in cases where we have to select the exchange type ourselves (e.g. in rows).
+    switch (type)
+    {
+        case db_int8:
+        case db_int16:
+        case db_int32:
+        case db_int64:
+            return db_int64;
+        case db_uint8:
+        case db_uint16:
+        case db_uint32:
+        case db_uint64:
+            return db_uint64;
+        default:
+            return type;
+    }
+}
--- a/src/core/statement.cpp
+++ b/src/core/statement.cpp
@@ -770,7 +770,7 @@ void statement_impl::describe()
         props.set_db_type(dbtype);
         props.set_data_type(backEnd_->to_data_type(dbtype));
 
-        switch (dbtype)
+        switch (backEnd_->exchange_dbtype_for(dbtype))
         {
         case db_string:
         case db_xml:
--- a/tests/sqlite3/test-sqlite3.cpp
+++ b/tests/sqlite3/test-sqlite3.cpp
@@ -894,6 +894,9 @@ TEST_CASE("SQLite row int64", "[sqlite][row][int64]")
     // small for the queried value, leading to a silent integer overflow and hence unexpected
     // query results.
     // This test case effectively ensures that this no longer happens.
+    //
+    // Note that this test is SQLite-specific because with the other backends we'd
+    // fail to insert the value in the first place.
     soci::session sql(backEnd, connectString);
     integer_table_creator creator(sql);
 
--- a/tests/sqlite3/test-sqlite3.cpp
+++ b/tests/sqlite3/test-sqlite3.cpp
@@ -11,6 +11,9 @@
 
 #include <catch.hpp>
 
+#include <limits>
+#include <cstdint>
+
 using namespace soci;
 using namespace soci::tests;
 
@@ -867,6 +870,31 @@ TEST_CASE("SQLite std::tm bind", "[sqlite][std-tm-bind]")
     REQUIRE(result.size() == 1);
     result.front().tm_isdst = 0;
     CHECK(std::mktime(&result.front()) == std::mktime(&datetime));
+}
+
+struct integer_table_creator : table_creator_base
+{
+    integer_table_creator(soci::session &sql)
+        : table_creator_base(sql)
+    {
+        sql << "create table soci_test(id integer primary key, val integer)";
+    }
+};
+
+TEST_CASE("SQLite row int64", "[sqlite][row][int64]") {
+    soci::session sql(backEnd, connectString);
+    integer_table_creator creator(sql);
+
+    const int id = 1;
+    const std::int64_t val = static_cast<std::int64_t>(std::numeric_limits<std::int32_t>::max()) + 1;
+    sql << "insert into soci_test(id, val) values (:id, :val)", use(id), use(val);
+
+    row r;
+    sql << "SELECT val FROM soci_test WHERE id = 1", into(r);
+    CHECK(r.size() == 1);
+    CHECK(r.get<std::int64_t>("val") == val);
+    // Querying into a int32_t will overflow -> this will throw an exception
+    REQUIRE_THROWS(r.get<std::int32_t>("val"));
 }
 
 // DDL Creation objects for common tests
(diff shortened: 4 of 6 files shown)


src/core/statement.cpp
```diff
@@ -770,7 +770,7 @@ void statement_impl::describe()
         props.set_db_type(dbtype);
         props.set_data_type(backEnd_->to_data_type(dbtype));
 
-        switch (dbtype)
+        switch (backEnd_->exchange_dbtype_for(dbtype))
```
Krzmbrzl, line 773: I'm happy to accept suggestions for a better name for this function :grinning:

tests/sqlite3/test-sqlite3.cpp
```diff
@@ -869,6 +872,31 @@ TEST_CASE("SQLite std::tm bind", "[sqlite][std-tm-bind]")
...
+    const int id = 1;
+    const std::int64_t val = static_cast<std::int64_t>(std::numeric_limits<std::int32_t>::max()) + 1;
+    sql << "insert into soci_test(id, val) values (:id, :val)", use(id), use(val);
+
+    row r;
+    sql << "SELECT val FROM soci_test WHERE id = 1", into(r);
+    CHECK(r.size() == 1);
+    CHECK(r.get<std::int64_t>("val") == val);
+    // Querying into a int32_t will overflow -> this will throw an exception
+    REQUIRE_THROWS(r.get<std::int32_t>("val"));
```
Sildra, line 897: I believe that this breaks production code that relies on handling different integer size.
You are also introducing discrepancies between DDL creation and retrieval stated in the documentation : https://soci.sourceforge.net/doc/master/utilities/#database-metadata
Dynamic bindings may also be affected : https://soci.sourceforge.net/doc/master/types/#dynamic-binding

Spixmaster:
@Krzmbrzl Thank you very much!

If I understand the solution correctly, it is almost like I suggested. For signed values std::int64_t is used and std::uint64_t for unsigned ones.

I have one question. Can I do something like const std::int8_t a = row->get<std::int8_t>(0); if in that case std::int64_t is got from the database?

tests/sqlite3/test-sqlite3.cpp
Krzmbrzl: It only throws if an actual overflow happens. Otherwise, no exception is thrown.

src/backends/sqlite3/statement.cpp
```diff
@@ -610,3 +610,28 @@ sqlite3_statement_backend::make_vector_use_type_backend()
 {
     return new sqlite3_vector_use_type_backend(*this);
 }
+
+db_type sqlite3_statement_backend::exchange_dbtype_for(db_type type) const
+{
+    // Due to SQLite not really having a type system, any integer type may hold
+    // values that should be way outside of its range.
```
vadz, line 617: I believe there is a typo here:
```suggestion
    // values that could be way outside of its range.
```

src/core/statement.cpp
vadz: I think the name is fine, but if you'd prefer, perhaps we should emphasize that this is used for dynamic/row-based API only by calling it `exchange_dbtype_for_row()` or `dynamic_exchange_dbtype_for()` perhaps?

include/soci/soci-backend.h
```diff
@@ -294,6 +294,8 @@ class statement_backend
     virtual vector_into_type_backend* make_vector_into_type_backend() = 0;
     virtual vector_use_type_backend* make_vector_use_type_backend() = 0;
 
+    virtual db_type exchange_dbtype_for(db_type type) const { return type; }
```
vadz, line 297: I know there are few comments in SOCI, but I'd like to change this and have more of them, at least for the new functions. How about something like this?

```suggestion
    // By default we suppose that there is a one-to-one mapping between DB
    // types and the types used in the database, but this may not be the case
    // for some backends (currently only SQLite3, see comment in its code), so
    // provide a way to use different intermediate types when using dynamic
    // (i.e. row-based) API.
    virtual db_type exchange_dbtype_for(db_type type) const { return type; }
```

tests/sqlite3/test-sqlite3.cpp
```diff
@@ -869,6 +872,31 @@ TEST_CASE("SQLite std::tm bind", "[sqlite][std-tm-bind]")
...
+struct integer_table_creator : table_creator_base
+{
+    integer_table_creator(soci::session &sql)
+        : table_creator_base(sql)
+    {
+        sql << "create table soci_test(id integer primary key, val integer)";
+    }
+};
+
+TEST_CASE("SQLite row int64", "[sqlite][row][int64]") {
```
vadz, line 884: Minor style nitpick:
```suggestion
TEST_CASE("SQLite row int64", "[sqlite][row][int64]")
{
```

tests/sqlite3/test-sqlite3.cpp
vadz: I think the (new) behaviour here is correct, but I wonder why are we testing this in SQLite-specific test only, shouldn't it work in the same way for the other backends too, i.e. can we move this test into `common/test-dynamic.cpp`?

And, especially if this is really SQLite-specific, it probably should be documented in the "Dynamic Binding" section of [SQLite3 backend docs](https://soci.sourceforge.net/doc/master/backends/sqlite3/).

vadz (commented):
Thanks for fixing this!

The changes look fine to me, I'd just like to run this test for all backends or, alternatively, document why does this work only for SQLite3 (and, more generally, document any SQLite3-specific behaviour when using dynamic binding).

Thanks again!

Krzmbrzl:
> I wonder why are we testing this in SQLite-specific test only, shouldn't it work in the same way for the other backends too, i.e. can we move this test into common/test-dynamic.cpp?

No. For other backends, it should be impossible to run into this as the DB itself will ensure that when using a 32bit integer type, you can't select a value from it that overflows a 32bit integer. Thus, this test would (should!) cause a DB error on insert.

vadz:
> > I wonder why are we testing this in SQLite-specific test only, shouldn't it work in the same way for the other backends too, i.e. can we move this test into common/test-dynamic.cpp?
> 
> No. For other backends, it should be impossible to run into this as the DB itself will ensure that when using a 32bit integer type, you can't select a value from it that overflows a 32bit integer. Thus, this test would (should!) cause a DB error on insert.

Sorry, I'm not sure I understand this. Why shouldn't it be possible to insert a `INT_MAX + 1` into a `int64` (aka `bigint`) column? AFAICS we could still work that inserting it works, but getting it back as `int32_t` fails, am I missing something?

I'd really like to merge this and finally make the release, but I need to update the documentation and I want to check that I understand things correctly myself before doing it.

Krzmbrzl:
@Spixmaster 
> If I understand the solution correctly, it is almost like I suggested.

Well, it is what I have suggested all along and since you mostly repeated that, yes it is also mostly what you suggested.

> Can I do something like const std::int8_t a = row->get<std::int8_t>(0); if in that case std::int64_t will be got from the database?

No. If you use a narrower type than what SOCI thinks the corresponding C++ type should be, you will get a `bad_cast` exception. You may only use wider types than what SOCI considers appropriate. This is the reason why it is fine to get the value as `int64_t` even though SOCI thinks a `int32_t` is sufficient.
This PR is only about ensuring that _if_ you provide a type wide enough to avoid overflow, you will actually get the result without overflowing. But this is effectively only a workaround for SQLite not enforcing types in the first place.

----

@vadz 
> Why shouldn't it be possible to insert a INT_MAX + 1 into a int64 (aka bigint) column?

That is certainly possible, but this is not what we are doing in this test case. If you use a bigint in the first place, SOCI will deduce the required type to `int64_t` and thus you can never get the value into a `int32_t` as that would be a narrowing conversion.

The point of this test case is that we are using a int32_t (integer) column but SQLite still allows us to insert a value into that that overflows int32_t. And that's what other backends won't allow (rightfully so).

> AFAICS we could still work that inserting it works, but getting it back as int32_t fails, am I missing something?

Yes, but that tests something unrelated. If that is not tested anywhere, it would likely be good to do so though.

include/soci/soci-backend.h
Krzmbrzl: I like the premise. I decided to use a slightly different wording though.

tests/sqlite3/test-sqlite3.cpp
Krzmbrzl: Added the docs :+1:

zann1x:
> > Can I do something like const std::int8_t a = row->getstd::int8_t(0); if in that case std::int64_t will be got from the database?
> 
> No. If you use a narrower type than what SOCI thinks the corresponding C++ type should be, you will get a `bad_cast` exception. You may only use wider types than what SOCI considers appropriate. This is the reason why it is fine to get the value as `int64_t` even though SOCI thinks a `int32_t` is sufficient.

I don't think this statement is correct. SOCI allows conversions to smaller types as well if the value fits inside it. I.e., if you store the value `14` in a column that is mapped to type `std::int64_t`, you should still be able to retrieve it as `std::int8_t`, `std::int16_t` etc. because the value is within the requested type's range.

Krzmbrzl:
> I don't think this statement is correct. 

Indeed. I just checked and what you wrote is entirely correct. Didn't know that. Thanks for correcting me! :+1:

vadz:
> > Why shouldn't it be possible to insert a INT_MAX + 1 into a int64 (aka bigint) column?
> 
> That is certainly possible, but this is not what we are doing in this test case. If you use a bigint in the first place, SOCI will deduce the required type to `int64_t` and thus you can never get the value into a `int32_t` as that would be a narrowing conversion.

I think you have already agreed in the comment following that one that this wasn't true, right?

> > AFAICS we could still work that inserting it works, but getting it back as int32_t fails, am I missing something?
> 
> Yes, but that tests something unrelated. If that is not tested anywhere, it would likely be good to do so though.

We may well have some similar tests already, which is why I'd like to put this one there too, to keep all the related tests together. I think the test should pass, as written, except for using the DB-specific type for big integer instead of the hard-coded `integer`, for the other backends too, but it would be nice to be sure about it, by running this test for all backends and not just SQLite.

Krzmbrzl:
> I think you have already agreed in the comment following that one that this wasn't true, right?

No

> except for using the DB-specific type for big integer instead of the hard-coded integer, for the other backends too, but it would be nice to be sure about it, by running this test for all backends and not just SQLite.

No it's a different test, if you use bigint.

vadz:
OK, thanks, I think I've finally convinced myself that this is the right thing to do, I'll squash merge it soon.

vadz:
Something like this?

```cpp
// This test checks that while we can insert a value not fitting into int32_t
// range into a column that would be created for the int32_t type, because
// SQLite doesn't distinguish between the integers of different sizes, we at
// least get an error if we try to get it back as int32_t.
//
// Note that this test is SQLite-specific because with the other backends we'd
// fail to insert the value in the first place.
```

Krzmbrzl:
Yeah - I chose a different wording. Hope it is still understandable

vadz:
OK, thanks, merged now.

PR #1218, opened 2025-03-22
FindMySQL.cmake fix include paths in vcpkg env

Krzmbrzl:


cmake/find_modules/FindMySQL.cmake
```diff
@@ -28,8 +28,17 @@ if (DEFINED VCPKG_TARGET_TRIPLET)
...
     add_library(MySQL::MySQL ALIAS ${LIBRARY_TARGET})
+
+    get_target_property(INCLUDE_DIRS ${LIBRARY_TARGET} INTERFACE_INCLUDE_DIRECTORIES)
+    if (NOT INCLUDE_DIRS)
+      message(FATAL_ERROR "Expected include paths to be set")
+    endif()
+    foreach(current IN LISTS INCLUDE_DIRS)
+      # vcpkg expects one to include mysql/mysql.h, which goes against how MySQL sets
+      # itself up, so we have to correct for that
+      target_include_directories(${LIBRARY_TARGET} SYSTEM INTERFACE "${current}/mysql")
```
vadz, line 40: Sorry, this doesn't make sense to me: if you need to include `mysql/mysql.h`, then the include directory name should _not_ include the `mysql` path segment, but here we're adding one. Am I missing something or is either the comment or the code wrong?

Krzmbrzl: Neither. The comment is correct in stating that vcpkg sets up the include path so that you need to include `mysql/mysql.h` but we don't do it that way (and neither do other means of obtaining MySQL). Hence, we need to add `mysql` to the include path so that we can include `mysql.h` directly.

vadz: Thanks for the explanation, this was a bit confusing but I see it now.

vadz:
Sorry, just another question: shouldn't we check that the include path doesn't already include `mysql`, i.e. wouldn't we check with `mysql/mysql/mysql` in the include path when the path is already correct?

cmake/find_modules/FindMySQL.cmake
```diff
@@ -28,8 +28,17 @@ if (DEFINED VCPKG_TARGET_TRIPLET)
   endif()
 
   if (${FOUND_VAR})
-    set(MySQL_FOUND TRUE)
```
vadz, line 31: Just to check: is removing this line intentional?

Krzmbrzl:
> i.e. wouldn't we check with mysql/mysql/mysql in the include path when the path is already correct?

yes, but I don't think the check would be worth it because
1. I doubt that vcpkg will change this as this would break all downstream users (and MySQL is by no means a niche, unknown package)
2. Having the wrong (unnecessarily nested) include path _as well_ (we don't remove the provided one) doesn't do any harm.

cmake/find_modules/FindMySQL.cmake
Krzmbrzl: Yes. This variable is set at the bottom of the file by `find_package_handle_standard_args`

cmake/find_modules/FindMySQL.cmake
Krzmbrzl: Let me try to rephrase the comment to make it more clear...

vadz:
> > i.e. wouldn't we check with mysql/mysql/mysql in the include path when the path is already correct?
> 
> yes, but I don't think the check would be worth it because

I thought it would be tidier not to have them, but I agree that it doesn't do any real harm, especially if it's only done for vcpkg (which I hadn't realized when just looking at the diff).

BTW, do we really need the `NOT INCLUDE_DIRS` check? It does no harm too, of course, but I don't see much point in having it too...

Krzmbrzl:
> BTW, do we really need the NOT INCLUDE_DIRS check? It does no harm too, of course, but I don't see much point in having it too...

Yes, because in CMake there are different properties that include dirs could be populated with and I am reading only a single one. Should this be changed, the rest of the code would continue to "work" (as in: no error would occur) but we would get a compile error because the include path is wrong again.

> I thought it would be tidier not to have them

Agreed, but it would involve trying to verify the vcpkg paths and potentially unsetting them. I think this would be rather tricky to achieve and would definitely blow up the complexity of this part of the code, which is why I chose not to do it.

PR #1219, opened 2025-03-23
Include cstdint for uint32_t

da115115:
With recent versions of GNU C++ compilers (for instance on Fedora Rawhide), we need to `#include <cstdint>` in order for `uint32_t` to be defined


vadz:
Yes, this makes sense, thanks.

PR #1221, opened 2025-03-24
Cmake backend include fixes

vadz:
@Krzmbrzl Please let me know if you see anything wrong here, thanks!

### DIFF
--- a/cmake/soci_define_backend_target.cmake
+++ b/cmake/soci_define_backend_target.cmake
@@ -22,7 +22,6 @@ include(soci_utils)
 # FIND_PACKAGE_FILES <file1> [... <fileN>] List of files used by find_package to locate one of the dependencies. Specified
 #                                          files will be installed alongside SOCI in order to be usable from the install tree.
 # HEADER_FILES <file1> [... <fileN>]       List of public header files associated with this backend target.
-# PRIVATE_INCLUDE_DIRS <dir1> [... <dirN>] List of private include directories
 # REQUIRED_COMPONENTS <cmp1> [... <cmpN>]  List of SOCI components (full target names) to link this backend target to
 # SOURCE_FILES <file1> [... <fileN>]       List of source files that shall be part of this backend component
 function(soci_define_backend_target)
@@ -38,7 +37,6 @@ function(soci_define_backend_target)
     "DEPENDENCIES"
     "FIND_PACKAGE_FILES"
     "HEADER_FILES"
-    "PRIVATE_INCLUDE_DIRS"
     "REQUIRED_COMPONENTS"
     "SOURCE_FILES"
   )
@@ -123,7 +121,7 @@ function(soci_define_backend_target)
     target_link_libraries(${DEFINE_BACKEND_TARGET_NAME} PUBLIC "${CURRENT_DEP}")
   endforeach()
 
-  target_include_directories(${DEFINE_BACKEND_TARGET_NAME} PRIVATE ${DEFINE_BACKEND_PRIVATE_INCLUDE_DIRS})
+  target_include_directories(${DEFINE_BACKEND_TARGET_NAME} PRIVATE "${PROJECT_SOURCE_DIR}/include/private")
 
   set_target_properties(${DEFINE_BACKEND_TARGET_NAME}
     PROPERTIES
--- a/cmake/soci_define_backend_target.cmake
+++ b/cmake/soci_define_backend_target.cmake
@@ -22,7 +22,6 @@ include(soci_utils)
 # FIND_PACKAGE_FILES <file1> [... <fileN>] List of files used by find_package to locate one of the dependencies. Specified
 #                                          files will be installed alongside SOCI in order to be usable from the install tree.
 # HEADER_FILES <file1> [... <fileN>]       List of public header files associated with this backend target.
-# REQUIRED_COMPONENTS <cmp1> [... <cmpN>]  List of SOCI components (full target names) to link this backend target to
 # SOURCE_FILES <file1> [... <fileN>]       List of source files that shall be part of this backend component
 function(soci_define_backend_target)
   set(FLAGS "")
@@ -37,7 +36,6 @@ function(soci_define_backend_target)
     "DEPENDENCIES"
     "FIND_PACKAGE_FILES"
     "HEADER_FILES"
-    "REQUIRED_COMPONENTS"
     "SOURCE_FILES"
   )
   cmake_parse_arguments(DEFINE_BACKEND "${FLAGS}" "${ONE_VAL_OPTIONS}" "${MULTI_VAL_OPTIONS}" ${ARGV})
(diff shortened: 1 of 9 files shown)


cmake/soci_define_backend_target.cmake
```diff
@@ -123,7 +121,7 @@ function(soci_define_backend_target)
     target_link_libraries(${DEFINE_BACKEND_TARGET_NAME} PUBLIC "${CURRENT_DEP}")
   endforeach()
 
-  target_include_directories(${DEFINE_BACKEND_TARGET_NAME} PRIVATE ${DEFINE_BACKEND_PRIVATE_INCLUDE_DIRS})
+  target_include_directories(${DEFINE_BACKEND_TARGET_NAME} PRIVATE "${PROJECT_SOURCE_DIR}/include/private")
```
Krzmbrzl, line 124: I think being explicit when calling the function is better than doing this implicitly here. Otherwise, we get back to the CMake structure being somewhat harder to understand

vadz: Sorry, I really don't see any difference between e.g. `target_include_directories(soci_core "${PROJECT_SOURCE_DIR}/include/soci")` and this line. Both define the include paths that should be used for all targets and I have trouble seeing how this one is more implicit than the other one.

OTOH I definitely do see a problem (not a huge one, but still) with repeating the same parameter unnecessarily in all function invocation. I don't know much about CMake, but you wouldn't have a function parameter whose value is always fixed in any other language, so why would you do it here?

vadz: Actually, looking at this function more carefully, I see that it has another parameter, `REQUIRED_COMPONENTS`, which is not only always set to the same value, but can't possibly be ever set to anything else, so I've added a commit removing it too.

I've realized by now that we have very different ideas about how maintainable CMake code looks like and I definitely could be wrong, but I just don't see any possible rationale for using function arguments for something that is inherently fixed.

Krzmbrzl: There is a huge difference. The difference is in visibility. If you want to know how a given backend is set up, you check the CMakeLists.txt file for that backend. And in one version all include directory and dependencies are listed right there and in your version (same as before the CMake refactor) they are not. Hence, the reader of that code doesn't know about them without having to backtrack through the CMake code in order to find the definition of the backend definition function.

This is exactly the reason why the old CMake setup was so difficult to figure out. Everything was happening implicitly and you effectively had to read the entirety of the CMake code in order to be sure to know what is going on.

vadz: I had never had this problem with the old CMake setup and I don't think I've heard anybody else (other than you, of course) complaining about it. The worst problem was the very confusing logic for enabling/detecting backends, but I had absolutely never worried about not being able to see that backend include directories contained `include/private` or that it was difficult to see that backends linked with "core" library and, frankly, I can't even imagine how could either of those be difficult to figure out.

Krzmbrzl: > but you wouldn't have a function parameter whose value is always fixed in any other language, so why would you do it here?

If the function in question was having global side-effects, you might want to keep apparently redundant parameters in there too, if those make those side-effects clear.

Besides, CMake is primarily a build system and not a real programming language per se. Hence, the support for tooling such as language servers etc. is just not there and hence you usually can't quickly jump to the definition of a given CMake macro/function. Thus, having things in-place makes it much easier to understand.

vadz: Also, just for completeness: it's _vastly_ simpler to me to grep for `target_include_directories()` than to figure out that I have to look at `PRIVATE_INCLUDE_DIRS` parameter of the function (especially because it is being actually used as `DEFINE_BACKEND_PRIVATE_INCLUDE_DIRS`, so the usual `grep -w` wouldn't even find this!).

I think the only reason you think the latter is more understandable is that you wrote it and know how it works.

vadz:
It seems like we have irreconcilable difference between our opinions on what is clear and what is not here, so I think the only thing to do is to ask if any others have any good arguments in favour of applying or rejecting this PR.

I'll leave it open to wait for such comments, but will merge it in a few days if there are none.

cmake/soci_define_backend_target.cmake
Krzmbrzl: Grepping is not really a valid approach (in general) to figure out a CMake setup as it misses context and might be returning matches for a completely unrelated target.

I would argue that it is much more natural to start the search for something like this in the CMake file for the component one is interested in...

Krzmbrzl:
@Spixmaster @zann1x any opinions on the matter?

To be clear: the discussion is about whether to have
```
soci_define_backend_target(
...
PRIVATE_INCLUDE_DIRS "include/private"
REQUIRED_COMPONENTS SOCI::Core
)
```
or to omit these parameters and instead always act as if they were set to these values in the definition of the function
```soci_define_backend_target(
...
)
```
when setting up a new backend target, e.g. at
https://github.com/SOCI/soci/blob/c80c9c4f112d36f3e0ffa5fa49788dc52358ff03/src/backends/sqlite3/CMakeLists.txt#L9-L33

Spixmaster:
I cannot contribute to that. I am too stupid for CMake.

Krzmbrzl:
Well - it's about readability and how hard it is to figure out what is going on. So one could argue your lack of deep CMake insights might make you the perfect candidate for giving an opinion on this :D

Spixmaster:
My take on this, I would reject to commits 97e8947ec4b0c75edb7e603287f48de6f3b3fde3 and 21c29ddb008b221d7189ab4763d2140d9f827060.

The variables `PRIVATE_INCLUDE_DIRS` and `REQUIRED_COMPONENTS` fit with the parallel variables. It makes potential change easier and gives a better overview over variables are set and used. It seems like the better long-term solution.

vadz:
While I could live with `PRIVATE_INCLUDE_DIRS`, if the push comes to shove, I'd really like to get rid of `REQUIRED_COMPONENTS`, having any other value for it (or no value) would be a catastrophic mistake. We must prevent the backends from being able to depend on anything else than "core", so the possibility to specify something else using this parameter is not just useless but actively harmful.

zann1x:
I'm on the side of @vadz here. A specific backend with its `DEPENDENCIES` stands for itself and is always based on the "core" module. Specifying this mandatory dependency as `REQUIRED_COMPONENTS` seems totally unnecessary and is just asking to be misused.

I understand the arguments in favor of `PRIVATE_INCLUDE_DIRS`. But I see `include/private` as part of the "core" module, so I would expect it to be available in a backend without explicitly specifying it. The only real use I would see for `PRIVATE_INCLUDE_DIRS` is if we were to distinguish between backend-specific private includes, i.e., `include/private/firebird` would only get included for "firebird". We don't do that though, so `PRIVATE_INCLUDE_DIRS` is unnecessary IMO.

TL;DR: `soci_define_backend_target` simply _adds_ a backend to the underlying "core" module and therefore should only specify the things it adds to SOCI's core. I would accept the PR.

Krzmbrzl:
> having any other value for it (or no value) would be a catastrophic mistake. We must prevent the backends from being able to depend on anything else than "core", so the possibility to specify something else using this parameter is not just useless but actively harmful.

I disagree - at least in the general case. It might very well be feasible to have a common implementation for a group of similar backends (e.g. MySQL and MariaDB) which could share some implementations which however don't belong into the `core` target.

Besides, to catch issues in dependencies we have have CI. Any PR that messes this up won't pass CI. So I don't think having the possibility of messing things up is an issue. I would say it is in the same ball park as having the possibility to commit code with a syntax error.

---

> soci_define_backend_target simply adds a backend to the underlying "core" module and therefore should only specify the things it adds to SOCI's core. I would accept the PR.

If you think of this function in this way, I would tend to agree with the changes. However, I see this function to serve a more general purpose in the sense that it defines a new target and takes care of the boilerplate required to make installation work properly. And if one checks the implementation, you'll see that it indeed this is the vast majority of the code.

I guess this is the root of the disagreement: the interpretation of what exactly the scope of this function is...

vadz:
Catching errors in the CI is good, but not allowing to even make them is better. And it's not guaranteed that you catch errors due to misuse of this function in the CI, of course.

But I agree that the root cause is that I see this function as a very specialized helper for avoiding boilerplate when writing SOCI backends makefiles, while you seem to think of it as a very general mechanism which could be used in the ways we don't use it yet (and, I'd add, don't even plan to). I'd definitely like to keep my build system, of all things, as simple as possible and removing unwanted generality helps with this, so I'll merge this.

PR #1223, opened 2025-03-27
[MySQL] Works around a false positive by LTO

da115115:
Works around a false positive `-Wuninitialized` error exposed by LTO

That is a patch I had to create on Fedora: https://src.fedoraproject.org/rpms/soci/blob/rawhide/f/soci-uninit.patch


vadz:
Thanks, pushing and will cherry-pick to 4.1 branch too.

PR #1227, opened 2025-04-09
Add back CMake `WITH_BOOST` option

vadz:
cc @Krzmbrzl


vadz:
Even Copilot doesn't want to review CMake :-/

CMakeLists.txt
```diff
@@ -42,6 +42,7 @@ endif()
 
 option(SOCI_SHARED "Enable building SOCI as a shared library" ${SHARED_DEFAULT})
 option(SOCI_TESTS "Enable building SOCI test cases" ${PROJECT_IS_TOP_LEVEL})
+option(WITH_BOOST "Enable Boost support" ON)
```
Krzmbrzl, line 45: I would recommend a name more in line with the new CMake options such as `SOCI_BOOST`. Then add some logic to the compat cmake file that sets the value of `SOCI_BOOST` if `WITH_BOOST` is defined.

Furthermore, I don't think the semantics of this option are quite clear. If we have a ON/OFF option, I would expect an error if it is set to `ON` but Boost is not found. However, it seems like the semantics are more akin to the `AUTO` mode for the backends?

An alternative would be to instead have an option `SOCI_DISABLE_BOOST` that leaves the current auto-detect logic if set to `OFF` and disables if set to `ON`

Krzmbrzl: Okay, fair enough

vadz:
BTW, have you seen that we're getting

```
CMake Warning at cmake/soci_utils.cmake:156 (find_package):
  By not providing "FindBoost.cmake" in CMAKE_MODULE_PATH this project has
  asked CMake to find a package configuration file provided by "Boost", but
  CMake did not find one.
  Could not find a package configuration file provided by "Boost" with any of
  the following names:
    BoostConfig.cmake
    boost-config.cmake
  Add the installation prefix of "Boost" to CMAKE_PREFIX_PATH or set
  "Boost_DIR" to a directory containing one of the above files.  If "Boost"
  provides a separate development package or SDK, be sure it has been
  installed.
Call Stack (most recent call first):
  src/core/CMakeLists.txt:81 (soci_public_dependency)
```

in the AppVeyor CI builds? Should we add `FindBoost.cmake` to `cmake/find_modules`?

Krzmbrzl:
No. That's a normal message if Boost is not found. We could add `QUIET` to the `find_package` calls to get rid of the message. 

Boost comes with a dedicated CMake config file these days that makes custom find modules superfluous (even before, a corresponding find package was bundled with CMake so you never had to write your own).

If Boost is in fact installed on the CI machine, we might have to set `CMAKE_PREFIX_PATH` when calling CMake so that CMake can find Boost's config file.

vadz:
> No. That's a normal message if Boost is not found.

Are you sure? For me the normal message is something like this:

```
-- Could NOT find boost_date_time (missing: boost_date_time_DIR)
```

i.e. the message given by `FindBoost` macro/function, not the message above about not being able to find it in the first place.

> We could add `QUIET` to the `find_package` calls to get rid of the message.

I don't think it's desirable to get rid of the messages from `FindBoost`.

> Boost comes with a dedicated CMake config file these days that makes custom find modules superfluous (even before, a corresponding find package was bundled with CMake so you never had to write your own).

Well, it looks like it's not bundled with CMake 4.0...

Krzmbrzl:
> For me the normal message is something like this:

That's when Boost is installed and found in general but the `date_time` module in particular is not found. The other message is if Boost is not found at all.

> Well, it looks like it's not bundled with CMake 4.0...

Yes, because recent Boost version provide a CMake config file, making the find module superfluous.

Krzmbrzl:
> AFAICS, prevents people from installing Boost manually, i.e. just unpacking the ZIP somewhere, and providing paths to it for CMake to use, doesn't it?

I don't see why this should be the case.

> but it's weird to think that we intentionally prefer this behaviour to the old one

Well, we now do it the way CMake intends. Everyone using CMake regularly will know what to do with that :shrug:

vadz:
@Krzmbrzl I'd just like to return to this because using Boost in AppVeyor CI builds has been broken by the great CMake rewrite and I have no idea about how to fix it without adding back `FindBoost.cmake`. Previously setting `BOOST_ROOT` was sufficient to find Boost installed in AppVeyor images, but now it doesn't work any more.

I don't care much about Boost myself, but we need to either fix or disable/deprecate using Boost with SOCI entirely.

Krzmbrzl:
You probably have to define `CMAKE_PREFIX_PATH` to include the directory (tree) into which Boost is installed. So I think setting `CMAKE_PREFIX_PATH=C:\Libraries\boost_1_60_0` is likely sufficient. See also https://cmake.org/cmake/help/latest/variable/CMAKE_PREFIX_PATH.html

vadz:
What I don't understand is how did it work before, as I can't find which change broke it. How did it detect Boost with just `BOOST_ROOT` set? Unless it never worked at all, but this seems unlikely...

Krzmbrzl:
`BOOST_ROOT` is a variable specific to the FindBoost script so once we stopped using that script this variable became mostly meaningless afaik.

vadz:
But where did we "stopped using that script" exactly? All I see is 0d85e899 (First draft of new cmake, 2024-01-03) which removed `cmake/dependencies/Boost.cmake`, but it just used `find_package(Boost)`, which is pretty similar to what we do now.

Krzmbrzl:
The script was packaged with CMake but given that Boost now comes with a proper CMake?config file the FindBoost script was removed. I think this happened somewhere around 3.20 or so.

PR #1228, opened 2025-04-09
Extend forward compatibility to v4.0 of CMake

vadz:
This is required to avoid build errors about CMake < 3.5 not being supported any longer.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -1,5 +1,5 @@
 # Note: we need >= v3.23 in order to make use of file sets for header installation
-cmake_minimum_required(VERSION 3.23...3.31 FATAL_ERROR)
+cmake_minimum_required(VERSION 3.23...4.0 FATAL_ERROR)
 
 set(CMAKE_MODULE_PATH "${CMAKE_CURRENT_SOURCE_DIR}/cmake")
 list(APPEND CMAKE_MODULE_PATH "${CMAKE_CURRENT_SOURCE_DIR}/cmake/find_modules")


vadz:
@Krzmbrzl Is this the right way to avoid 

```
CMake Error at CMakeLists.txt:9 (cmake_minimum_required):
  Compatibility with CMake < 3.5 has been removed from CMake.
```

?

Krzmbrzl:
Yes. Though with our setup, we shouldn't run into this specific issue any time soon even without the change

vadz:
> Yes. Though with our setup, we shouldn't run into this specific issue any time soon even without the change

Err. Not only we risk running into this, but we already did, which is why I made this change. But it doesn't help, we still get the error, see e.g. [this build](https://ci.appveyor.com/project/SOCI/soci/builds/51857541/job/als8km34tpc1qf13).

I guess I'll have to build my own CMake 4.0 as it's not in Debian yet (well, only in experimental...) to see how to avoid it.

Krzmbrzl:
The error is not from SOCI but from SQLite which we build ourselves on the Windows CI. Hence the error persisting after changing the SOCI files. 

To fix it, the CMakeLists.txt file of SQLite needs to be patched. Presumably upgrading to a newer SQLite release would solve the issue as well (and with less effort)

vadz:
Oops, you're right, of course, sorry for not noticing this.

Unfortunately we use some [non-official SQLite version](https://github.com/snikulov/sqlite.cmake.build.git) with CMake tackled on top of it for the AppVeyor builds, so updating it is not simple.

I'd like to add SQLite as submodule to SOCI and add support for building it as part of it, as this would ensure that we always have at least one functional backend. I'll try to do this instead...

PR #1229, opened 2025-04-10
Use built-in SQLite if system library not found

vadz:
In particular, this fixes AppVeyor CI build as it didn't work with CMake 4.0 any more.

### DIFF
--- a/include/soci/sqlite3/soci-sqlite3.h
+++ b/include/soci/sqlite3/soci-sqlite3.h
@@ -31,7 +31,24 @@
 namespace sqlite_api
 {
 
-#include <sqlite3.h>
+// Don't include sqlite3.h from outside SOCI: this header might not be
+// available when using built-in SQLite3 as we don't install it in this case.
+#ifdef SOCI_SQLITE3_SOURCE
+    #include <sqlite3.h>
+#else // !SOCI_SQLITE3_SOURCE
+// We need just a couple of forward declarations to make this header itself
+// compile.
+struct sqlite3;
+struct sqlite3_stmt;
+
+#if defined(_MSC_VER)
+  typedef __int64 sqlite3_int64;
+  typedef unsigned __int64 sqlite3_uint64;
+#else
+  typedef long long int sqlite3_int64;
+  typedef unsigned long long int sqlite3_uint64;
+#endif
+#endif // SOCI_SQLITE3_SOURCE/!SOCI_SQLITE3_SOURCE
 
 } // namespace sqlite_api
 
--- a/include/soci/sqlite3/soci-sqlite3.h
+++ b/include/soci/sqlite3/soci-sqlite3.h
@@ -353,6 +353,10 @@ struct SOCI_SQLITE3_DECL sqlite3_session_backend : details::session_backend
         }
     }
 
+    // Get information about SQLite3 version used.
+    static const char* libversion();
+    static int libversion_number();
+
     sqlite_api::sqlite3 *conn_;
 
     // This flag is set to true if the internal sqlite_sequence table exists in
--- a/include/soci/sqlite3/soci-sqlite3.h
+++ b/include/soci/sqlite3/soci-sqlite3.h
@@ -31,18 +31,9 @@
 namespace sqlite_api
 {
 
-#if SQLITE_VERSION_NUMBER < 3003010
-// The sqlite3_destructor_type typedef introduced in 3.3.10
-// https://www.sqlite.org/cvstrac/tktview?tn=2191
-typedef void (*sqlite3_destructor_type)(void*);
-#endif
-
 #include <sqlite3.h>
 
 } // namespace sqlite_api
-
-#undef SQLITE_STATIC
-#define SQLITE_STATIC ((sqlite_api::sqlite3_destructor_type)0)
 
 #ifdef _MSC_VER
 #pragma warning(pop)
--- a/src/backends/sqlite3/CMakeLists.txt
+++ b/src/backends/sqlite3/CMakeLists.txt
@@ -1,7 +1,9 @@
 include(soci_define_backend_target)
 
 if (SOCI_SQLITE3_AUTO)
-  set(DEPENDENCY_MODE "DISABLE")
+  # Unlike most other backends, we can fall back on using the built-in
+  # version of SQLite3 if the system version is not found.
+  set(DEPENDENCY_MODE "BUILTIN")
 else()
   set(DEPENDENCY_MODE "ERROR")
 endif()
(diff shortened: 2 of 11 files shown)


src/backends/sqlite3/CMakeLists.txt
```diff
@@ -1,7 +1,9 @@
 include(soci_define_backend_target)
 
 if (SOCI_SQLITE3_AUTO)
```
vadz, line 3: @Krzmbrzl Somehow this test doesn't seem to work when rerunning CMake. I.e. it works fine when running it the first time or with `--fresh` later but not when rerunning it without that option, including when `cmake --build` reruns it automatically. I've tried to understand what's going on here, but the mechanics of defining `${BACKEND}_AUTO` are too confusing, so I'd be grateful if you could please have a look at this. TIA!

vadz: OK, so this seems to happen because we intentionally override the cached variable with `ON`. I don't think we need to do this, so I'll create a separate PR to stop doing it.

include/soci/sqlite3/soci-sqlite3.h
```diff
@@ -31,19 +31,27 @@
 namespace sqlite_api
 {
 
-#if SQLITE_VERSION_NUMBER < 3003010
-// The sqlite3_destructor_type typedef introduced in 3.3.10
-// https://www.sqlite.org/cvstrac/tktview?tn=2191
-typedef void (*sqlite3_destructor_type)(void*);
+// Don't include sqlite3.h from outside SOCI: this header might not be
+// available when using built-in SQLite3 as we don't install it in this case.
```
Krzmbrzl, line 35: Not installing SQLite most likely means that the install tree will be broken as the installed SOCI version will check for required packages and SQLite won't be found in that case.

Do we statically link SQLite when using the builtin version? If so, the SQLite library must not be added to the public dependencies of soci_sqlite3. Plus, we must either install the header file(s) or make sure to not include any SQLite header when using the installed SOCI lib.

vadz: I don't see why installing SOCI should provide SQLite, this has never been in the contract (just as installing it doesn't provide PostgreSQL or Oracle or whatever). And I think it would be wrong to install a header of a different library as part of SOCI installation.

`soci_sqlite3_builtin` is a static library and is a `PRIVATE` dependency of `soci_sqlite3`, so I think this is already fine.

Krzmbrzl: > I don't see why installing SOCI should provide SQLite

Well, that's exactly what it is doing now, isn't it?

> soci_sqlite3_builtin is a static library and is a PRIVATE dependency of soci_sqlite3, so I think this is already fine.

No, the CMake system is set up to populate dependencies (the "public deps" ) to the config file. The way you have set up SQLite, a dependency is still created and therefore when using SOCI's SQLite backend from the install tree will almost certainly look for the SQLite library. The way I understood things, this is no longer intended, is it? :thinking:

vadz: > > I don't see why installing SOCI should provide SQLite
> 
> Well, that's exactly what it is doing now, isn't it?

No, of course not. Why do you think it does? It uses the existing system SQLite, it never installs it.

> > soci_sqlite3_builtin is a static library and is a PRIVATE dependency of soci_sqlite3, so I think this is already fine.
> 
> No, the CMake system is set up to populate dependencies (the "public deps" ) to the config file. The way you have set up SQLite, a dependency is still created and therefore when using SOCI's SQLite backend from the install tree will almost certainly look for the SQLite library. The way I understood things, this is no longer intended, is it? 🤔

Would it populate it even if the dependency is not found? It shouldn't...

But I have no idea how to test this, and our CI jobs don't test it either, AFAICS, do they?

Krzmbrzl: > No, of course not. Why do you think it does? It uses the existing system SQLite, it never installs it.

Doesn't this PR add SQLite directly into SOCI itself rather than requiring it to be installed on the target system? It not being installed, I see. But I don't see how providing a built-in SQLite is not the same as SOCI providing SQLite (in cases in which it doesn't exist on the system). :eyes: 

> Would it populate it even if the dependency is not found? It shouldn't...

If the dependency is not found the system disables the entire backend, meaning it won't get installed anyway. But this semantic is now changed with the introduction of builtin support for SQLite, which can make the backend enablement succeed even if the dependency was not found.

> But I have no idea how to test this, and our CI jobs don't test it either, AFAICS, do they?

It can be tested by installing SOCI via `make install` into a given directory and then consuming it via CMake from that directory (explicitly requesting the `SQLite3` component). I'll do that locally and report back what I find.
We do this on CI, but only with the `Empty` component
https://github.com/SOCI/soci/blob/e337a092d49641f2ba29ccaa7eb79023d0819641/scripts/ci/build_empty.sh#L32-L33

src/backends/sqlite3/CMakeLists.txt
```diff
@@ -32,3 +34,31 @@ soci_define_backend_target(
 if (NOT SOCI_SQLITE3)
   return()
 endif()
+
+# If SQLite3 was found, this target must have been created.
+if (NOT TARGET SQLite::SQLite3)
+  # But if it wasn't, use our built-in version, after checking that it is
+  # available.
+  if (NOT EXISTS "${PROJECT_SOURCE_DIR}/3rdparty/sqlite3/sqlite3.c")
+    message(WARNING "SQLite3 not found and built-in version not available, have you cloned SOCI repository with --recurse-submodules?")
```
Krzmbrzl, line 43: Shouldn't this be a error rather than a warning? After all, if we reach this point we have enabled the backend which so far does not have the required link libraries configured. And if we don't have the submodule, this will guarantees compile and/or link errors, doesn't it?

include/soci/sqlite3/soci-sqlite3.h
Krzmbrzl: Okay turns out that by using the builtin SQLite option, we never call `soci_public_dependency` which in turn means that the CMake part dealing with auto-generating the SOCI config file never gets to know about this component. Hence, even though we install `soci_sqlite` it won't be discoverable by downstream users via the config file. In other words: the current system builds on the assumption that all installed components (except `Core`) have at least one public dependency.

While this incidentally solves the issue I was pointing out, I think we should be more intentional about it being installed or not. Installing the library but not have it usable from the install tree doesn't seem like a desirable situation. If we don't want to install SQLite along with SOCI (which seems reasonable), we should not set the SQLite backend up for installation when using the builtin SQLite library (and add a CMake warning that will tell the user that this is what's going to happen).

----

I have found another issue with the builtin version:  if we build SOCI as a static library, we can't make SQLite a private dependency as static libraries won't get linked into other static libraries. This means that `libsoci_sqlite3.a` will claim to have no dependencies for downstream users but will lead to linker errors as
```
$ nm --undefined-only libsoci_sqlite3.a
...
session.cpp.o:
...
                 U sqlite3_busy_timeout
                 U sqlite3_close
                 U sqlite3_errmsg
                 U sqlite3_exec
                 U sqlite3_free
                 U sqlite3_libversion
                 U sqlite3_libversion_number
                 U sqlite3_open_v2
...
```
that is, we still need to link to the SQLite library when linking to `libsoci_sqlite3.a`.

vadz:
There is indeed a problem when using static SOCI libraries.

The simplest way to solve all this would be to avoid having a separate SQLite library and just compile SQLite as part of SOCI SQLite backend itself, and this is what I had initially tried to do. Unfortunately this resulted in compilation problems due to `soci_compiler_interface` being used by the backend: the options there are not suitable for compiling C SQLite code. If we could restrict all these options (or even the use of `soci_compiler_interface` entirely) to C++ code only, this would solve all problems — can we do this somehow?

Krzmbrzl:
> can we do this somehow?

you can keep the scope of all definitions local to any given target. So as long as we retain different targets for C++ and C code, this should be achievable, yes.

vadz:
Any hints about how can I make `target_compile_options(soci_compiler_interface)` apply only to C++ code? TIA!

Krzmbrzl:
I think it would only work by only linking C++ targets to `soci_compiler_interface`. I don't think there are language-specific compile options for a single target.

vadz:
Hmm, wait, maybe I can just use `set_source_files_properties()` to _replace_ the C++ flags with C ones for this file? Let me try it...

Krzmbrzl:
I would have recommended building SQLite as a separate target, build the SOCI lib as a seaparate target and then link both of them together into a single target.

Using per source file properties would be an alternative, I guess.

vadz:
For the record, my actual attempt to fix it is in #1243.

PR #1231, opened 2025-04-11
CMake fixes

vadz:
@Krzmbrzl Please let me know if you see anything wrong here.


Krzmbrzl:
Yes, this (https://github.com/SOCI/soci/pull/1231/commits/3a82b1ab88c7a257ef693a6d217b412aba0d7634) breaks the system. Other parts of the CMake code expect these values to be either `ON` or `OFF` and are not designed to deal with `AUTO`.

The resetting is done intentionally for the following reasons
- It allows to have the special handling of the different possible values in only a single place instead of everywhere where we need information on whether or not a given backend is enabled or not
- Prevents getting the same "dependency not found, disabling backend" warnings every time you re-run CMake. I don't think this is desirable as using `AUTO` means that the user is okay with the backend being disabled if dependencies are not found. Otherwise, they would have used `ON`. So I don't see value in creating additional noise in the CMake output in subsequent CMake runs

vadz:
> Yes, this ([3a82b1a](https://github.com/SOCI/soci/commit/3a82b1ab88c7a257ef693a6d217b412aba0d7634)) breaks the system. Other parts of the CMake code expect these values to be either `ON` or `OFF` and are not designed to deal with `AUTO`.

Where exactly does this happen? I.e. how can I find what is broken by this change and fix it?
 
> The resetting is done intentionally for the following reasons
> 
>     * It allows to have the special handling of the different possible values in only a single place instead of everywhere where we need information on whether or not a given backend is enabled or not

This could be achieved by having a normal, non-cache variable storing the result of the check. Although I'd say that using `if (SOCI_XXX)` doesn't seem too complicated to me.
 
>     * Prevents getting the same "dependency not found, disabling backend" warnings every time you re-run CMake. I don't think this is desirable as using `AUTO` means that the user is okay with the backend being disabled if dependencies are not found. Otherwise, they would have used `ON`. So I don't see value in creating additional noise in the CMake output in subsequent CMake runs

It's not about the additional noise, but about the very unintuitive behaviour. I definitely do expect the checks to happen again when rerunning CMake with "AUTO" value, e.g. it's common for me to run cmake to configure some package, get an error about missing dependency, install this dependency and rerun cmake again. Having to add `--fresh` is very user-unfriendly (assuming you even know that you need to do this in the first place).

So I absolutely want to avoid overriding the cache variable. If this is not the right way to do it, can you please recommend a better one?

Krzmbrzl:
> I.e. how can I find what is broken by this change and fix it?

Search for every occurrence of those variables and check whether the logic still works. I don't know of a better way, unfortunately.

> Although I'd say that using if (SOCI_XXX) doesn't seem too complicated to me.

Which is exactly what you must not do. The string `AUTO` will always evaluate to `true` in CMake conditionals, but if the backend dependencies weren't actually found this will take the wrong code path.

> I definitely do expect the checks to happen again when rerunning CMake with "AUTO" value, e.g. it's common for me to run cmake to configure some package, get an error about missing dependency, install this dependency and rerun cmake again.

Hm, right. I only considered the use case of either having or not having a given dependency on your system already.

> This could be achieved by having a normal, non-cache variable storing the result of the check.

In theory yes, but in practice we need this information to be transported across different directory levels (including upwards) which doesn't work with normal variables. We have
https://github.com/SOCI/soci/blob/e337a092d49641f2ba29ccaa7eb79023d0819641/src/backends/CMakeLists.txt#L78
which kinda serves this purpose, but not quite (it doesn't give direct per-backend information of whether or not it is enabled).

So I guess we'd need a separate cache variable to set instead. Ideally with a name that makes it obvious that it is supposed to be used for all checks internally instead of `SOCI_XXXXX`

vadz:
> > I.e. how can I find what is broken by this change and fix it?
> 
> Search for every occurrence of those variables and check whether the logic still works. I don't know of a better way, unfortunately.

I've already done this, of course, but didn't find anything that would be broken.

Of course, grepping with CMake is almost useless because variable names are often constructed dynamically :-(
 
> So I guess we'd need a separate cache variable to set instead. Ideally with a name that makes it obvious that it is supposed to be used for all checks internally instead of `SOCI_XXXXX`

`SOCI_FOUND_XXX`?

Krzmbrzl:
> SOCI_FOUND_XXX ?

Maybe `SOCI_XXX_ENABLED` to better encode the exact meaning of the variable? :thinking:

vadz:
I've started making changes, but it turns out that `SOCI_XXX_ENABLED` is not a great name because it conflicts with `CURRENT_ENABLED_VAR` (which has different meaning) in `src/backends/CMakeLists.txt`.

So I've finally did the simplest possible thing in https://github.com/SOCI/soci/pull/1231/commits/9ecef4e7f9c514768b890acd04b6ad28902f04eb and just forced preserving of the originally specified value. This avoids the need for the new variable and still behaves correctly.

I've also simplified things further as I just don't see any point in having extra variables which _must_ have some fixed values, this is much more confusing than flexible...

vadz:
@Krzmbrzl Can I merge this?

PR #1233, opened 2025-04-16
Simplify Boost dependency

vadz:
Don't use `soci_public_dependency()` for it to make things much simpler.

@Krzmbrzl I couldn't deal with the subtleties of this function and all the levels of indirection between its call in the sources and the use of macro names in the generated config file, so I've simply removed all of this because I really, really don't want to have to deal with all that complexity if this is at all avoidable. If we absolutely need to define Boost targets in the config file, I prefer to add something to `soci-config.cmake.in` to do it manually.

### DIFF
--- a/cmake/soci_define_backend_target.cmake
+++ b/cmake/soci_define_backend_target.cmake
@@ -105,7 +105,7 @@ function(soci_define_backend_target)
         endif()
       endforeach()
       list(APPEND PUBLIC_DEP_CALL_ARGS
-        "NAME ${CURRENT_DEP} DEP_TARGETS ${CURRENT_DEP_TARGETS} TARGET SOCI::${DEFINE_BACKEND_ALIAS_NAME} REQUIRED"
+        "NAME ${CURRENT_DEP} DEP_TARGETS ${CURRENT_DEP_TARGETS} TARGET SOCI::${DEFINE_BACKEND_ALIAS_NAME}"
       )
     endif()
   endforeach()
--- a/cmake/soci_define_backend_target.cmake
+++ b/cmake/soci_define_backend_target.cmake
@@ -13,11 +13,9 @@ include(soci_utils)
 # TARGET_NAME <target>                     Name of the CMake target that shall be created for this backend
 # DEPENDENCIES <spec1> [... <specN>]       List of dependency specifications. Each specification has to be a single
 #                                          argument (single string) following the syntax
-#                                          <find_spec> YIELDS <targets> [DEFINES <macros>]
+#                                          <find_spec> YIELDS <targets>
 #                                          where <find_spec> will be passed to find_package to find the dependency. Upon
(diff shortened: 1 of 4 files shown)


Krzmbrzl:
I think this is a bad idea. This now requires all dependencies at all levels to be duplicated in the CMake files and also in the config file.

The function you removed is quite easy to use: use it for public dependencies, that is all dependencies that the final libraries have.

If you prefer a manual approach, I think you must test consumption from the install tree via CI for **all** backends. Then if someone forgets to duplicate the dependencies in the config file, at least CI fails.

However, I don't think that the config file can be simplified a lot due to the need to take care of transient dependencies. So all in all, I don't see a huge simplification in removing what I implemented :shrug:

vadz:
I think we can/should keep this function for the backends, although perhaps it should be in `cmake/soci_define_backend_target.cmake` itself instead of `cmake/soci_utils.cmake`. But Boost is a sufficiently different use case to not warrant shoehorning it into the same framework.

vadz:
I'm not sure what to do about the CI error, it's just due to the "empty" command not having any dependencies (note that this is not directly related to this PR, it also happens on master with `WITH_BOOST=OFF`, or if Boost is not found at all). We could either test for the lists being empty explicitly, or add a dummy dependency to "empty" to make it more like the other backends.

Any thoughts?

Krzmbrzl:
I see. Yeah, I agree that the Boost limbo makes things much more complicated. If we simply required `date_time` if Boost is used, then the code would become much simpler.

As for the issue with the Empty backend: I'll need to look into that once I'm back at my computer. Probably sometime over Easter :)

vadz:
I think I've found a relatively decent way to fix the problem when there are no dependencies, but there is indeed a problem with Boost when it's installed, so I'll have to do something about this.

> This now requires all dependencies at all levels to be duplicated in the CMake files and also in the config file.

I wonder how do the other projects deal with this. For example, [wxWidgets CMake config](https://github.com/wxWidgets/wxWidgets/blob/master/build/cmake/wxWidgetsConfig.cmake.in) doesn't seem to contain any of the dependencies that are used by the various libraries, how does it achieve this?

vadz:
I think this is ready to be merged and I'll do it soon unless you have some strong objections to it.

Krzmbrzl:
> For example, [wxWidgets CMake config](https://github.com/wxWidgets/wxWidgets/blob/master/build/cmake/wxWidgetsConfig.cmake.in) doesn't seem to contain any of the dependencies that are used by the various libraries, how does it achieve this?

They do:
https://github.com/wxWidgets/wxWidgets/blob/9495129c5eda0b0a1147cf90f122408787dd4fde/build/cmake/wxWidgetsConfig.cmake.in#L142-L146

The information which library depends on what other things is auto-generated by CMake in the `*Targets` files but these dependencies must have been resolved by the time these files are included.

PR #1243, opened 2025-04-24
Improve using bundled SQLite

vadz:
@Krzmbrzl I've finally done it like this as it doesn't seem possible to override the inherited properties using `set_source_files_properties()` (the documentation says that it overrides the existing properties, but after checking, it only overrides the properties defined in this target itself).

This seems to work fine, but please let me know if you see anything wrong.


Krzmbrzl:
Sorry for the late response.

Seems reasonable. I feel like there should be a better way of achieving it but I don't have the time to dig into that. Anyway, as long as it works I think this should be good enough :+1:

PR #1246, opened 2025-04-25
Don't use GLOB for the header files

vadz:
This results in an extra "GLOB verification" step when building and just doesn't seem worth it, especially as we already list all the source files and backend-specific headers explicitly.

This also fixes an annoying "GLOB mismatch" error which happened when using Vim as its swap files (".foo.h.swp") were wrongly matched by the GLOB wildcard and so switching to another file in Vim changed the GLOB expansion.

----

@Krzmbrzl FYI.


Krzmbrzl:
Only issue I can see is that forgetting to add a new header there will only be noticed in form of a broken install by downstream users. The only way to detect this would be to create a CI job that consumes the installed library and somehow uses _all_ available header files to see whether they have been installed :eyes:

vadz:
The problem with this is that a test including all headers would need to be updated after adding a new header too, and if we remember to update it, we presumably would remember to update the makefile as well.

Krzmbrzl:
Yep, I agree

PR #1248, opened 2025-05-06
Remove ABI_VERSION from Windows build to prevent including version in DLL name and subsequent crash

rolandhill:
The DLL name dynamically generated on Windows when loading a backend using a backend name (eg 'sqlite3') includes the version (eg soci-sqlite3_4_1.dll). The build system no longer uses version numbers so your application crashes at run time when the DLL cannot be found. 

This PR removes the version number. Tested on Windows and Linux for SQLite3 backend only and only opening via the backend name.

Note that soci_sqlite3_test.exe works before and after this PR because it doesn't load the SQLite3 dll via a backend name.

### DIFF
--- a/src/core/CMakeLists.txt
+++ b/src/core/CMakeLists.txt
@@ -164,9 +164,7 @@ target_link_libraries(soci_core
     ${CMAKE_DL_LIBS}
 )
 
-if (WIN32)
-  set(ABI_VERSION "${PROJECT_VERSION_MAJOR}_${PROJECT_VERSION_MINOR}")
-elseif(UNIX)
+if(UNIX)
   # Use SOVERSION, which is only the major version
   set(ABI_VERSION "${PROJECT_VERSION_MAJOR}")
 endif()
--- a/tests/sqlite3/test-sqlite3.cpp
+++ b/tests/sqlite3/test-sqlite3.cpp
@@ -44,6 +44,7 @@ TEST_CASE("SQLite connection string", "[sqlite][connstring]")
     params.set_option("nocreate", "1");
     CHECK_THROWS_WITH(soci::session(params),
                       Catch::Contains("Cannot establish connection"));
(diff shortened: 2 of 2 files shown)


vadz:
Oh, and I forgot to ask: would it be possible to add a unit test checking for this, i.e. loading the backend by its name dynamically, to ensure that this doesn't get broken again?

vadz:
I'm finally going to release 4.1.1 without this because I'm still not sure to understand what the problem is.

rolandhill:
Apologies, I got caught up in work. I've created a PR that will demonstrate this issue using soci_sqlite3_test (https://github.com/SOCI/soci/pull/1250).

Apply that PR first and run the test. You will see that the test fails on Windows and throws a message "Failed to find shared library soci_sqlite3_4_1.dll". The build system doesn't create this file. It should be soci_sqlite3.dll. All tests pass on Linux, this is only Windows to my knowledge.

Apply PR https://github.com/SOCI/soci/pull/1250 and all tests will pass. It also then works in my application as a real-world test.

vadz:
Thanks, I see the problem now.

Unfortunately I also see why I had failed to detect what changed before: nothing did, in the sources, it's just that the DLLs are created with wrong names now :-(

E.g. in 4.0 we had `soci_core_4_0.dll` and `soci_odbc_4_0.dll` by default, but now we have just `soci_core.dll` and `soci_odbc.dll`. @Krzmbrzl I don't think this was an intentional change, was it?

I think we need to restore it, there was even a comment in the now removed `soci_target_output_name()` from `cmake/SociUtilities.cmake` saying that we intentionally encode ABI version into the library name, so I've created #1252 fixing it and will close this because, again, IMO it's not the right thing to do, even if there is indeed a real problem, so thanks for pointing it out (and I definitely should have waited a bit more with 4.1.1 release...).

Krzmbrzl:
> @Krzmbrzl I don't think this was an intentional change, was it?

No, it was not 👀

PR #1250, opened 2025-05-14
Add unit test to ensure Sqlite3 db can be created using the database name

rolandhill:
This test attempts to open a Sqlite3 database by name ("sqlite3") rather than by specifying soci::factory_sqlite3(). It demonstrates a regression in recent versions where the name of the sqlite3 DLL on Windows is incorrectly generated to include the version, which then causes a crash because the build system no longer creates a DLL with the version included.

See https://github.com/SOCI/soci/pull/1248 for a proposed solution.

### DIFF
--- a/tests/sqlite3/test-sqlite3.cpp
+++ b/tests/sqlite3/test-sqlite3.cpp
@@ -44,6 +44,7 @@ TEST_CASE("SQLite connection string", "[sqlite][connstring]")
     params.set_option("nocreate", "1");
     CHECK_THROWS_WITH(soci::session(params),
                       Catch::Contains("Cannot establish connection"));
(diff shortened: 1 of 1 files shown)


vadz:
We need to test for `SOCI_SHARED` somehow but we don't currently have any way to do it under Unix (under Windows we can check for `SOCI_DLL`). Not sure what would be the best way to handle this, maybe define some `SOCI_TEST_SHARED` preprocessor symbol at CMake level just for the tests?

rolandhill:
I'm not sure how that can be achieved. By testing for the shared library directly you would be bypassing the dynamic name creation routine as well. The build system and dynamic name creation need to agree.

vadz:
I've added a commit (see above) illustrating what I had in mind, let's see if it works in the other PR.

vadz:
The commit in #1252 seems to work, so closing this one.

PR #1252, opened 2025-05-14
Restore ABI version use in Windows DLL names

vadz:
This was lost during CMake rewrite, but it makes sense to keep doing it, as it's a common convention under Windows due to the absence of SOVERSION equivalent, and it broke dynamic backend loading by name.

See #1248.

### DIFF
--- a/tests/sqlite3/test-sqlite3.cpp
+++ b/tests/sqlite3/test-sqlite3.cpp
@@ -1030,6 +1030,11 @@ public:
         return test_context_base::initialize_connect_string(argFromCommandLine);
     }
 
+    std::string get_backend_name() const override
+    {
+        return "sqlite3";
+    }
+
     table_creator_base* table_creator_1(soci::session& s) const override
     {
         return new table_creator_one(s);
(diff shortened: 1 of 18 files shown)


vadz:
@Krzmbrzl Please let me know if you see anything wrong here, otherwise I'll merge it soon.

tests/sqlite3/test-sqlite3.cpp
```diff
@@ -59,6 +59,15 @@ TEST_CASE("SQLite connection string", "[sqlite][connstring]")
 
 }
 
+// Test loading backend dynamically if possible (this symbol is defined by
+// the build system if shared libraries are used).
+#ifdef SOCI_DYNAMIC_BACKEND
+TEST_CASE("SQLite dynamic backend", "[sqlite][dynamic]")
+{
+    CHECK_NOTHROW(soci::session("sqlite3", "dbname=:memory: nocreate"));
```
Krzmbrzl, line 67: Might be nice if we could test dynamic loading for all tested backends (e.g. do it as part of the common tests).

Krzmbrzl (approved):

tests/sqlite3/test-sqlite3.cpp
vadz: Yes, I admit I was too lazy to do it, but I've just done it now, thanks for poking me.

Krzmbrzl (approved):

PR #1254, opened 2025-05-16
SQLite: Ensure timeout setting first

Spixmaster:
The timeout setting is set first for SQLite as setting other settings might fail with a locked database and no timeout.

### DIFF
--- a/src/backends/sqlite3/session.cpp
+++ b/src/backends/sqlite3/session.cpp
@@ -14,6 +14,7 @@
 
 #include <functional>
 #include <memory>
+#include <sqlite3.h>
 #include <sstream>
 #include <string>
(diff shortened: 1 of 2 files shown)


vadz:
Thanks, looks good, I've just simplified/made safer the test a bit and also reduced the number of threads from 512 which seems excessive.

Spixmaster:
For clarification, I just checked whether working with database :memory: also works. It does not. We need the file. My initial assumption was correct. I assume that the memory database creates a new one for every connection.

Spixmaster:
The threads can be decreased. However, this might lead to the test not throwing an exception anymore if the fix is not applied to the source code.

PR #1268, opened 2025-06-06
Introduce Safe String Copy Utility (soci_strncpy_safe) for ODBC Backend

EmBachlitzanakis:
This pull request introduces a custom, safer string copy function, soci_strncpy_safe, to the ODBC backend alongside the necessary common.cpp class for further utilities

### DIFF
--- a/src/backends/odbc/CMakeLists.txt
+++ b/src/backends/odbc/CMakeLists.txt
@@ -25,7 +25,6 @@ soci_define_backend_target(
     "common.cpp"
   HEADER_FILES
     "${PROJECT_SOURCE_DIR}/include/soci/odbc/soci-odbc.h"
-    "${CMAKE_CURRENT_SOURCE_DIR}/common.h"
   MISSING_DEPENDENCY_BEHAVIOR "${DEPENDENCY_MODE}"
 )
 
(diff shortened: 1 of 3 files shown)


EmBachlitzanakis:
Besides the naming convention. I wanted a function that is safe and my msvc compiler will stop complaining about it strncpy in error.cpp but it fails to pass the build tests

vadz:
SOCI build system uses `-D_CRT_SECURE_NO_WARNINGS` to avoid warnings about the standard C functions, why are you not using them?

EmBachlitzanakis:
-D_CRT_SECURE_NO_WARNINGS 
I dont want to.. I want to create a easy to use template for c++ microservices that only uses vcpkg libraries and just starts from the visual studio without anything else, I dont want to set anything else, no ugly #includes and not preprocessor ....

vadz:
Sorry, this makes no sense to me. You still need to have a CMakefile (or a project or whatever) and you can define this in it.

Anyhow, using `strncpy()` here would be acceptable, but I don't want to do anything more complicated (and "ugly", as you say) in the code, so I don't believe this PR is going to be useful.

PR #1271, opened 2025-06-07
Cmake and type checks simplifications

vadz:
@Krzmbrzl @zann1x Please let me know if you have any comments.

### DIFF
--- a/cmake/find_modules/FindMySQL.cmake
+++ b/cmake/find_modules/FindMySQL.cmake
@@ -97,7 +97,9 @@ if (NOT TARGET MySQL::MySQL)
     endif()
   endif()
 
-  if (NOT MySQL_LIBRARIES)
+  # Under Windows, it's normal to not have neither pkg-config nor mysql_config,
+  # so don't warn about it, it's just annoying.
+  if (NOT MySQL_LIBRARIES AND NOT WIN32)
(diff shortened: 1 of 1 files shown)


src/core/CMakeLists.txt
Krzmbrzl, line 1: What is the intention for these changes? It seems to make the code more complex (if only slightly) and I don't see why we wouldn't do the linking when using MSVC 🤔

vadz: Different commits do different things, but if you mean the threads check, then it avoids making a number of checks that are never necessary under Windows, which not only saves some time, but also takes ~25% of CMake configure output in the logs — I was surprised to see how much of the log was just useless in the MSVC CI builds.

Generally speaking, I think the question should be: why do we need to perform configure-time checks when there is nothing to configure here?

Krzmbrzl: Well, I don't have a strong opinion on this but usually I follow the philosophy of preferring the most simple code (that is, with as few special cases as possible) that gets the job done, all other things being equal. And imo whether or not CMake produces more output during configuration doesn't matter as either users are ignoring it entirely either way or they will be able to easily filter through it to find the necessary info.

I am surprised though that searching for the threads library would produce that much output 👀

PR #1273, opened 2025-06-09
consistent DSO symbol export across POSIX/Windows + SOCI_DLL usage (redux)

phetdam:
## Summary

Clean re-application of changes in #1192 after the 4.1 release.

See #1192 for the summary details with the following exceptions:

* `soci_backend` has been replaced with `soci_define_backend_target`
* `callbacks.h` now has `failover_callback` fully inline so no `SOCI_DECL` is necessary anymore
* `soci_tests_common_static` doesn't exist anymore since SOCI builds shared/static separately using CMake

https://github.com/SOCI/soci/pull/1192#issuecomment-2594432862 is still applicable here. One extra change is done to ensure that the CMake configuration step fails if the SOCI source tree was checked out without `--recurse-submodules` when an external SQLite3 installation is not available.

Also, since `SOCI_<backend>_SOURCE` is defined using `DEFINE_SYMBOL`, it is not available when building static libraries. This means changes to `soci-sqlite3.h` and the SQLite3 backend sources to have the `sqlite3.h` header included in the `sqlite3_api` namespace, and only included in the C++ source files, leaving only forward declarations in `soci-sqlite3.h`.


src/backends/sqlite3/CMakeLists.txt
```diff
@@ -46,7 +46,11 @@ if (NOT TARGET SQLite::SQLite3)
   # available, and if we're allowed to use it.
   if (NOT "${SOCI_SQLITE3_BUILTIN}" STREQUAL "OFF")
     if (NOT EXISTS ${SQLITE3_SOURCE_FILE})
-      message(WARNING "SQLite3 not found and built-in version not available, have you cloned SOCI repository with --recurse-submodules?")
+      message(
+        FATAL_ERROR
```
phetdam, line 50: This warning can be missed amongst all the configuration output so I took the liberty of making it fatal.

phetdam:
> Ah, the failing CI stuff for SQLite3 seems like an IWYU error. I'll ensure `<sqlite3.h>` is included in the affected files.

In relation to this is `#include <sqlite3.h>` no longer being done inside the `sqlite_api` namespace in `include/soci/sqlite3/soci-sqlite3.h`. I'm not entirely sure why having `extern "C"` decls in a namespace was the right move in 2006, given that these names can still be referred to unqualified (due to C linkage), but that means the `sqlite3.h` includes need to be as follows:

```cpp
namespace sqlite_api
{

#include <sqlite3.h>

} // namespace sqlite_api
```

src/backends/sqlite3/CMakeLists.txt
Krzmbrzl: I agree that this is likely the better approach

include/soci/sqlite3/soci-sqlite3.h
```diff
@@ -31,13 +31,13 @@
...
 {
 
-// Don't include sqlite3.h from outside SOCI: this header might not be
-// available when using built-in SQLite3 as we don't install it in this case.
-#ifdef SOCI_SQLITE3_SOURCE
-    #include <sqlite3.h>
-#else // !SOCI_SQLITE3_SOURCE
-// We need just a couple of forward declarations to make this header itself
-// compile.
+// We don't include sqlite3.h from outside SOCI since this header may not
```
vadz, line 34: Why can't we keep this unchanged by just checking for `SOCI_SOURCE` instead of `SOCI_SQLITE3_SOURCE`? It looks like this would reduce the number of changes without any real drawbacks.

include/soci/soci-platform.h
```diff
@@ -75,26 +75,33 @@ namespace std {
 #endif
 
 #ifdef _WIN32
+// VS2015 support: https://msdn.microsoft.com/de-de/library/6sehtctf.aspx
```
vadz, line 78: It would be nice to avoid mixing up formatting/whitespace-only changes with significant ones.

vadz (commented):
I'm sorry, while I don't have any real objections to this PR, I just don't understand what do we gain by applying it. AFAICS the only real change is that `SOCI_DLL` is now defined under Unix too — which is, on one hand, weird (DLL is just not a Unix term) and, OTOH, could be done on its own without affecting anything else.

The PR summary speaks of consistent export, but what is inconsistent about it now? I.e. what problem does this fix?

FWIW this _will_ create extra work for me, personally, as I don't use SOCI CMake build system in my own project, so I'll have to update it to define `SOCI_XXX_SOURCE` too and while this probably won't be very difficult, I don't see what justifies spending extra time on this when we could just... not apply this PR?

phetdam:
> what is inconsistent about it now? I.e. what problem does this fix?

To me the inconsistency is when symbol visibility is applied. On Windows, `SOCI_DLL` is defined to achieve this, but it makes sense to me that when using the library as shared in general (where symbol visibility matter), `SOCI_DLL` being defined should be part of the usage requirements. This is also shown in the TnFOX example in the [GCC visibility wiki](https://gcc.gnu.org/wiki/Visibility).

> FWIW this will create extra work for me

Sorry, was not aware of this :( But let me try my best to make some points.

> what justifies spending extra time on this

It's pretty repetitive to have to manually `#define` a special symbol when building SOCI core or one of the backends. Although `DEFINE_SYMBOL` is CMake-specific, it mirrors a common practice in libraries that can be built static or shared in that for one of these flavors (typically shared, which is why I assume `DEFINE_SYMBOL` applies only the shared libraries), some macro is defined during compilation that adjusts the project-specific visibility macro being used to drive symbol export. curl has [an example](https://github.com/curl/curl/blob/1886260a95f442d947b4d1fc01be95ee55e5f5a1/include/curl/curl.h#L122C8-L136), albeit in reverse as generally it is used via the shared library instead of statically being linked.

I've gone back and thought about some things, and given how the `SOCI_SOURCE` and `SOCI_<backend>_SOURCE` macros have been originally used, I'll just consistently define these across static/shared builds. It will also rectify the situation going on with the `soci-sqlite3.h` header, and I can revert the `#include`-related commits I've made there.

Let me also address some of the code-related comments too.

include/soci/soci-platform.h
phetdam: Alright, let me drop these changes then. It's just what I had in #1192 so I had copied it over.

include/soci/sqlite3/soci-sqlite3.h
phetdam: Each separate backend/library does need its own unique build-time macro to be defined.

As mentioned [below](https://github.com/SOCI/soci/pull/1273#issuecomment-2957415854) I can revert the `#include`-related changes for the sqlite3 backend sources by unconditionally defining `SOCI_<backend>_SOURCE` during the backend compilation, even if the backends are built as static libraries.

vadz:
Sorry for complaining so much, but these CMake-related changes seem to be taking inordinate amount of the time which could (should?) be better spent on something more productive. It looks like 80% of the time I spend on SOCI is spent on struggling with CMake :-( And the Great CMake Rewrite didn't really improve this (I actually suspect the contrary, but maybe it's just because I'm particularly fed up with it right now).

I'll probably end up applying this because I don't want your work to go to waste, but I'd like to return to this a bit later, once I am capable of thinking about CMake-related issues without so much loathing again.

phetdam:
> Sorry for complaining so much, but these CMake-related changes seem to be taking inordinate amount of the time

Haha, don't worry about it. CMake definitely helps with the messy reality of building C/C++ projects, but as much as I appreciate its usefulness, I also have moments where I wonder why some CMake things take so much work.

> I'll probably end up applying this because I don't want your work to go to waste, but I'd like to return to this a bit later, once I am capable of thinking about CMake-related issues without so much loathing again.

No problem; I'll periodically keep the branch up-to-date with HEAD but won't be adding to it.

vadz (commented):
I've looked at this again and this seems to be incomplete: if we remove, say, `SOCI_FIREBIRD_SOURCE`, we should remove the test for it in `include/soci/firebird/soci-firebird.h` and use `SOCI_DECL` instead of `SOCI_FIREBIRD_DECL` in this header etc (i.e. the same thing for all the other backends), shouldn't we?

Am I missing something here?

phetdam:
Hey, thanks for reviewing. I'm a bit busy right now due to some work-related things but I'll get back to this when I can.

phetdam:
> I've looked at this again and this seems to be incomplete: if we remove, say, `SOCI_FIREBIRD_SOURCE`, we should remove the test for it in `include/soci/firebird/soci-firebird.h` and use `SOCI_DECL` instead of `SOCI_FIREBIRD_DECL` in this header etc (i.e. the same thing for all the other backends), shouldn't we?

Do you mean this part of the header?

https://github.com/SOCI/soci/blob/ce348126e7b8f8f851c0461ec4b5cf8e38f848d6/include/soci/firebird/soci-firebird.h#L14-L18

If so, no; this should remain as-is. It's because when building as shared, each of the backends is built as a separate shared library. If, for example, they all used `SOCI_DECL`, which only marks symbols for export when `SOCI_SOURCE` is defined, and they all relied on `SOCI_SOURCE` being defined during compilation, then MSVC would incorrectly mark *all* `SOCI_DECL` symbols as being exported from *all* the backends too. The general rule is for each DLL being built, you need a unique import/export macro, which is controlled by a unique build-time macro when indicating symbols are being exported.

So in SOCI's case:

1. Each backend must have its own `SOCI_<backend>_DECL` visibility macro for symbol import/export
2. `SOCI_<backend>_DECL` must resolve on Windows to `__declspec(dllexport)` when `SOCI_<backend>_SOURCE` is defined
3. `SOCI_<backend>_SOURCE` is defined *only* during compilation for the given <backend>

This ensures that in the case of shared libraries being built, that for each backend, the corresponding `SOCI_<backend>_DECL` resolves to `__declspec(dllimport)` during compilation of any source *not* part of the backend library, while resolving to `__declspec(dllexport)` during compilation of any source that *is* part of the backend library.

Admittedly, it is a bit of a confusing state of affairs. Although the idea of symbol visibility is good, the other option if not using the MSVC `__declspec` extension is to use a [`.def` file](https://learn.microsoft.com/en-us/cpp/build/reference/module-definition-dot-def-files?view=msvc-170), which is rather tedious to use.

PR #1297, opened 2025-08-14
Rename target SOCI::soci to SOCI::SOCI

Krzmbrzl:
We still keep a SOCI::soci target around, which we explicitly mark as deprecated. This way, code that is already using it will remain functional but they will be informed that they are supposed to use SOCI::SOCI instead.


vadz:
I've merged all the outstanding PRs but I'm not sure if I resolved merge conflicts correctly, could you please check the diff with [my pu branch](https://github.com/vadz/soci/tree/pu) and let me know if anything needs to be changed? TIA!

vadz:
Thanks, I've corrected this and force pushed the changes to `pu` — does it look good now?

PR #1300, opened 2025-08-21
Allow specifying odbc_option_driver_complete as part of the DSN

vadz:
It is convenient to allow specifying this option as part of the connection string itself rather than requiring setting it using set_option(), e.g. like this it can be used in the connection string used by the CI service without requiring any changes in the code, so support giving it in this way.


vadz:
Merged in the commit above.

PR #1301, opened 2025-08-21
Look for backend DLLs dependencies in their own directory first

vadz:
Use LoadLibraryEx(LOAD_WITH_ALTERED_SEARCH_PATH) to load backend dependencies from the directory containing the backend DLL itself first, while still looking in the standard search path if the dependency is not found there.

This is important to prevent the DLLs found in the application directory, which is searched before the directory containing the DLL in the standard search path, from being loaded instead and breaking SOCI backend.

The motivating example is using SOCI from Microsoft Excel add-in, which would load OpenSSL libcrypto-3-x64.dll from EXCEL.EXE directory, which contains an old version of this DLL incompatible with recent libpq.dll versions and so preventing PostgreSQL DLL, and hence SOCI Postgres plugin, from being loaded.


vadz:
Merged in the commit above.

PR #1302, opened 2025-08-23
CMake config Windows fixes

phetdam:
This PR makes some tweaks to the CMake config that fixes some issues seen on Windows:

1. The MD5 hashing in `src/CMakeLists.txt` used to check if `soci-config.h` needs to be regenerated does not serve its purpose on Windows as [`file(WRITE)`](https://cmake.org/cmake/help/latest/command/file.html#write) writes with CRLF endings. Therefore, the string and file content hashes always differ.
   * This is rectified by using [`file(GENERATE)`](https://cmake.org/cmake/help/latest/command/file.html#generate) which enables us to enforce LF line endings and also correctly avoid regeneration unless the config content has changed (and thus avoid needless rebuilding).
2. Ensure that when using CMake multi-config generators like the [Visual Studio generators](https://cmake.org/cmake/help/latest/manual/cmake-generators.7.html#visual-studio-generators) or the [Ninja Multi-Config](https://cmake.org/cmake/help/latest/generator/Ninja%20Multi-Config.html) generator, that `show_config` produces correct "Build type:" output as `CMAKE_BUILD_TYPE` is not used by multi-config generators.
   * This is addressed by using a CMake [generator expression](https://cmake.org/cmake/help/latest/manual/cmake-generator-expressions.7.html) that is conditional on [`GENERATOR_IS_MULTI_CONFIG`](https://cmake.org/cmake/help/latest/prop_gbl/GENERATOR_IS_MULTI_CONFIG.html), evaluating to [`CMAKE_CONFIGURATION_TYPES`](https://cmake.org/cmake/help/latest/variable/CMAKE_CONFIGURATION_TYPES.html) or [`CMAKE_BUILD_TYPE`](https://cmake.org/cmake/help/latest/variable/CMAKE_BUILD_TYPE.html) as appropriate.
3. Ensure that `CMAKE_BUILD_TYPE` is not left as empty if not set as a cache variable with `-D`.
   * This can be done by setting `CMAKE_BUILD_TYPE` to "Release" if empty/not defined. Therefore, this avoids "Build type:" from `show_config` from being empty if someone forgets to set `CMAKE_BUILD_TYPE` with `-D`.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -49,6 +49,11 @@ cmake_dependent_option(SOCI_LTO "Enable link time optimizations in release build
 # useful later when printing the SOCI configuration in show_config.
 get_property(SOCI_IS_MULTI_CONFIG GLOBAL PROPERTY GENERATOR_IS_MULTI_CONFIG)
 
+# Default build type to Release for single-config generators
+if(NOT SOCI_IS_MULTI_CONFIG AND NOT CMAKE_BUILD_TYPE)
+  set(CMAKE_BUILD_TYPE Release)
+endif()
+
 # Configure LTO for anything but Debug builds (if enabled in the first place)
 set(CMAKE_INTERPROCEDURAL_OPTIMIZATION ${SOCI_LTO})
 set(CMAKE_INTERPROCEDURAL_OPTIMIZATION_DEBUG OFF)


vadz:
Looks good, thanks! Merging now.

PR #1305, opened 2025-09-25
Const function for soci::blob

Spixmaster:
Added const functions to include/soci/blob.h.
Added const functions to include/soci/soci-backend.h and all backends.

### DIFF
--- a/include/soci/blob.h
+++ b/include/soci/blob.h
@@ -42,13 +42,13 @@ public:
     void initialize(session &s);
     void initialize(details::blob_backend *backend);
 
-    std::size_t get_len();
+    std::size_t get_len() const;
 
     // offset is backend-specific
 #ifndef SOCI_SOURCE
     [[deprecated("Use read_from_start instead")]]
 #endif
-    std::size_t read(std::size_t offset, void * buf, std::size_t toRead);
+    std::size_t read(std::size_t offset, void * buf, std::size_t toRead) const;
 
     // Extracts data from this blob into the given buffer.
     // At most toRead bytes are extracted (and copied into buf).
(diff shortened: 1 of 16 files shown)


include/soci/blob.h
```diff
@@ -57,22 +57,22 @@ class SOCI_DECL blob
...
-        std::size_t offset = 0);
+        std::size_t offset = 0) const;
 
     // offset is backend-specific
 #ifndef SOCI_SOURCE
     [[deprecated("Use write_from_start instead")]]
 #endif
     std::size_t write(std::size_t offset, const void * buf,
-        std::size_t toWrite);
+        std::size_t toWrite) const;
```
vadz, line 67: It seems weird to make `write()` const, even if it doesn't modify the BLOB itself, it still clearly modifies something. Why do we want to do this?

include/soci/blob.h
```diff
@@ -57,22 +57,22 @@ class SOCI_DECL blob
...
     // offset starts from 0
     std::size_t write_from_start(const void * buf, std::size_t toWrite,
-        std::size_t offset = 0);
+        std::size_t offset = 0) const;
 
-    std::size_t append(const void * buf, std::size_t toWrite);
+    std::size_t append(const void * buf, std::size_t toWrite) const;
 
-    void trim(std::size_t newLen);
+    void trim(std::size_t newLen) const;
```
vadz, line 75: Same as `write()`: these functions shouldn't be const IMO.

include/soci/soci-backend.h
```diff
@@ -342,7 +342,7 @@ class blob_backend
 
     virtual void trim(std::size_t newLen) = 0;
 
-    virtual session_backend &get_session_backend() = 0;
+    virtual session_backend &get_session_backend() const = 0;
```
vadz, line 345: This one seems wrong to me as well: `const` accessor should return const reference.

vadz (commented):
Sorry, I'm not really sure what is the motivation here, but I don't think we want to do this. It's not clear what does it mean for a BLOB member function to be const and it seems easier not to change anything but to just operate with non-const objects in your program.

include/soci/soci-backend.h
Spixmaster: `const` functions just mean that no member variable is modified.

Spixmaster:
Personally, I think every function that can be `const` should be.

The function `get_len()` should be.

vadz:
> Personally, I think every function that can be `const` should be.

This is not a useful approach. You can make all member functions `const` and then the situation is indistinguishable from not using `const` at all. The whole point of `const` is to separate between operations modifying the object and accessors only inspecting it.
 
> The function `get_len()` should be.

Yes, I agree. But I'm less sure about `read()` and `write()`. In fact, the really bad thing is that you could convincingly argue that _either_ of them should be const but that the other one should not be. Which is why I'd prefer to just not open this box at all.

PR #1337, opened 2026-02-25
Fixed propagation of NULL indicators in the soci::values::set() function

Innokentiy-Alaytsev:
Originally, the indicator passed by the user was copied into a stored-for-later  variable. This value could have been overwritten during conversion of the bound  parameter to its base type. The fixed code preserves the value of an indicator.

### DIFF
--- a/include/soci/values.h
+++ b/include/soci/values.h
@@ -165,8 +165,17 @@ public:
             indicator * pind = new indicator(indic);
             indicators_.push_back(pind);
 
+            // If a NULL indicator was passes by the user, we must preserve it,
+            // even if conversion to the base type was successful.
+            indicator conv_ind = i_ok;
+
             base_type baseValue{};
-            type_conversion<T>::to_base(value, baseValue, *pind);
+            type_conversion<T>::to_base(value, baseValue, conv_ind);
+
+            if (indic != i_null)
+            {
+                *pind = conv_ind;
+            }
 
             details::copy_holder<base_type> * pcopy =
                     new details::copy_holder<base_type>(baseValue);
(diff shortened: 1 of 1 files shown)


Innokentiy-Alaytsev:
> To be honest, I'd consider just forbidding uses values and positional binding together because it doesn't seem to make much sense and we can be sure that nobody is using it, knowing that it doesn't work... But if you consider it useful enough to keep it, let's merge this PR.

Forbidding positional binding for `soci::values` is a reasonable solution. It will both eliminate the potential of sumbling on the same error and limit the amount of code that requires support (all positional `get()` and `set()` functions can be deleted in this case). Most of this PR will be deleted, except for the code fixing the indicator handling in the named binding case: no problemn ever emerged only because the `convert_to_base()` is called twice in the "binding pipeline".

It's up to you to decide, I'll accept any decision.

> I still dislike the name of the new function, but I have trouble coming up with something better. Maybe we could invert the function semantics and call it something like `should_skip()`?

The name is bad, I agree. Maybe `skip_binding()`? I'm not sure about proper comment documentation, because it is a kind of a crutch.

vadz:
> It's up to you to decide, I'll accept any decision.

If you think it's worth having this for your bindings, I'm ready to accept it. If you don't care (enough) about it, I'd rather make things simpler by just forbidding this.

> > I still dislike the name of the new function, but I have trouble coming up with something better. Maybe we could invert the function semantics and call it something like `should_skip()`?
> 
> The name is bad, I agree. Maybe `skip_binding()`? I'm not sure about proper comment documentation, because it is a kind of a crutch.

Yes, `skip_binding()` would be better. The comment should basically just say that this is needed to avoid binding `values` themselves — or at least this is how I understand it.

PR #1346, opened 2026-03-19
Allow multiple find_package calls

### LINKED ISSUE #1344: [CMake]  add_library cannot create target "soci_interface"
>add_library cannot create target "soci_interface" because another target
  with the same name already exists.  The existing target is an interface
  library created in source directory

An error will be reported when the second time I called find_package(SOCI CONFIG).
add_library probably shouldn't  be used in config.cmake

Krzmbrzl:
We now only define our targets if they don't exist yet.

Note that multiple calls to find_package(SOCI) that specify different components will cause all call-sides to use the superset of all requested components as targets are global (and hence also their link dependencies).

Fixes #1344

### DIFF
--- a/soci-config.cmake.in
+++ b/soci-config.cmake.in
@@ -157,17 +157,23 @@ unset(__dep_dep_targets)
 check_required_components(SOCI)
 
 if (NOT DEFINED SOCI_FOUND OR SOCI_FOUND)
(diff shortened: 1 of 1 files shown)


vadz:
Applied in the commit above, thanks again.

PR #1351, opened 2026-05-31
Harden find modules

Krzmbrzl:
This now avoids issues in case we search for a dependency more than once and switches to creating imported targets instead of interface ones. The former have the correct semantics when it comes to things like installation etc. and also they will be scoped to the calling directory or lower, avoiding clutter in unrelated (super) projects.


Krzmbrzl:
The ODBC CI failures seem unrelated to the changes suggested here so I will ignore them for the time being :eyes: 

Not sure what the SQLite timeout in the "MSVS" CI run means... since I didn't touch anything SQLite related I tend to think that this might also be an independent CI breakage?

vadz:
It looks like Ubuntu dbgsym repositories are broken again. I am not sure what to do about this, to be honest...

Windows CI error was just a fluke, it went away after rerunning the job.

vadz:
Looks like I've fixed this in #1352, and improved it, too, so I'll merge this soon.

PR #1352, opened 2026-05-31
Use debuginfod in Ubuntu CI jobs if it is available

vadz:
This avoids problems with dbgsym packages that regularly get out of sync and should also be faster, as the debug symbols will only be retrieved if necessary instead of being done every time the job runs.

This commit is best viewed ignoring whitespace-only changes.


vadz:
Applied in the commit above.

PR #1355, opened 2026-06-13
Installation: Fix issues in dependency handling

### LINKED ISSUE #1354: `fmt` dependency not properly propagated to config file
When installing SOCI that was built with an external `fmt` library (instead of the bundled version), the generated (and installed) cmake config script doesn't properly handle the `fmt` dependency.

Krzmbrzl:
Fixes #1354

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -139,9 +139,6 @@ elseif(NOT SOCI_FMT_BUILTIN)
   endif()
 endif()
 
-# Name of the library to link with.
-set(soci_fmt fmt::fmt)
-
 if(soci_use_bundled_fmt)
   if (NOT EXISTS ${PROJECT_SOURCE_DIR}/3rdparty/fmt/CMakeLists.txt)
     message(
@@ -152,9 +149,6 @@ if(soci_use_bundled_fmt)
   endif()
 
   add_subdirectory(3rdparty/fmt EXCLUDE_FROM_ALL)
-
-  # Use the header-only version to avoid forcing SOCI users to link with it.
-  set(soci_fmt $<BUILD_INTERFACE:fmt::fmt-header-only>)
 endif()
 
 # Helper function appending prefix and suffix to the library name.
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -149,7 +149,10 @@ if(soci_use_bundled_fmt)
   endif()
 
   set(FMT_INSTALL ${SOCI_INSTALL})
-  add_subdirectory(3rdparty/fmt)
+  if (CMAKE_VERSION VERSION_GREATER_EQUAL 3.25)
+    set(SYSTEM_FLAG "SYSTEM")
+  endif()
+  add_subdirectory(3rdparty/fmt ${SYSTEM_FLAG})
 endif()
 
 # Helper function appending prefix and suffix to the library name.
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -148,7 +148,8 @@ if(soci_use_bundled_fmt)
     )
   endif()
 
-  add_subdirectory(3rdparty/fmt EXCLUDE_FROM_ALL)
+  set(FMT_INSTALL ${SOCI_INSTALL})
+  add_subdirectory(3rdparty/fmt)
 endif()
 
 # Helper function appending prefix and suffix to the library name.
--- a/cmake/soci_define_backend_target.cmake
+++ b/cmake/soci_define_backend_target.cmake
@@ -149,8 +149,13 @@ function(soci_define_backend_target)
   endforeach()
 
   # Set settings common to all backends.
-  target_link_libraries(${DEFINE_BACKEND_TARGET_NAME} PUBLIC SOCI::Core PRIVATE $<BUILD_INTERFACE:fmt::fmt-header-only>)
+  target_link_libraries(${DEFINE_BACKEND_TARGET_NAME} PUBLIC SOCI::Core)
   target_include_directories(${DEFINE_BACKEND_TARGET_NAME} PRIVATE "${PROJECT_SOURCE_DIR}/include/private")
+  soci_private_dependency(
+    NAME fmt
+    DEP_TARGETS fmt::fmt
+    TARGET "SOCI::${DEFINE_BACKEND_NAME}"
+  )
 
   soci_build_library_name(full_backend_target_name "${DEFINE_BACKEND_TARGET_NAME}")
 
--- a/cmake/soci_define_backend_target.cmake
+++ b/cmake/soci_define_backend_target.cmake
@@ -149,7 +149,7 @@ function(soci_define_backend_target)
   endforeach()
 
   # Set settings common to all backends.
-  target_link_libraries(${DEFINE_BACKEND_TARGET_NAME} PUBLIC SOCI::Core PRIVATE ${soci_fmt})
+  target_link_libraries(${DEFINE_BACKEND_TARGET_NAME} PUBLIC SOCI::Core PRIVATE $<BUILD_INTERFACE:fmt::fmt-header-only>)
   target_include_directories(${DEFINE_BACKEND_TARGET_NAME} PRIVATE "${PROJECT_SOURCE_DIR}/include/private")
 
   soci_build_library_name(full_backend_target_name "${DEFINE_BACKEND_TARGET_NAME}")
--- a/cmake/soci_utils.cmake
+++ b/cmake/soci_utils.cmake
@@ -50,19 +50,6 @@ function(soci_verify_parsed_arguments)
   endif()
 endfunction()
 
-
-# Initialize variables populated by soci_public_dependency
-set(SOCI_DEPENDENCY_VARIABLES
-  "SOCI_DEPENDENCY_SOCI_COMPONENTS"
-  "SOCI_DEPENDENCY_NAMES"
-  "SOCI_DEPENDENCY_TARGETS"
-)
-if (NOT DEFINED SOCI_UTILS_ALREADY_INCLUDED)
-  foreach(VAR_NAME IN LISTS SOCI_DEPENDENCY_VARIABLES)
-    set("${VAR_NAME}" "" CACHE INTERNAL "")
-  endforeach()
-endif()
-
 # This function declares a public dependency of a given target in such a way that
 # the dependency is automatically populated to a generated SOCI config file.
 #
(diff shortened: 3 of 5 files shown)


Krzmbrzl:
With this patch applied, the vcpkg port targeting what currently is `master` builds nicely (at least on Linux) without the need for any extra patches :+1:

CMakeLists.txt
```diff
@@ -139,9 +139,6 @@ elseif(NOT SOCI_FMT_BUILTIN)
   endif()
 endif()
 
-# Name of the library to link with.
-set(soci_fmt fmt::fmt)
```
vadz, line 143: Sorry, I don't want to remove this. I want the superproject to tell SOCI to use its version of fmt by predefining this target before including SOCI submodule, so we should link with `fmt::fmt` at least when this target is already defined.

vadz (changes requested):
I think the correct behaviour in case the system library is present is to link with it (as we already do) and depend on it (as we don't). Using header-only library is simpler, but not always desirable.

Krzmbrzl:
I think we should either always link `fmt::fmt` or always link`fmt::fmt-header-only`. Swapping around seems like it adds complexity without real benefit.

> Using header-only library is simpler, but not always desirable.

When would you rather have another dependency? If you want the smallest binary possible?

> I want the superproject to tell SOCI to use its version of fmt by predefining this target before including SOCI submodule, so we should link with fmt::fmt at least when this target is already defined.

If that's the goal, there should be a variable for it. That seems way more flexible than using a specifically named target

vadz:
> I think we should either always link `fmt::fmt` or always link`fmt::fmt-header-only`. Swapping around seems like it adds complexity without real benefit.

I don't want to add any new dependencies to SOCI, but I do want to allow people to opt in into linking with `fmt::fmt`.

> > Using header-only library is simpler, but not always desirable.
> 
> When would you rather have another dependency? If you want the smallest binary possible?

When you already use `fmt::fmt`, as many projects do. In fact, in this case it makes no sense at all to use the bundled header-only library: this is slower (extra stuff to compile), results in more code and risks creating surprises due to using different versions of the library in the main application and SOCI.
 
> > I want the superproject to tell SOCI to use its version of fmt by predefining this target before including SOCI submodule, so we should link with fmt::fmt at least when this target is already defined.
> 
> If that's the goal, there should be a variable for it. That seems way more flexible than using a specifically named target

Maybe, but this variable should still have a default value. And the default value which makes sense corresponds to the current behaviour. So for me the priority is to fix the behaviour with the current logic, adding a variable to make it more flexible could be useful but definitely not critical.

Krzmbrzl:
> In fact, in this case it makes no sense at all to use the bundled header-only library: this is slower (extra stuff to compile), results in more code and risks creating surprises due to using different versions of the library in the main application and SOCI.

With the current set of changes we would still be using the system (or already used) version of `fmt`. The (presumably slight) compilation overhead would still be there though. I would be surprised if that actually turns out to be a real noticeable effect 👀

vadz:
In my use case we would use different versions because `fmt::fmt` in the superproject doesn't have to be (and is not, in my case) the same version as the bundled `fmt::header-only`.

Really, let's please keep using `fmt::fmt` if it's already defined, it's not complicated.

Krzmbrzl:
> the same version as the bundled fmt::header-only.

We're not using the _bundled_ version at all in such a case. That's what I'm trying to say.

> Really, let's please keep using fmt::fmt if it's already defined, it's not complicated.

If we don't use the header-only version we have to properly propagate the dependency throughout the config file. That's what makes it less trivial and would require more sophisticated testing to avoid running into the same issue as current master has.

vadz:
> > the same version as the bundled fmt::header-only.
> 
> We're not using the _bundled_ version at all in such a case. That's what I'm trying to say.

Then the build will probably just fail because I predefine my own `fmt::fmt` which works correctly but nothing else, i.e. not the header-only target.
 
> > Really, let's please keep using fmt::fmt if it's already defined, it's not complicated.
> 
> If we don't use the header-only version we have to properly propagate the dependency throughout the config file.

You can use the header-only version of the system library if you're so inclined. IMO it would be a regression but one that I care less about.

Ideal would, of course, be indeed to propagate dependency on the system library when we use it.

Krzmbrzl:
> Then the build will probably just fail because I predefine my own fmt::fmt which works correctly but nothing else, i.e. not the header-only target.

Yes, if you mimic only parts of `fmt` the build will be broken. Afaict that is never really defined behavior in general. And if you specifically use it for `soci` you could either change the name of your target or add a corresponding alias to mock a header only target. That'll work perfectly as long as you're not installing soci via cmake.

> You can use the header-only version of the system library if you're so inclined

That's what this PR does.

> Ideal would, of course, be indeed to propagate dependency on the system library when we use it.

Well that's basically what I said earlier: but imo we need to make a decision. Either always use the header-only version or never. That always leaves us with a clear path of what happens and what needs to be done.

vadz:
> > Then the build will probably just fail because I predefine my own fmt::fmt which works correctly but nothing else, i.e. not the header-only target.
> 
> Yes, if you mimic only parts of `fmt` the build will be broken.

Yet it works just fine right now and I don't see what does anybody gain from breaking it.

> > You can use the header-only version of the system library if you're so inclined
> 
> That's what this PR does.

No, it also uses it when using predefined fmt::fmt and this is why I don't want to merge it.
 
> > Ideal would, of course, be indeed to propagate dependency on the system library when we use it.
> 
> Well that's basically what I said earlier

So we both agree, good.

> but imo we need to make a decision. Either always use the header-only version or never.

What decision? I don't see any problems that wouldn't be addressed by what is just above and which you seem to agree with.

We're going in circles, but I have no idea why, things seem pretty clear to me. To summarize, there are 3 use cases:

1. Superproject predefines `fmt::fmt` target. SOCI should use it and shouldn't do anything for installing it because it's never going to be installed in this case.
2. System library exists. SOCI should use it (if not disabled) and should expose its dependency on it in the installed files.
3. System library doesn't exist or its use is disabled. SOCI should use bundled header-only version and shouldn't do anything for installing it as it's a build-time only dependency.

The *only* problem is in the scenario (2). And there are 2 ways to address it:

- Do it properly, which we all agree is a good idea, but apparently too complicated to do.
- Use header-only version of system library which is not ideal but I can live with that.

But we don't need to and shouldn't change anything for the scenarios (1) and (3).

Do you disagree with anything here?

Krzmbrzl:
> Do you disagree with anything here?

Yes.

> SOCI should use it and shouldn't do anything for installing it because it's never going to be installed in this case.

This assumption is not necessarily true. Even if SOCI is used in a superproject it might get installed. And then it's broken unless the predefined target happens to be header-only.

> SOCI should use bundled header-only version and shouldn't do anything for installing it as it's a build-time only dependency.

No, if other routes may end up using the non-header-only version then we should stick to that. Then SOCI will have a private dependency on `fmt` no matter what instead of only sometimes (which seems super weird and inconsistent to me).

That would streamline handling inside of SOCI and also when using SOCI as a dependency.

> Do it properly, which we all agree is a good idea, but apparently too complicated to do.

I never said it's too complicated. The "do it properly" approach should just be "always use the same dependency regardless of where it came from instead of sometimes header-only and sometimes not". It's slightly _more_ complicated than always using the header-only version but certainly doable.

Krzmbrzl:
While implementing the necessary changes to make linking to `fmt::fmt` work properly I noticed that there are a couple of other complications. I'll try to include fixes for them in this PR as well

vadz:
> > SOCI should use it and shouldn't do anything for installing it because it's never going to be installed in this case.
> 
> This assumption is not necessarily true. Even if SOCI is used in a superproject it might get installed.

This is not a supported use case. There has never been any demand for this and I refuse to complicate CMakefiles further without any good reason, so supporting installing the library used from a superproject is an explicit non-goal.

In fact, I think all the install-related code should be entirely disabled if `NOT ${PROJECT_IS_TOP_LEVEL}`.

> > SOCI should use bundled header-only version and shouldn't do anything for installing it as it's a build-time only dependency.
> 
> No, if other routes may end up using the non-header-only version then we should stick to that.

You didn't quote the context but is only for the case (3). And in this case there are no other routes.

> Then SOCI will have a private dependency on `fmt` no matter what instead of only sometimes (which seems super weird and inconsistent to me).

This results in absolutely 0 problems to the best of my knowledge. And it's perfectly fine to have different dependencies depending on which libraries are available. So changing this is, again, an explicit non-goal.
 
Please keep in mind that we want to support the 3 use cases above and _no_ others.

Krzmbrzl:
> There has never been any demand for this and I refuse to complicate CMakefiles further without any good reason, so supporting installing the library used from a superproject is an explicit non-goal.

That's a standpoint that doesn't make sense to me and I do not share it. It's perfectly normal to want to do something like this. In fact, we do make use of it. So it may not be a goal of yours, but it's certainly one of mine.

Besides, there isn't really any additional issues if we simply handle installation properly in all other cases.

> You didn't quote the context but is only for the case (3). And in this case there are no other routes.

The context is the general one. I want to use a single dependency for all of your three use cases. Deviating from that is what causes unnecessary complication of cmake files.

> This results in absolutely 0 problems to the best of my knowledge.

Us having this conversation in the first place is a counterexample to this statement. The install issue has only remained undiscovered because there were different possibilities of depending on `fmt` and not all have been tested. With increased variability the required testing amount grows combinatorically.

----

What issues do you see with the approach to link to `fmt::fmt` in all scenarios?

EDIT: You can refer to the current state of this PR for how the associated CMake changes would/could look like.

vadz:
> > There has never been any demand for this and I refuse to complicate CMakefiles further without any good reason, so supporting installing the library used from a superproject is an explicit non-goal.
> 
> That's a standpoint that doesn't make sense to me and I do not share it. It's perfectly normal to want to do something like this. In fact, we do make use of it. So it may not be a goal of yours, but it's certainly one of mine.

OK, let's postpone it after 4.1.3 release. It is definitely not a goal for it and, again, I refuse to do something that nobody has asked for (do you even have a use case for this yourself?) and which doesn't make sense to me.
 
> Besides, there isn't really any additional issues if we simply handle installation properly in all other cases.
> 
> > You didn't quote the context but is only for the case (3). And in this case there are no other routes.
> 
> The context is the general one. I want to use a single dependency for all of your three use cases. Deviating from that is what causes unnecessary complication of cmake files.

There can be no single dependency as sometimes we're going to use a library and sometimes a header-only version.
 
> What issues do you see with the approach to link to `fmt::fmt` in all scenarios?

This is not going to work when using bundled library with static SOCI libraries.

Krzmbrzl:
> This is not going to work when using bundled library with static SOCI libraries.

Why? I tested it and it seemed to work just fine.

vadz:
> > This is not going to work when using bundled library with static SOCI libraries.
> 
> Why? I tested it and it seemed to work just fine.

I don't know what you tested, but when you link with static SOCI libraries using fmt as a library, you also need to link with fmt and my intention is to not add an extra dependency to the application using SOCI unless it opts in into it.

I'm sorry but you seem to be completely missing the intention/rationale/goal here. It would really help if you could accept the fact that fmt is supposed to be an internal implementation detail not impacting SOCI users in any way. Unconditionally linking with fmt::fmt is *never* going to satisfy this goal.

Krzmbrzl:
Okay sorry but then your whole point about not always linking the header-only version doesn't make any sense. If the only point was that users have a customization option then just make that customization option be the `fmt::fmt-header-only` target as I suggested before. This makes for much, much, much clearer semantics anyway. You're not forced to make your customization point a header-only library but then you'll have to live with installs being broken (and the target name hints at this at least for folks that know what they are doing).

vadz:
I'm sorry, I don't understand what problem do you see. In which of the 3 scenarios listed above does it happen and what exactly is it?

Krzmbrzl:
You mean in current master? It's broken when using an installed version of SOCI that didn't use the bundled `fmt` version.

vadz:
The simplest fix for this is to just not use the system library unless specifically requested. This would be a trivial fix and I can make it myself.

Krzmbrzl:
It's the simplest fix but it's also the wrong one. Every single package maintainer I ever talked to doesn't want to use bundled dependencies (and this is also true for vcpkg). And many users don't want that either (it doesn't scale well if every dependency bundles its dependencies instead of using system libraries).

vadz:
You said it was not complicated to fix this. If you can do it for *just* this case, i.e. when using the system library, it would be great. If not, we'll just have to live with this for now.

Krzmbrzl:
What is your problem with fixing it properly? Why would you want to restrict the fix to a subset when the fix for the entire thing already exists? That doesn't make any sense to me.

We have a working fix that works perfectly. The only thing you don't quite like is the fact that when using a static library users have to link `fmt` as well. But that is just how static libraries work. The user won't even know that this is happening unless they specifically check the invoked commands.
I don't see how this would be a problem (cmake handles everything automatically).

We have yet another solution that keeps `fmt` entirely hidden. However, you say that this doesn't work for you because you are doing something nonstandard with custom `fmt::fmt` target that for some reason can't have a different name. Again, I don't understand why this target name couldn't be adapted or aliased.

Despite all this you prefer to not fix the issues in favor of... what exactly? What's the benefit of a buildsystem that has known cases in which it breaks down?

You keep stating that you want the cmake files to remain as simple as possible, yet you insist on keeping nonstandard behavior for your custom use case embedded in the official cmake build files. That seems contradictory as this just increases the complexity of the build files (albeit only slightly).
Wouldn't be the best solution if the buildsystem just works instead of people reporting issues that have to be investigated. Correctness beats simplicity in my book.

Perhaps there is something I am missing. If that is the case, please describe some details of a use case scenario in which the proposed fixes would cause problems (as in: something that leads to an error or different set of dependencies _after_ linking). At the moment I just really don't understand your point of view at all.

vadz:
I listed 3 scenarios that should work above. Your PR breaks at least one of them (I didn't look into the details to see if it breaks more). I am sorry but I just can't spend time continuing this discussion when you simply ignore my concerns/goals and instead try to explain to me that I shouldn't have them in the first place. This is not a productive way to discuss and I think we should just close this if you refuse to support the listed use cases.

vadz:
I'm closing this because I believe #1356 solves your problem (and only the first commit there is really needed for this, i.e. it's a truly trivial change and if it's really all that's needed, it is very disappointing to have spent so much time on discussing it). Please explain what still doesn't work for you if it doesn't.

Krzmbrzl:
This PR makes all 3 scenarios work as it currently is ("work" as in user can use SOCI and will never have any built issues because of it). Please just test it.

Also, this PR fixes more than just the `fmt` dependency issue that your alternative suggestion doesn't deal with at all.

If you don't insist on keeping everything as simple as possible I can even make the "we sometimes use header only but sometimes not" scenario work. But that necessitates a bit more logic to work properly.

EDIT: I just went ahead and did that, assuming that your list is more important to you than the simplicity argument. -> #1357

PR #1356, opened 2026-06-14
Fix dependency on fmt in SOCI config file

vadz:


Krzmbrzl (changes requested):
This breaks if the pre-defined `fmt;:fmt` dependency is non-static and not imported. And in cases where `fmt` is a static library and SOCI is a shared library, this adds an unnecessary dependency on `fmt` into the config file. Lastly, this can break because the dependency on the `Threads` library is not properly propagated to the config file (becomes an issue if SOCI is built as a static library).

All of that is fixed by #1357 (while handling the entirety of your listed requirements).

vadz:
> This breaks if the pre-defined `fmt;:fmt` dependency is non-static and not imported.

What exactly breaks? As we've established, if `fmt::fmt` is provided by the superproject, this superproject is responsible for providing/installing it, so I don't see any problem in this case.

> And in cases where `fmt` is a static library and SOCI is a shared library, this adds an unnecessary dependency on `fmt` into the config file.

Fair enough, but is this actually a problem? Especially knowing that the system library is probably never going to be static in practice.

> Lastly, this can break because the dependency on the `Threads` library is not properly propagated to the config file (becomes an issue if SOCI is built as a static library).

This is something I'm not aware about but looks like a completely different problem, so it makes sense that it's not addressed by this PR.
 
> All of that is fixed by #1357 (while handling the entirety of your listed requirements).

That PR is 10 times larger than this one and while it may do the right thing, it's really non obvious to me that it does. Other than `Threads`, what is the advantage of applying it rather than this one?

Krzmbrzl:
> What exactly breaks? As we've established, if fmt::fmt is provided by the superproject, this superproject is responsible for providing/installing it, so I don't see any problem in this case.

SOCI doesn't declare a dependency on it. So even if the superproject installs `fmt`, SOCI won't use it and thereby produce linker errors.

> Fair enough, but is this actually a problem? Especially knowing that the system library is probably never going to be static in practice.

In vcpkg the default is for libraries to be static. So if you have your dependencies in vcpkg while compiling SOCI, this will make the error surface.

> This is something I'm not aware about but looks like a completely different problem, so it makes sense that it's not addressed by this PR.

It's unrelated to `fmt` but has basically the same underlying issue.

> That PR is 10 times larger than this one and while it may do the right thing, it's really non obvious to me that it does. Other than Threads, what is the advantage of applying it rather than this one?

It unifies the handling of private dependencies with all other dependencies. We have a method that is tested and works for propagating dependencies and it seems overall simpler to me if we can use the exact same framework for public and private dependencies. This significantly reduces the risk of introducing another breakage with another dependency (as was done with `fmt`).

On top, most of the size of #1357 comes from docstrings for the newly introduced utility functions along with adapting the variable names in the implementation to reflect that the surrounding function is now called differently. If you discount lines changed because of that the PR is only about 2-3 times as big but also works for in more cases (in fact, all cases I considered and tested).

> it's really non obvious to me that it does. 

Do you want me to explain better how it works? The gist of it really is we use `soci_private_dependency` in full analogy to how we use `soci_public_dependency`. The main change in the handling of those functions is a step determining whether a given dependency is build-time or link-time and only in the latter case propagate to the SOCI config.

vadz:
Replaced by #1357.

PR #1357, opened 2026-06-15
Installation: Fix issues in dependency handling

Krzmbrzl:


vadz:
OK, I'll merge this as I don't see any problems with the changes here even though I still wish there were fewer of them.

Thanks!

PR #1370, opened 2026-08-29
Specify minimum fmt version

Krzmbrzl:
SOCI uses `fmt::underlying` which seems to have been introduced via https://github.com/fmtlib/fmt/commit/0b2862a1e4cf24797424501a3d13e3e563a90528 and the oldest tag containing this commit is `9.1.0`.

Without this SOCI will fail to build on Ubuntu 22.04 with `libfmt-dev` installed from the repositories (which is version 8.1) with this error message:
```
src/core/statement.cpp:815:35: error: ‘underlying’ is not a member of ‘fmt’
  815 |                              fmt::underlying(dbtype)));
      |                                   ^~~~~~~~~~
```


vadz:
Sure, thanks.

We could also avoid using `fmt::underlying()` but I think it's reasonable to require fmt 9.1 by now.

Krzmbrzl:
Yeah, I also think 9.1 is old enough to use as a minimum requirement :+1:

